PR #517, opened 2015-01-18
Recoded HTML-based documentation

darkness3560:
I've recoded almost every documentation file in /usr/share/doc to be better organized and more consistent.


mavrothal:
Thanks a lot!
Are you "darkcity"?
Can you mention a couple of the improvements? 
Is the transition machine made? ie does the text needs proof-reading?
If there are text changes can you point them out separately? (I do think I have the time to read 32 pages :-)

darkness3560:
No, "darkcity" is someone else.
The regular expressions tutorial page, as an example, is greatly improved.  The code displays only use one <code> element per display instead of an entire table each.  The code across all pages is much, much clearer via use of the <section> tag, as well as the table-based layout being dropped entirely.  Puppy.css has been removed as it was essentially a duplicate of manual.css.  Indentation is better as well.
The changes were done by hand, but although I corrected a few small grammatical errors, the documentation still needs lots of proofreading and maybe content updates.  My recoding should, however, make the process of updating the pages' content easier.

mavrothal:
OK. Sold! ;)

mavrothal:
Oooops.
I just realized that your pull request was for the master branch.
All pull requests should be going to the testing branch as puppies are build from this branch and changes are properly tested before they are merged into the master branch.
So please issue another pull request with the same changes but for the testing branch

PR #520, opened 2015-02-05
Delete some unneccessary echo commands

MochiMoppel:
Deletion of some echo commands which most likely were inserted into the code for (variable) testing purposes but are not needed anymore


mavrothal:
It would appear that you want to merge the testing branch back into the master branch.
I'm sure is by mistake.
Please issue the pull request from your testing branch to the woof-CE testing branch.
Closing

PR #521, opened 2015-02-05
Delete unnecessary echo commands

MochiMoppel:
These echo commands seem to be a leftover of KRG's testing session. They unnecessarily clutter console output when using petget in a text console.


mavrothal:
There you go :-)

PR #522, opened 2015-02-06
Prevent zero length file <pkgname>.files when user renames pet pkg

MochiMoppel:


mavrothal:
I understand is a test, but if I pull this request ill always be in commit history.
More important though I have a small problem with the patch itself. (see your comment on your commit)

MochiMoppel:
Sorry about the 2 garbage files in the pull request. It seems that all commits  in testing go to the request, I can't exclude them. Can you cherry pick only the patch and reject the rest?

As for the patch I don't understand your scenario of a malformed package. How does it relate to the patch?

Firstly: I don't think it's realistic to assume that a user will never rename a pet package. Some of the names are very long, so why shouldn't he? Just to say "Your fault!" is not the right problem solution.

Secondly: The patch does what the original code does, only more robust and error proof. 
Example: One of the pets in my collection is named icewm-1.3.7-i686-s.pet
The file structure inside the package is
./icewm-1.3.7-i686-s/usr/bin/icehelp
./icewm-1.3.7-i686-s/usr/bin/icewm      etc.
In order to allow later uninstallation, the file icewm-1.3.7-i686-s.files has  to be created with entries like
/usr/bin/icehelp
/usr/bin/icewm
So the task for the script is to remove the  first 2 directory levels. There  is no security advantage if the script - as it does now - removes the levels only when the pet filename is the same as directory names used within the package. It's just an unneccessary restriction that can lead to an empty pkgname.files. And this should never happen.

If in my example I would rename the pet to icewm.pet, with the patch the file  icewm.files would be created, correctly with the same content as for a not renamed pet. The only downside of renaming would be a missing description in the uninstallation dialog, but that's hardly critical.

mavrothal:
Regardless of malformed or malicious etc (which is probably an overstatement...) we should not support package name change by the user because is trouble. 
Example you instal your renamed "icewm.pet" after a month something goes wrong with icewm and you install icewm-1.3.7-i686-s.pet which installs as a different pet since it has a different name. Then you remove this "non-functioning"  "icewm.pet" and icewm is gone! Obviously the  icewm-1.3.7-i686-s.pet is "broken"...
Package renaming is bad idea. We should not support it.
The only reasonable way to do it, is to check if the package is install, extracting the pet.specs file and verifying the pet from fields 2 and 3 in user installed packages (field one comes from the pet name) before you copy the files into the system.
Even then, is too much trouble and system slowdown, just to support user whims...
So, I will not pull it but will leave it open for some time in case other woof-CE members think is OK, so they can pull it.

MochiMoppel:
Yes, package renaming is a bad idea and nobody wants to encourage it, but we can't prevent it, and if a small correction to the code can make the devastating effects of such "user error" less painful, I still think we should do it.  "Too much trouble and system slowdown" - please be reasonable. The  code is shorter than the original code and probably even marginally faster. I'm not proposing a change that would us bend over backwards just to support the remotest scenario. 

Sorry, but I find this  discussion very disappointing. Do whatever you think is right. You are in the driver seat.

mavrothal:
Slowdown refers to extracting, checking pet.specs if it is OK and the going on, not to the sed command.
And even if we do that then all the PPM scripts should be revised to get pet mane from fields 2 and 3.
Thus the trouble and slowdown.

mavrothal:
What about if we add this at the top of petget instead ? (after the removal check)
```
if [ "`echo -n "$PASSEDPARAM" | grep '.pet'`" != "" ];then
 mkdir -p /tmp/petget_check 
 tar xf "$1" -C /tmp/petget_check
 PETNAME=$(cat /tmp/petget_check/*/pet.specs | cut -f 1 -d '|')
 [ "$(echo -n "$1" | grep $PETNAME)" = "" ] && echo error && \
  rm -rf /tmp/petget_check && exit 1
 echo OK
 rm -rf /tmp/petget_check
fi
```
With the proper UI explanations of course.

mavrothal:
Or maybe this solution is simpler: https://github.com/puppylinux-woof-CE/ppm_auto/commit/80a24fcd75a1c2eb24468b9b221ead66495d7428

MochiMoppel:
"Slowdown refers to extracting, checking pet.specs if it is OK and the going on,"

Why on earth do you see a need to do this? You unpack, check pet.specs and if the name is not idential with the filename bring installation to a halt?  If you really want to check for renamed files you could simply take the current code. The current pattern already provides a rename test, so there is absolutely no need for your new check. If sed finds no match, the file was renamed, it's that  simple. 

It's still beyond me what keeps you from correcting  the sed pattern and I also don't understand  your concern for fields 2 and 3. The purpose of the sed command has nothing  to do with fields in DB_ENTRY. Petget can create DB_ENTRY from the pkg filename, and surely fileds 2 and 3 will contain "best guess" values taken from the  filename. Good enough for installation. Petget doesn't bother to open the pet and checks if pet.specs contains different names, so why would you?  PPM uses field 1 for the uninstall list, user-installed-packages will have the new filename in field 1 and pet.files will be named accordingly. Apart from field one none of the other fields is relevant for the installation process. I simply don't understand what your concerns are.

mavrothal:
You asked : "what keeps you from correcting the sed pattern"
You said: "package renaming is a bad idea and nobody wants to encourage it". 
Apparently you want to encourage it or at least allow it.  I do not.
Anyway, if someone else with commit rights wants to encourage/allow package renaming, can just merge this pull request.

MochiMoppel:
Oh boy. How much I had hoped that someone would at least test my request and then tell me if and where it fails. A solid technical discussion. I never expected to meet so much resistance and now insults.

Believe it or not, I bothered to look at your code and I'm not delighted.  Another layer of band-aid on an already severely battered script. You erect another wall to fight the  user instead of fixing the code. Will not work. And you will need more band-aid: So far you "fixed" pet files, but renaming .deb packages can produce even worse problems because the current code  doesn't check certain variables properly. Can be fixed, but would require an additional line of code ... or a new wall.

mavrothal:
I'm not sure about insults or who is insulting whom. I'll just assume is just a cultural difference and leave it to that.
Regarding your suggestion I simply think is wrong to allow package renaming, and your demand to allow it is not going to change my idea. To be able to install a package under 2 or more different names is a source of trouble regardless of the technical solution that would allow this. 
The only thing that I find acceptable is to provide code that would allow a renamed package to be verified/installed/uninstalled under its original name. 
So, hopefully you'll allow me the right of an opinion. 
Besides your code is here, and whoever wants can simply pull it or add it to its puppy.

mavrothal:
After 5 months I think we can close this.
Besides the code is still available here and [also in the forum] (http://www.murga-linux.com/puppy/viewtopic.php?p=850286#850286)

PR #528, opened 2015-04-16
Added Devuan support

dimkr:
The stock configuration produces a booting Puppy with working X and the package selection is heavily trimmed, so there should be no unneeded bloat.

100% systemd-free.


mavrothal:
Looks good to me. 
Leave it for a day or so for others to check too and then pull.
Where is the build? ;)

dimkr:
Well, the ISO is not very useful. It has nothing except working X with Leafpad :)

It's fully reproducible, though: everything is done right, via woof-CE - no manual tweaking needed.

My aim is to make Devuan support available in upstream woof-CE, for 4 reasons:
1) To give Puppy at least one bloat- and systemd-free way forward
2) As a proof-of-concept that might help other small distros that struggle to remain systemd-free
3) To start a new generation of the dpup family, based on Jessie
4) To show the world Devuan is a serious project which sees collaboration as a good thing

Sure, I don't mind waiting a couple of days.

PR #532, opened 2015-05-03
a few cosmetic changes in /usr/sbin/filemnt to improve readability

jlst:


mavrothal:
There are so many places that puppy scripts could use the bash string manipulation options instead of external utilities. Would make them much  faster.

PR #534, opened 2015-05-03
/usr/sbin/filemnt: added more file extensions

jlst:
I added more file extensions. There is a file with custom PUPPY mimetypes:
/usr/share/mime/packages/puppy.xml

I don't see it in woof-CE, that file needs to be updated with new mimetypes.

Also need to place more files in:
/root/Choices/MIME-types

And there is the free dekstop MIME types, we should also add the latest version to woof-CE, paving the way for a new Puppy.
/usr/share/mime/packages/freedesktop.org.xml

### DIFF
--- a/woof-code/rootfs-skeleton/etc/group
+++ b/woof-code/rootfs-skeleton/etc/group
@@ -27,3 +27,4 @@ cdrom::107:root,spot,fido
 tape::108:root
 tty::109:root,spot,fido
 plugdev::110:root,spot,fido
+mlocate:x:1001:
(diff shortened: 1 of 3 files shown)


mavrothal:
Please do not bunch together unrelated commits in the same pull request.

jlst:
Actually those are made-up extensions, or maybe not. I guess 2fs, 3fs are too. I have seen a udf file years ago, but I'll remove it.

I think vfat is a convenient way to rename a file to let the script choose a convenient method to mount an otherwise unknown image file.

============================

I thought that new pushes would not be counted in the current pull request, I didn't know.

I have problems reverting commits, specially when they're already pushed, I guess I have to make a hard reset locally and then force it into my testing branch.

mavrothal:
{2,3,4}fs are indeed made-up puppy-specific extensions to indicate the file system of the pupsavefile.
The rest I haven't seen and I do not think filemnt needs to "know" about those.

Regarding git and hard reset what I usually do is first to ```git format-patch <commit-hash-to-revert-to>``` to generate patches that I can later reapply and modify as needed. Then instead of revert I do a ```git reset --hard <commit-hash-to-revert-to>``` and take it from there using the patches as needed.

PR #536, opened 2015-05-04
/usr/sbin/filemnt: revert 'abort if file extension is not recognized'

jlst:
The file extension is actually the mount option,
so, when using filemnt, one must rename a file

ex:
mv diskimage.img diskimage.vfat
filemnt diskimage.vfat


jlst:
I have a better idea.

PR #537, opened 2015-05-04
/usr/sbin/filemnt: the file extension is actually the mount type.

jlst:
The script behaves as usual, but it can also be used to mount
and unmount whatever is supported by the mount utility.

sh-4.3# mv freedos10.img freedos10.vfat
sh-4.3# filemnt ${PWD}/freedos10.vfat
  Mount: /mnt/home/freedos10.vfat
sh-4.3# filemnt ${PWD}/freedos10.vfat
  UnMount: /mnt/home/freedos10.vfat


mavrothal:
I will actually pull this but can be really hardly used! 
For one the extensions are not really used and secondly ROX will not know to call filemnt when it "sees" such an extension.

PR #538, opened 2015-05-05
/usr/sbin/filemnt: comment out code to detect wrong squashfs version

jlst:
squashfs v3 is long dead. I have never come across an sfs file with squashfs v3
(I use Puppy since 2012)

That code could be reused when there is a new major release of squashfs (v5)

http://squashfs.sourceforge.net/

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/filemnt
+++ b/woof-code/rootfs-skeleton/usr/sbin/filemnt
@@ -23,27 +23,27 @@ export TEXTDOMAIN=filemnt
 export OUTPUT_CHARSET=UTF-8
 . gettext.sh
 
-#v423...
-KERNELVER="`uname -r`"
-if vercmp $KERNELVER ge 3.0;then #111016
- SFSSTR='squashfs, version 4'
- SFSMAJOR=4
-else
- KERNELSUBVER=`echo -n "$KERNELVER" | cut -f 3 -d '.' | cut -f 1 -d '-' | cut -f 1 -d '_'` #100831
- KERNELSUBSUBVER=`echo -n "$KERNELVER" | cut -f 4 -d '.' | cut -f 1 -d '-'`
- SFSSTR='squashfs, version 3'
- SFSMAJOR=3
- if [ $KERNELSUBVER -eq 27 ];then
-  if [ $KERNELSUBSUBVER -gt 46 ];then #100606
-   SFSSTR='squashfs, version 4'
-   SFSMAJOR=4
-  fi
- fi
- if [ $KERNELSUBVER -gt 28 ];then
-  SFSSTR='squashfs, version 4'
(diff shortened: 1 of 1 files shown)


mavrothal:
woof can still build "old puppies" with "old" kernels.
I'll have to reject it

jlst:
It's like saying Puppy can still support Bash v3, but I accepted code that will only be understood by Bash v4

jlst:
squashfs v4.2 can be included/compiled in a kernel as old as 2.6.29. 

I remeber I compiled Bash 4.1 to use it in both  Wary and Precise Puppy, I needed to use my scripts I wrote in Precise.

Taking that in mind, there should a official minimun requirements list for puplet builders.

jlst:
What lacks in Puppy is a true community effort to update software/packages, that's why Wary is completely outdated and Precise is way behind. 

I consider Puppy a project unsuitable for contributions and updates. So I'll go my own way again to update the whole system for my personal use.

mavrothal:
A decision to stop supporting building older puppies has not been made. 
If we decide to stop supporting build of older puppies and the existing kernels, a number of other changes are also required.
I do not think we gain anything dropping squashfs 3 support in filemnt in filemnt function or otherwise.
So I would rather keep it.
I'll re-open it in case some other team member wants to pull it, but I will not pull

jlst:
There is a sfs converter, and of top of that squashfs can easily be included in an old kernel. I many 2.6.x kernels compiled by the creator of Puppy Linux already come with squashfs v4 support.

Well right now I'm using Precise Puppy, I compiled the latest version of almost all core utilities. And I'm just checking the code, if I find something strange, I check the woof-CE code also to see if it has already been fixed. 

Overall, I'm running a more updated Precise than tharpup.. why is that? It's because I compiled some binaries/libs.

I replaced libraries like OpenSLL, libreadline, etc. with the latest version... compiled nss & nspr, and I'm seeing a more stable system overall.

Why I use precise Puppy in this board when Wary supposedly should run faster?

- I have made many changes to achieve this blazing speed
- Wrote my own lightweight scripts to handle a variety of situations.
- Runs as fast was Wary (when I first use Precise it was a bit heavy in this mainboard)
- In some cases runs even faster than Wary.

jlst:
I think I better close it, as I will not make anymore pull requests. It takes too much time using git, and I lose brain cells at alarming rate trying to explain every change in this language when I should be just modifying.

mavrothal:
Sorry to hear that.
I would think though that "community" effort implies different views and consensus building.
Otherwise you fork and get it rom there.

jlst:
Yes I will!... nah just kidding. 

I like to be behind curtains and appear when time is right. I also don't like being an evangelist, creating a new church every now and then.

I'm mainstream, but if the mainstream is going nowhere, well, I just do it alone.

PR #550, opened 2015-05-09
wmexit, wmreboot, wmpoweroff, wmrestart: use busybox ps

wdlkmpx:
In late 2013, when I firt experimented with building a puplet, I didn't know that packages-templates modified every package. So when everything was done, I placed some packages without the template filtering.

Everything seemed to be ok, but when I tried to shutdown, nothing happened.

ps was a symlink to busybox, but 'ps -C' required ps-FULL. 'ps' was a script, a hack to handle the variety of scripts that were already written.

To prevent such an incident from happening again, better use 'busybox ps'.

### DIFF
--- a/woof-code/rootfs-skeleton/usr/bin/restartwm
+++ b/woof-code/rootfs-skeleton/usr/bin/restartwm
@@ -13,13 +13,13 @@ echo -n "$NEXTWM" > /tmp/wmexitmode.txt
 #w478 testing 2.6.18.1 kernel, pup_event_frontend_d did not exit when it
 #detected /tmp/wmeximode.txt, but became 'defunct'...
 KCNT=0
-PSPEFD="`ps -C pup_event_frontend_d | grep 'pup_event_front' | grep -v 'defunct'`"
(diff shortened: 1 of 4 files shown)


mavrothal:
The ```busybox bin_app``` and the ```bin_app-FULL``` case is allover in puppy...
The suggested change is sound for the specific scripts and the current busybox version, but I was wondering if forcing the busybox applet may break things when building puppies with older/different busybox.  
Opinions?

wdlkmpx:
I downloaded Puppy 412 from here:

http://412collection.neocities.org/iso.html

Everything seems to be OK. However it looks like it doesn't use a recent version of woof.  .Bash version is 3.00. I guess the jwm version in Puppy 412 can't handle SVG images.

But I know for sure that GNU Coreutils & GNU BASH v4 compile even in prehistoric GCC versions.

01micko:
I reckon keep the use of ps-FULL. That's the way the `/bin/ps` script is designed and busybox ps may indeed change over time. 

As for this pull request, I'm closing it, however, I did see #551 and agree that it could all be handled by one script with symlinks to the other callers.

PR #556, opened 2015-06-02
improve readability of svg_bar text

mrfricks:
adding 'bold' to the text line greatly improves the readability of text in pmount and other apps using svg_bar

### DIFF
--- a/woof-code/rootfs-skeleton/usr/lib/gtkdialog/svg_bar
+++ b/woof-code/rootfs-skeleton/usr/lib/gtkdialog/svg_bar
@@ -33,7 +33,7 @@ echo '<?xml version="1.1" encoding="UTF-8"?>
   <rect style="fill:url(#LG_02);" width="'$(($1-3))'" height="'$(($SVG_BAR_HEIGHT-3))'" x="1.5" y="1.5"/>
   <path style="fill:none;stroke:#ffffff;stroke-width:3" d="M 0,'$SVG_BAR_HEIGHT' '$1','$SVG_BAR_HEIGHT' M '$1',0 '$1','$SVG_BAR_HEIGHT'"/>
   <text
-	 style="fill:'$SVG_BAR_COLOR_TEXT';font-family:Mono;font-size:'$(($SVG_BAR_HEIGHT/2))';text-anchor:middle"
+	 style="fill:'$SVG_BAR_COLOR_TEXT';font-family:Mono:bold;font-size:'$(($SVG_BAR_HEIGHT/2))';text-anchor:middle"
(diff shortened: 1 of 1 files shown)


mavrothal:
This is Zigbert's code I believe, so he may want to pull or comment.
However, I do not think he is "watching" woof-CE now days so you may want to PM him if he does not respond after a couple of days. 
BWT do you have a link to "before and after" images?

mrfricks:
<a href="http://s1083.photobucket.com/user/MrFricks/media/svg_bar%20compre_zps1mafrbms.png.html" target="_blank"></a>

01micko:
Um, technically that has just introduced a bug. There is no "Mono:bold" ..
SVG 'styles' follow the CSS specification. Should be font-weight:bold;,
however you may have to change the 'font-family' to something else. As it
is is now it just defaults to the internal default font, and that code
could give unexpected results, possibly 'break-point trap' in gtkdialog.

mavrothal:
You are actually right. Just changing the font-family to "sans-serif" or "sans" or "Sans" without bold has the same result as Phil's change. "serif"  look equally small as "mono" and font-weight:bold improves readability but is still not nice. Is it ok to change to  "sans-serif" or "sans"?

01micko:
Hmmm.. I'm not exactly sure on how this is working. It seems to default to the default _system_ font (probably GTK as set in ~/.gtkrc-2.0 - note, I assume there are a lot of aliases for font names too), which is 'Sans' (font-family - derived from 'DejaVu Sans'). I suppose this makes sense since if SVG can't find a font then it has to use something. SVG will use whatever font-family you tell it to use. An easy way to find the ones on your system is to fire up Abiword and check them out. They all work, as long as you type the _full_ font-family name as displayed in Abiword. 

Fonts are still somewhat of a mystery to me. There is so much info/mis-info about them it is difficult to know what is correct. 

So, the upshot is that it is likely safe to go with `font-family:Sans;` in the code. The main bug in Phil's code is that it contains `font-family:Mono:bold;` which is a bug in the CSS3 spec (2 colons) and is the reason I said it could (and will at some point for somebody) cause "unexpected results".

01micko:
What about just increase font size? 
```font-family:Mono;font-size:'$((2 * $SVG_BAR_HEIGHT/3))```
partview

mavrothal:
I had tried that (as well as 3/5) but  in some cases the space is not enough to fit more than 14 characters (pmount).
Just changed it to "sans-serif" https://github.com/puppylinux-woof-CE/woof-CE/commit/c76d8c27f2b23cffcd90024648b052574720e97b

PR #559, opened 2015-06-23
Cleanup and fix a missed bit in the save-folder code in "init"

gyrog:
The cleanup commit simply changes the code to consistently use {} around variable names.
The fix commit relates to code that backs-out of updating a save-file/folder to do pupmode=5. This code was missed in all previous save-folder patches.


mavrothal:
or ```reset --hard <prev_hash>```   and recommit
BTW is safer to just ```rm "$CREATEPUPSAVE2FS"```. Ensures that only links will be removed

gyrog:
Here's the fixed version.
I think using an "if" statement is more obvious.

01micko:
If the condition succeeds then the variable is definitely there so no need to quote.

PR #561, opened 2015-06-24
savefolder - cleanup - quoted some more variables

gyrog:
I found some "if" statements that need quoting, so I quoted some "rm" statements as well.


mavrothal:
I hope is OK to delay merging few days till the fdrv branch is settle so we do not have two evolving branches.
Also please notice that the init used with the new style (huge) kernels is the ./woof-code/huge_extras/init, so please add relevant changes also there.
Thx

PR #574, opened 2015-09-22
fixes for perl_tiny & file templates

mrfricks:
added extra perl modules to perl_tiny template ...needed by pplog

'file' template altered , needed if you are using file & libmagic from ubuntu repos


mavrothal:
Would it be a problem if building from another distro?

mrfricks:
concerning perl_tiny, i had a look at  slacko & dpup iso's and slacko has overloading.pm & utf8.pm,  dpup has base.pm & utf8.pm so evidently they are needed.

with 'file'  with tahr64  i needed to use the version from the ubuntu repos.  most pups use a .pet but the option is in DISTRO_PKGS_SPECS to use the one from the repos.  the devuan build uses the repo packages and so would have the /usr/share/bug folder stripped  this might not matter,  but the  /usr/share/misc/magic symlink that is also stripped probably would matter,

mavrothal:
BTW, you should be able to pull yourself ;)

01micko:
@mrfricks did you get those 2 perl libs I mentioned in the tahr64 thread?
libintl-perl and one other.

mrfricks:
01micko  .... thanks, almost missed that ...added them to the devx

PR #584, opened 2015-10-03
add larger fontsizes for old eyes

mrfricks:
larger font sizes were requested in the tahrpup development thread for an older gentleman with poor eye sight

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/fontmanager
+++ b/woof-code/rootfs-skeleton/usr/sbin/fontmanager
@@ -65,6 +65,13 @@ S='
             <item>96</item>
             <item>102</item>
             <item>108</item>
+            <item>114</item>
+            <item>120</item>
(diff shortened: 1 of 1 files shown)


01micko:
Not being rude, but was it suggested to him to change screen resolution?

mrfricks:
that's one solution but you lose a lot of screen real estate.
surely the better solution is to use the fontmanager as that's what it's for

mavrothal:
Actually might be useful in more than poor sight. Xorg is bad in handling automatically modern high resolutions screens. 150 is good for a 200dpi screen (XO's for example) and for "retina" you need even more (FD701 is a pain in my macbook). I do not see any issue with the fontmanager UI either, so I would pull if no further arguments.

01micko:
Ok, tested it out and it works fine. Merging.

Issue #588, opened 2015-10-04
Wallpapers


01micko:
Barry has code in there that chops off the wallpapers at 8 images. The default images are rather aged (and most are ugly) too. I reckon we have no limit and only one default (a grey 1x1 px image) and that will do for minimalists. Then a builder can have their "wallpapers.xyz.pet" and not worry about hacking around later to fix them up.

mavrothal:
I would agree but add a "woof-woof" to it to "mark" the build sytem or "force" the change ;)

dimkr:
Yes! Delete all of them! They're so ugly I'm ashamed to have them in my puplets.

01micko:
NOTE: only fixed this for `3builddistro-Z` .. which may change soon :smile_cat:

dimkr:
We should merge 3builddistro and 3builddistro-Z. The difference should be 2-3 "if" statements IIRC.

01micko:
Maybe, but I just deleted and deleted when I altered it. I'm not sure that supporting the old initrd with modules is such a good idea. There were always lots of problems unless you followed Barry's kernel config implicitly.

mavrothal:
Now that we have the SVG backgrounds, is there any way to have ROX display an overlay or watermark on SVGs the way HTML/XML can? I would think JWM could handle it in XML but what about ROX.
The (obvious) idea hear is to have a puppy logo or the pupplet name/version showing over the solid colour backgrounds.

01micko:
>Now that we have the SVG backgrounds, is there any way to have ROX display an overlay or watermark on SVGs the way HTML/XML can?

I don't think so .. but .. I could write a small C utility that uses pango-cairo to render the DISTRO_FILE_PREFIX (or whatever) to a smallish image that should scale well. In fact I have already written something similar that was intended for labels, but could easily be adapted.

01micko:
Ok, take a look at this. The prog is called 'mkwallpaper'.

```C
/** This program is designed to be used in Puppy Linux.
 * It is designed to be used in Woof-CE or in a running Puppy as root or
 * an ordinary user.
 */
 
#include <cairo-svg.h>
#include <pango/pangocairo.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <unistd.h>
#include <sys/stat.h>

#define _GNU_SOURCE
#define PROG "mkwallpaper"
#define THIS_VERSION "0.1"

char out_file[128]; 

[...]

PR #589, opened 2015-10-04
Bye, wallpapers! (#588)

dimkr:
#588


01micko:
Almost solves #588 .. I'll have a look at the rest later (re: limit of 8)

Issue #570, opened 2015-07-05
merge2out & 0setup breaks templates for 64bit ubuntu builds


mrfricks:
currently if a 64bit build is chosen merge2out attemps to fix the templates (lines 139-157) moving all lib dirs and contents to lib64 and altering all FIXUPHACKs accordingly.

 however ubuntu has its 64bit libs in /usr/lib/x86_64-linux-gnu so most packages that go through the templates process get broken

mavrothal:
Instead of moving lib64 on a distro/version specific manner wouldn't it be more reasonable to symlink accordingly at the z_hack level?

01micko:
@mrfricks

Phil

Since it works in 32 bit with the `/usr/lib/i386-linux-gnu`, then how about
we skip that step in `merge2out`?

try @line139

`if [ "${COMPATDISTRO}"  = "slackware" -a "${TARGETARCH}" = "x86_64" ];then`

That should skip processing the templates.

01micko:
Just so you know, formatting markdown didn't work over email so skip the `backticks` :+1:

mrfricks:
0setup line 71 also needed your fix

01micko:
@mrfricks  I'll close this one but feel free to reopen if it is still an issue.

PR #593, opened 2015-10-07
01mickos fix needed to build 64bit ubuntu pups

mrfricks:
this was discussed here https://github.com/puppylinux-woof-CE/woof-CE/issues/570

### DIFF
--- a/merge2out
+++ b/merge2out
@@ -136,7 +136,7 @@ sync
 echo "Copying woof-arch/${TARGETARCH}/target/*"
 cp -a -f --remove-destination woof-arch/${TARGETARCH}/target/* ../woof-out_${HOSTARCH}_${TARGETARCH}_${COMPATDISTRO}_${COMPATVERSION}/
 sync
-if [ "${TARGETARCH}" = "x86_64" ];then
+if [ "${COMPATDISTRO}" = "slackware" -a "${TARGETARCH}" = "x86_64" ];then
  # fix up the templates for 64 bit build
  TEMPLATES="../woof-out_${HOSTARCH}_${TARGETARCH}_${COMPATDISTRO}_${COMPATVERSION}/packages-templates"
  LIST=`ls ${TEMPLATES}|grep -v pcmciautils`
--- a/woof-code/0setup
+++ b/woof-code/0setup
@@ -68,7 +68,7 @@ fi
  RUNNINGPUP='no'
 # 140109 64 bit support #140116 improve
 . ./WOOFMERGEVARS
-if [ "$WOOF_TARGETARCH" = "x86_64" ];then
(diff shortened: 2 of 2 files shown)


mavrothal:
Noticed that in #594 that your build failed. Is it related to this one or something entirely different? ie do the original templates work fine in 64bit ubuntu or should be modified to get the libs in  /usr/lib/x86_64-linux-gnu?

01micko:
@mavrothal we are using they symlink in debian/ubuntu builds so /lib64, /usr/lib64 technically shouldn't exist in these builds. It's up to @mrfricks if he wants to go the whole hog and change libdirs to  /usr/lib/x86_64-linux-gnu or /usr/lib/i386-linux-gnu, but I think @dimkr might want some input on that.

mavrothal:
@01micko so do we pull or not?

dimkr:
@01micko I think /lib and /usr/lib should be links to /lib64 and /usr/lib64, respectively, with x864-linux-gnu and i386-linux-gnu as links to .., so stuff like ldconfig doesn't get messed up. Otherwise, many package templates (FIXUPHACK and pinstall.sh stuff) need to be fixed.

01micko:
Will the idea of this pull request is such that it doesn't need to touch the FIXUPHACKS. No harm in having the {/usr}/lib64 link though.

I vote pull it!

01micko:
And by the way, I prefer not to have lib64 links for slacko purposes.

For best compat with upstream it would be better if they are real directories, but whatever. I'll adapt to whatever the flow is. :v:

mavrothal:
a) You can pull too ;)
b) why not get a 'case' statement and change dirs according to the $binarycompat instead of links and horseshoe'ing?

01micko:
That doesn't fix the hard coded templates though.

For now I'll go with a)

mavrothal:
> That doesn't fix the hard coded templates though.

But couldn't the ```case``` call a function to ```sed``` through them as needed?

peabee:
I believe that there may be problems with sfs-load which try to
install contents into lib64 if it is a link..... 

For instance the
Chromium64 sfs will not install without amendment onto Tahrpup64 which
has lib64 as a link 

Cheers 

peebee 

On 08/10/2015 9:16, Mick Amadio
wrote: 

> And by the way, I prefer not to have lib64 links for slacko
purposes. 
> 
> For best compat with upstream it would be better if they
are real directories, but whatever. I'll adapt to whatever the flow is.

> 
> --
> Reply to this email directly or view it on GitHub [1].

--

PB.............
 

Links:
------
[1]
https://github.com/puppylinux-woof-CE/woof-CE/pull/593#issuecomment-146452758

PR #595, opened 2015-10-08
bring back puppy.css ...lost earlier this year

mrfricks:
puppy.css was lost with this commit https://github.com/puppylinux-woof-CE/woof-CE/pull/517
the picture doesn't do it justice, but the home and the help page look a lot nicer with it

<a href="http://s1083.photobucket.com/user/MrFricks/media/snapshot-2015-10-08-10-23-38_zpsy3wqpr7f.png.html" target="_blank"></a>

### DIFF
--- a/woof-code/rootfs-skeleton/usr/share/doc/forums.htm
+++ b/woof-code/rootfs-skeleton/usr/share/doc/forums.htm
@@ -5,6 +5,7 @@
 	<meta charset="utf-8">
 	<link href="./manual.css" rel="stylesheet" type="text/css">
 	<link href="./forums.css" rel="stylesheet" type="text/css">
+	<link href="./puppy.css" rel="stylesheet" type="text/css" />
 </head>
 <body>
 	<div id="container">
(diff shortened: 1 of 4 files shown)


mavrothal:
I believe that the whole idea of this was to have unified css instead a multitude of styles depending on the page. If we do not like the specific css we can change it to something better but do we really want to bring back this "polymorphism"?

mavrothal:
Oh, and when it comes to aesthetics, do not forget the power of "familiarity" ;)

mrfricks:
i've been familiar with both...  and the change from a plain background to the pup logos was like "hmmm very nice and slick.... much improved" 
and the change back to a plain background felt more like someone had just come along and popped my balloon :)

mavrothal:
I see. Hopefully you are not going to tell me that you also like BK's Green-Lime themes :D
I think the new style is more crisp and unassuming, (bordering boring ;) but we just try to get some info out and focus user attention on the info primarily. As I said, it could be improved but let's try to keep it unified and "clean".

01micko:
Sorry.. balloon is popped :sob:

mrfricks:
fair enough......  r. i. p.  puppy. css

PR #616, opened 2015-10-26
Extra privacy

dimkr:
I don't like the fact Puppy "phones home", for two reasons:
- Security - if you have the ability to capture the traffic of a Linux machine, it's easy to identify Puppy. As fingerprinting is one of the first steps an attacker performs (for example, when trying to find an exploit that works against a remote Linux machine), we want to make it harder. However, this ping call is our way to test a network connection, so we have to keep it.
- Privacy - using a different site (that is known to respect the user's privacy, unlike Google) is all we can do to improve privacy in this respect.

@01micko @mavrothal Is there any additional anti-privacy feature that most users don't use? If yes, we should disable it by default.

### DIFF
--- a/woof-code/rootfs-skeleton/usr/bin/pman
+++ b/woof-code/rootfs-skeleton/usr/bin/pman
@@ -176,6 +176,6 @@ fi
 #exec defaulthtmlviewer http://threads.seas.gwu.edu/cgi-bin/man2web?program=${NAME}
 #v431 fix thanks to technosaurus...
 [ "${SECTION}" != "" ] && exec $HTMLVIEWER "http://linux.die.net/man/${SECTION}/${NAME}"
-exec $HTMLVIEWER "http://www.google.com/search?&q=man+\"${NAME}\"+site:linux.die.net&btnI=Search"
+exec $HTMLVIEWER "http://www.duckduckgo.com/?q=man+\"${NAME}\"+site:linux.die.net"
  
 ###END###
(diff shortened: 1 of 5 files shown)


01micko:
I'll merge it. Lately though, I have been using `wget --spider`

dimkr:
Oh crap, welcome1stboot.bac pings sourceforge.net. Needs to be edited and recompiled :(

01micko:
Ugh. Bacon. Can we leave it until we switch to mdview?

dimkr:
Sure. That gives me enough time to conduct an experiment: connect a machine running a live Librepup directly to my laptop (with 1 in /proc/sys/net/ipv4/ip_forward), so I can capture all outgoing traffic to detect other annoyances I missed :)

01micko:
I can do it. Just compiled. Damn.. cant attach a .tar... 

see.. http://www.murga-linux.com/puppy/viewtopic.php?p=869527#869527

01micko:
Woops, includes the source, 32 bit - static hug

01micko:
I removed the murga post, as I missed a ping anyway. Fixed and committed with https://github.com/puppylinux-woof-CE/woof-CE/commit/5732467783d736a7cefa995dbe99828aadebf1ac

mavrothal:
When I run the  puppies  from VMs  I monitor the outgoing traffic. Comparing to the browsers, puppy calling is totally benign. Firefox has become particularly offensive lately. 
Regarding search engines, [startpage] (https://startpage.com/eng/protect-privacy.html?&hmb=1) offers google quality with high privacy.
```https://startpage.com/do/dsearch?query=man+\"${NAME|}"+site:linux.die.net``` works fine.
Having said all these, I believe that is very little you can do if targeted by professionals and obscurity plus the absence of any services in puppy, protects you from "hobby intruders".
As far as fingerprinting goes, try for [yourself] (https://panopticlick.eff.org/) and see.
But there is nothing wrong with the changes.

Issue #607, opened 2015-10-18
Remove unionfs


01micko:
Union fs has been a dead project for a number of years. It is about time we removed all reference to it in `init` and userspace.

PR #617, opened 2015-10-28
Removed Unionfs support (#607)

dimkr:


01micko:
Looks pretty good to me. Mergeing. Any bugs (I doubt it) we should be able to sort before any release.

PR #618, opened 2015-10-29
Add getflash in woof-code/rootfs-packages

ASRIedu:
getflash v1.5.7


01micko:
Thanks!

ASRIedu:
Hello micko !
The few times I used the git was to modify existing files. For this, I used the web interface.

To respond to your proposal, I tried to add the files of the getflash utility.
I hope this is okay, but I let you check the available files. For my part, I will propose a package on the forum.

Regards

01micko:
It's ok :smile: . I will give it a full test on the week end and notify you if there is problems.

Cheers!

mavrothal:
Is missing the relevant entry in 3bilddistro-Z I think

01micko:
What entry? Should get filtered with this: ( @ line 322 )

```shell
			if grep "$d" DISTRO_PKGS_SPECS-${DISTRO_BINARY_COMPAT}-${DISTRO_COMPAT_VERSION} |\
			grep -q '^yes' ;then
				state=false # don't overwrite user chosen specs
				def=false
			fi
```

mavrothal:
I meant to set it true in line 307. But I guess is up to the builder to include it.
BTW Shino has sfs_load 2.3.7 out.

01micko:
That's what the line does if the entry is not part of the specs :wink:

mavrothal:
I think the current approach in puppy is that translations are going in a separate pet, one for each language and are not part of woof/puppy, to keep size low.
I believe that sooner rather than latter will need a new directory (woof-nls?) where pot/po/mo files for all the rootfs-{skeleton,packages} scripts should be kept. The pot files could be going with devx while the others should be packed as pets per language. 3buiddistro should be modified accordingly.
Maybe open an issue to discuss who could we best proceed, hopefully with some input for the experience translators of the puppy community.
For now though, I would say keep it out.

ASRIedu:
@ mavrothal

Thank you for your explanations.
So, I keep the translation out.

Eventually, I will restart the git intended for translations https://github.com/puppylinux-translations/woof-CE

Regards

01micko:
@ASRIedu this is failing because the pinstall doesn't work in woof building!

The problem is that you need an absolute path to create the prefs file.

By the way... `/usr/sbin` is not a nice place for config files. Config files that are non-editable (such as `getflash-conf`) should go to `/etc` and config files that are written to (such as `getflash-prefs`) should go to `$HOME` .. which in default puppy is `/root`

Don't worry about @zigberts 's (old? - actually I believe it was NathanF's :wink: ) method in the top of the script. Just define the path to conf  (in `/etc`) and prefs (in `$HOME`) instead. Alternatively you can put them both in `$HOME` but perhaps a hidden subdirectory is best for this, such as `$HOME/.getflash`. You see if we ever go multiuser this simplifies things.. and apart from that only packages should write to the bin - sbin dirs.

As for the `pinstall.sh`, you can easy use a _relative_ path.

An example.

```shell
if  [ `pwd` != '/' ];then # note NOT equals
    echo "this goes to the woof built system" > root/somefile.txt # note no leading slash
else
    echo "this goes to running system" > /root/somefile.txt
fi
```

HTH ! :smile:

mavrothal:
And while at it, I'm not sure why we need to run an update check (and a connection to adobe) by default.
The default should be off and user activated as an option.

01micko:
>The default should be off and user activated as an option.

Yes.

mavrothal:
also when you fix the app and the pinstall paths, make sure you add a removal step for old config/prefs in /usr/sbin, in pinstall in case of updates.

ASRIedu:
Thank you for the improvements.
Note: I just change jobs. Therefore, in the coming weeks, I will have very little free time. So I let you make the necessary changes.
Regards,

01micko:
No problem @ASRIedu . Good luck with the new job!

PR #630, opened 2015-11-07
Add ROX pin button and tooltips in icons options

mavrothal:
Add a button in the pdesktop "Desktop icons" options to change the ROX pin.
I find it handy, but is Zig's domain ;)


mavrothal:
This is how pdesktop will look after clicking on the "Desktop Icons" button.

pdesktop

01micko:
It's a nice idea.. and I'm sure @zigbert can fix the button widths to be consistent in short order.

I did see a typo in the code.. "who" should be "how" iirc.

mavrothal:
Actually button widths where variable before too.
Regarding typos, is the reason that for years I avoided the terminal... :-(

zigbert:
What is the newbie-friendly syntax for ROX-pin. Haha, what a word that is... :-D

mavrothal:
> What is the newbie-friendly syntax for ROX-pin. Haha, what a word that is... :-D

Does this implies that you do not *want* a button to change the assortment of icons on the desktop or that you find the name inappropriate (and the tool tip inadequate)?

01micko:
@mavrothal I don't think so!

> What is the newbie-friendly syntax for ROX-pin. Haha, what a word that is... :-D

Maybe 'desktop layout' ? :confused:

zigbert:
:)
I just think our use of technical words is user hostile.
desktop layout is sure better than ROX-pin - but far from descriptive of which icons are shown on desktop and their alignment/placement.

01micko:
Is difficult to sum up in a very short string so if a descriptive tooltip is used then hopefully it reduces some pain... :grin:

mavrothal:
There are 2 issues mixed. a) do we want a button to choose pin and b) how can it look good in the UI.
If we want the button then we need to have fixed width buttons that can take more than a time of text or (better yet I thing) have indicative names and a good tooltip text to explain what is really doing.
The whole thing refers to desktop icons anyway so "Add app" "Drives" "Arrangement" and "Theme" should be OK as labels and still do not mess up the UI
pdesktop2

zigbert:
I like this one!

If we want longer button labels, just use 2 rows with 2 buttons each. If there are more humans like me, tooltips are not read... But I think this will work just great.

01micko:
Who is on first? How? :laughing: (re same typo)

mavrothal:
Old picture. Check the [commit] (https://github.com/mavrothal/woof-CE/commit/0addbdd77ac6f89f1d3d4050c4a423e1f86b2e38) ;)

01micko:
Yes, already commented on commit :+1:

PR #632, opened 2015-11-08
Flashfix branch to fix bugs in getflash

01micko:
Hopefully fixes up some bugs in getflash


mavrothal:
The fix is fine but the pull is not since it has fixes from testing back to testing that messes up history which means is more difficult to know which changes were done together and in what order.

01micko:
Ok, I'll kill it and just commit to testing.

Issue #631, opened 2015-11-07
some jwmconfig/ptheme issues


mavrothal:
- Left mouse click brings up the "activate mouse action" banner when clicking in an empty area on the desktop. Besides annoying and with no way to eliminate it, it interferes with the next left click as it used to close the banner and not the function intended (ie close a window, open an app etc).
- Since we are going with the minimal desktop icons by default, the top tray should  have autohide=off as a default, so a first time user can see what is available instead of going through the long menus.
- "Connect" should be included in the top tray as more puppians use it in laptops with wifi.
- Top tray has pMusic. Should be default audio/video player

mavrothal:
See also [here] (http://www.murga-linux.com/puppy/viewtopic.php?p=871482#871482) for an error due to the unknown original position of the top tray

BTW the 'left mouse click brings up the "activate mouse action" banner' above goes away if you select "blank" from the mouse action menu and restart X. However "blank" should be the default for left click

zigbert:
- Left mouse brings up "activate mouse actions":
This is as intended, and I think still this is how it should work.
It is nice to show new feature, and it is not in the users way if not having an unintentionally-clicking-habbit. If so, it can easily be deactivated.
- Toptray to be autohided:
Disagree. Let the desktop be clean. Our mission is to help the user fill the desktop - not actually fill it.
- Include connect icon:
Hmmm - what a minor issue :-) The setup icon is there...
- Change pmusic icon:
I don't know - maybe skip it instead. Mime settings should handle multimedia files....

01micko:
> - Change pmusic icon: I don't know - maybe skip it instead. Mime settings should handle multimedia files....

Yes skip pburn and defaultmediaplayer too and maybe add the 'i' icon for help?

zigbert:
The 'problem' with the info-button, is that it does not has a *.desktop file - so jwm_config/tray is not capable to set this. - It has to be edited manually, which is not a problem at all - just a matter of principles. 

defaulthtmlviewer file:///usr/share/doc/index.html

01micko:
Hmm I see. Hard coding doesn't allow Icon theme change. 

I forget the layout off hand, but does it have defaultpaint?

mavrothal:
> Left mouse brings up "activate mouse actions": This is as intended, and I think still this is how it should work. It is nice to show new feature, and it is not in the users way if not having an unintentionally-clicking-habbit.

[...]

PR #633, opened 2015-11-10
Set top tray to visible and bellow but above when hidden 

mavrothal:
As discussed [here] (https://github.com/puppylinux-woof-CE/woof-CE/issues/631#issuecomment-155274133 ), within #631


zigbert:
I don't like the idea to set the default toptray to visible without offering a mirrored autohidden toptray in the 'above' layer. I think this is _maybe_ a target for next generation of tray management.

Hopefully, Mick will use the Buntoo theme as default.
After all, how to set up trays in the next Puppy is up to him.

mavrothal:
Well, I do not know about the double tray, and we can keep discussing about visible as default, but I think setting a hidden tray "above" no matter what the user has chosen (by mistake/misunderstanding?) is a good idea.

zigbert:
Not sure...
In DuDE ( http://www.murga-linux.com/puppy/viewtopic.php?t=53680 ), I used trays in 'below' layer in combination with autohide. So I know there exist times when this combination is valid. I don't think it is smart to override the user choice.

mavrothal:
> In DuDE ( http://www.murga-linux.com/puppy/viewtopic.php?t=53680 ), I used trays in 'below' layer in combination with autohide. So I know there exist times when this combination is valid.

Not to be picky but what is it telling us that is valid other than you used it in some spin?
More important what would be  a "use case" where you want a hidden tray to remain hidden behind other things while you are trying to access it?

mavrothal:
> After all, how to set up trays in the next Puppy is up to him.

OK then. Let's give the builder a choice though.
What about this patch?
```
diff --git a/woof-code/rootfs-packages/ptheme/pinstall.sh b/woof-code/rootfs-packages/ptheme/pinstall.sh
index 09b73c8..a26cea0 100755
--- a/woof-code/rootfs-packages/ptheme/pinstall.sh
+++ b/woof-code/rootfs-packages/ptheme/pinstall.sh
@@ -60,6 +60,28 @@ cp -af usr/share/jwm/themes/"${PTHEME_JWM_COLOR}-jwmrc" root/.jwm/jwmrc-theme
 cp -af usr/share/jwm/themes/"${PTHEME_JWM_COLOR}-colors" root/.jwm/jwm_colors
 echo "jwm colors: ${PTHEME_JWM_COLOR}"
 
+HIDETOP=$(grep autohide usr/share/jwm/tray_templates/"$PTHEME_JWM_TRAY"/jwmrc-tray2 \
+ | cut -f2 -d ' ' | cut -f2 -d '"')
+if [ "$HIDETOP" != "off" ]; then
+ echo Do you want the "$HIDETOP"-side application tray to be hidden on first boot? [Y/n]
+ read hide
+ if [ "$hide" = "n" ];then
+  sed -i "s/$HIDETOP/off/" \
+   usr/share/jwm/tray_templates/"$PTHEME_JWM_TRAY"/jwmrc-tray2
+ else
+  BELOW=$(grep layer=\"below\" usr/share/jwm/tray_templates/"$PTHEME_JWM_TRAY"/jwmrc-tray2)
+  if [ "$BELOW" != "" ]; then
+   echo The hidden "$HIDETOP"-side tray will appear bellow any open
+   echo windows that may cover that area. Do you want it to appear 
+   echo above them instead? [y/N]
+   read above
+   if [ "$above" = "y" ]; then
+    sed -i 's/below/above/' usr/share/jwm/tray_templates/"$PTHEME_JWM_TRAY"/jwmrc-tray2
+   fi
+  fi
+ fi
+fi
+
 cp -f usr/share/jwm/tray_templates/"$PTHEME_JWM_TRAY"/jwmrc-tray* root/.jwm/
 echo "jwm tray: ${PTHEME_JWM_TRAY}"
 
```

zigbert:
...You really want the toptray to be visible :)
What if tray2 is not the chosen template for the 'Extra' tray? Why not trust the theme-designers to make worthy desktops?

Why not do as Mick, build one hell better theme that Puppy builders would prefer? You are then in the position to decide every aspect of the theme. Not only if the top-tray should be

mavrothal:
...You really do not want the themes touched ;)
Why force the builder rediscover the wheel just to change an option or do not see as possibility that the weight of form and function might be different between designer and builder? The defaults are what the designer sets anyway.
BTW there is already a [suggestion] (http://www.murga-linux.com/puppy/viewtopic.php?p=871537#871537) for visible tray

01micko:
Maybe we close this pull request and pull the 'Buntoo' and 'pop' themes in `rootfs-packages`?

mavrothal:
I do not really mind but give it some time to see if any feedback, though both slacko and slacko64 circumvent the issue :sunglasses: 
But regardless of defaults and the specific pull, what about config options at build time as suggested [above] (https://github.com/puppylinux-woof-CE/woof-CE/pull/633#issuecomment-1556804320)

01micko:
Regardless.. 'pop' and 'buntoo' have had a good reception so maybe I'll put them in there anyway, but just lighten buntoo trays a bit so the default icons are more visible.

01micko:
Just thinking on this a little further... 

It's not logical/moral/ethical/whatever to alter another designer's UI choice. They designed with specifics in mind for whatever reason. The themes are not big so why not add your own? I know it may be duplication but really it's all there already so it is not much, just the JWM stuff. That way, if a user has a tray sitting there that they don't like they can get a parallel version with a hiding tray. Actually it's probably less than the old way of shipping a bunch of jwm_theme_XXX's. Side note: I'm not a fan of the 'pop' theme - it was simply put there to get the iso out while this discussion was pending.

It's difficult to say what a user wants. Some like a bunch of icons. Some like separate launchers. Some like a clean desktop. Some hate JWM!

Personally, I like the Unity UI, until you open a window and the window stuff goes to the task bar ala Mac. They obviously mimic a lot of Mac behaviour but make it different enough to avoid litigation! I suppose this has something to do with 'convergence'.

I also am used to KDE (was my first Linux experience) and it is known that windows has pilfered a lot of ideas from KDE.

Also, JWM is evolving rapidly and if a new stable version is released soon everything here will be broken anyway! So version testing should be of a reasonably high priority.

2¢

mavrothal:
> It's not logical/moral/ethical/whatever to alter another designer's UI choice.

Wow. That is heavy!
I hope we agree that all themes are subjected to the configuration options. 
If so the why is logical/moral/ethical for the user to change these but not for the puppy builder? Why the user can have a choice and the builder that choses the specific theme for whatever reason, can not? 
Are we really claiming that in the FOSS/GPL world the code or the design is immutable because of ethics or otherwise? What if the builder, since we do not provide an option to modify some defaults, resorts to the all familiar z_hacks? Still immoral?
Wow again!

But beyond "ethics", I honestly have not seen a (mainstream linux, or otherwise) OS that comes with hidden by default desktop elements. Can someone point to an OS where a permanent desktop element like a tray, dock, wbar, etc can only be found either serendipitously or by RTFM? Even smartphones with practically no real estate provide some hints that "more is there". Few dots, a hint of a tray, something... I believe is written in HUG (though I could not find it again).

Anyway, It looks to me that we give no choice to the builder, for "designer's copyright" reasons! Hopefully this will not extend to the user too :tongue:

01micko:
>Wow. That is heavy!

Blame the :beers: .. it was late and I couldn't find the right adjective, hence ".../.../.../...". Thankfully I am better with file descriptors! 

>Anyway, It looks to me that we give no choice to the builder, for "designer's copyright" reasons! Hopefully this will not extend to the user too :tongue:

Opposite! Choice is what it is all about! Designer choice, Builder choice, User choice!

I know it's hard to strike a happy medium (like buntoo not working in [RHS](http://www.murga-linux.com/puppy/viewtopic.php?p=873058#873058)) but I gave it a shot and has had a good reception otherwise. 

The upshot I believe is more themes! :+1:  ( @zigbert when you get some time :wink: )

We have to remember PTheme is brand new and is now out there in the wild. I just hope we get decent feedback to fix any issues.

zigbert:
Looking at it...

zigbert:
I suggest Buntoo rearrange itself.
The side-tray should not be defined by x/y. Instead use relative positioning.
Switch tray1 and tray2 so the side-tray is below top-tray. This should make the theme compatible with jwm_config/tray. At least after my fix for the headerlines <include> and <command>.

The Buntoo tray files should also go into /root in case jwm fails and use the builtin templates. For 6.3.0 it returns to stardust_dark which is the default in woof.

A new side-tray rc

```
<?xml version="1.0"?>
<JWM>

<Tray autohide="left" insert="right" layer="above" halign="left" valign="top" width="48" layout="vertical" >
    <Spacer width="30" height="40"/>
    <TrayButton popup="Rox File Manager" border="false" icon="/usr/local/lib/X11/pixmaps/home48.png">exec:rox</TrayButton>
    <TrayButton popup="Web Browser" border="false" icon="/usr/local/lib/X11/pixmaps/www48.png">exec:defaultbrowser</TrayButton>
    <TrayButton popup="Chat" border="false" icon="/usr/local/lib/X11/pixmaps/chat48.png">exec:defaultchat</TrayButton>
    <TrayButton popup="Urxvt terminal emulator" border="false" icon="/usr/local/lib/X11/pixmaps/console48.png">exec:urxvt</TrayButton>
    <TrayButton popup="Pprocess process manager" border="false" icon="/usr/local/lib/X11/pixmaps/pc48.png">exec:pprocess</TrayButton>
    <TrayButton popup="Abiword word processor" border="false" icon="/usr/local/lib/X11/pixmaps/word48.png">exec:abiword</TrayButton>
    <TrayButton popup="Gnumeric Spread Sheets" border="false" icon="/usr/local/lib/X11/pixmaps/spread48.png">exec:gnumeric</TrayButton>
    <TrayButton popup="MTPaint paint program" border="false" icon="/usr/local/lib/X11/pixmaps/paint48.png">exec:mtpaint</TrayButton>
    <TrayButton popup="Geany Text Editor" border="false" icon="/usr/local/lib/X11/pixmaps/edit48.png">exec:geany</TrayButton>
    <Spacer width="30" height="30"></Spacer>
    <TrayButton popup="Puppy Setup" border="false" icon="/usr/local/lib/X11/pixmaps/configuration48.png">exec:wizardwizard</TrayButton>
    <TrayButton popup="Puppy Package Manager and Universal Installer" border="false" icon="/usr/local/lib/X11/pixmaps/pet48.png">exec:dotpup</TrayButton>
    <TrayButton popup="Edit tray" border="false" icon="/usr/local/lib/X11/pixmaps/utility48.png">exec:/usr/local/jwm_config/tray</TrayButton>
    <Spacer width="30" height="2000"></Spacer>
</Tray>
</JWM>
```

zigbert:
Buntoo includes the jwmrc-colors file which not longer are required in a Woof-built Puppy

01micko:
>The Buntoo tray files should also go into /root in case jwm fails and use the builtin templates. For 6.3.0 it returns to stardust_dark which is the default in woof.

@zigbert - no. What should happen is that at install time in woof they get copied to the backup so that the original shipped theme becomes the default. I'll look at the pinstall.sh that installs the default theme.

As a favour, and since it sits on your hard drive, can you commit the modified 'Buntoo' theme to rootfs-packages when you get the time? I'll be sure to delete it from slacko specs (when done) as it will get picked up anyway.

zigbert:
I can upload Buntoo...

PR #643, opened 2015-11-24
/usr/bin/pman improvements

wdlkmpx:
* I added support for groff [console and graphical environment], make default if available

* Workaround: depending on the program, pman may delete the generated html page before viewing,
 see here [url]http://murga-linux.com/puppy/viewtopic.php?t=99634[/url]

* Various other edits


mavrothal:
I'll pull it this time but in the future please give a better commit message title. 
The wild card does not make any sense and usually the title is in present tense. ie "Add support for groff...." (50 characters max).
Also since the changes including the author and date  are now indicated in the git log we do not add them at the script header.

PR #645, opened 2015-11-25
xterm: use bash specific syntax

wdlkmpx:


wdlkmpx:
This looks ok to me, BUT

  EXECCOMMAND=${EXECCOMMAND##*-e }
 
can be changed to

  EXECCOMMAND=${EXECCOMMAND#*-e }
  
As ${EXECCOMMAND#*-e } will replace only the first match 'xterm -hold -e XXXX -e' > XXXX -e

PR #646, opened 2015-11-25
xterm: use bash specific syntax

wdlkmpx:


wdlkmpx:
hmmm something's not right

PR #647, opened 2015-11-25
xterm: user bash specific syntax

wdlkmpx:
This speeds up xterm by replacing external programs with bash builtin functions.

I know I've been playing in git with this script, but I needed to make sure this is ok,
but now I can attest that this works the same as before

### DIFF
--- a/woof-code/rootfs-skeleton/usr/bin/xterm
+++ b/woof-code/rootfs-skeleton/usr/bin/xterm
@@ -1,19 +1,22 @@
-#!/bin/sh
+#!/bin/bash
 #rxvt does not understand the '-hold' option.
 #urxvt 20120214
 
-if [ "`echo -n ${*} | grep '\-hold '`" != "" ];then
- EXECCOMMAND="`echo -n ${*} | grep -o ' \-e .*' | sed -e 's/ \-e //'`"
- if [ "$EXECCOMMAND" != "" ];then
-  echo '#!/bin/sh' > /tmp/xterm_simulate_hold.sh
-  echo "$EXECCOMMAND"  >> /tmp/xterm_simulate_hold.sh
-  echo 'echo' >> /tmp/xterm_simulate_hold.sh
-  echo 'echo -n "FINISHED. PRESS ENTER KEY TO CLOSE THIS WINDOW: "' >> /tmp/xterm_simulate_hold.sh
-  echo 'read simuldone' >> /tmp/xterm_simulate_hold.sh
-  chmod +x /tmp/xterm_simulate_hold.sh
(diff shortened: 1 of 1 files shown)


mavrothal:
I have no problem with the change, I just wonder what problem does it solve? What is the speed impact? Could you give an example of the effect before and after?

wdlkmpx:
Actually it could even be slower hahaha. I guess it just qualifies as a coding style change

But something like this edited new2dir is like 15-20 times faster
http://pastebin.com/dZiWkDc5

mavrothal:
Interesting. Do you have any numbers?
I tried to run it on a dir (as it says it can do) and it totally failed.

mavrothal:
Also tried it after a compile. Looks OK but if you call dir2pet at the end it quits after the first pet regardless if you split the files in doc, dev, exe etc.

wdlkmpx:
It's not finished, but you can defintely compile with it. I lost interest some months ago. Just try something like 

./configure
make
new2dir release

I have 2 companion scripts
- compile
- configure

With these 3 I usually only give one URL or local file to 'compile' and it does the rest

wdlkmpx:
You have to compile something that installs hundreds or thousands of files, there you can compare how fast ca be, in my experience with this almost PentiumIII, the difference is insane

01micko:
Regarding `new2dir` .. mabe you should apply the magic to [petbuilds}(https://github.com/puppylinux-woof-CE/petbuilds)

Also, installwatch is unmaintained and only a matter of time before it dies from bitrot. `find` can do a more than adequate job. Plus nearly all souces support  `DESTDIR` so perhaps testing the Makefile(s) for this is a good idea and will speed stuff up considerably.

wdlkmpx:
I always use tar, I forgot to test the final part properly

UHmm micko I have my own "build system", take a look
http://pastebin.com/KMBFbLs2

It can be updated and fixed, you think it can go there?

01micko:
>UHmm micko I have my own "build system", take a look
http://pastebin.com/KMBFbLs2

Looks nice. I might cherry pick some ideas if you don't mind.

wdlkmpx:
There other scripts you might find useful:

'configure'
http://pastebin.com/ezBPfzpk

'whichgcc'
http://pastebin.com/dGFDtcst

qt4 (qt4 environment (/opt/4.8))
http://pastebin.com/74csS0gr
(this is to compile and run qt4 apps, you con adapt it to compile qt5 apps)


This is an example of pkgs.sh , the generic script used by 'compile', located in $PROJECT_DIR 
http://pastebin.com/1qBtQ3am

PR #656, opened 2015-12-03
Merge restartwm, wmpoweroff, wmreboot & wmexit into one script	

wdlkmpx:


wdlkmpx:
I attach a file so you can test these changes
# tar --directory=/ -vzxf wmexit.tar.gz 

[wmexit.tar.gz](https://github.com/puppylinux-woof-CE/woof-CE/files/51516/wmexit.tar.gz)

wdlkmpx:
It's because that will require editing several other files in and out of woof

It can be done though.. the code already contains a way to trigger the correct part by giving an argument

PR #665, opened 2015-12-09
Add puppy mime packages

wdlkmpx:
Two files:
custom.xml
puppy.xml

custom.xml: some mimetypes I myself added...

puppy.xml is currently in shared-mime-info.pet and must be removed from there.

    /usr/share/mime/packages
     freedesktop.org.xml
     puppy.xml

The shared-mime-info.pet contains also a processed [outdated] freedesktop.org.xml (already split into many files)
Should contain only 3 files

    /usr/bin/update-mime-database
    /usr/share/mime/packages/freedesktop.org.xml
    /usr/share/pkgconfig/shared-mime-info.pc

And the post install script should process /usr/share/mime/packages

    update-mime-database /usr/share/mime

In fact, only /usr/bin/update-mime-database should be in shared-mime-info.pet

### DIFF
--- a/woof-code/rootfs-skeleton/usr/share/mime/packages/custom.xml
+++ b/woof-code/rootfs-skeleton/usr/share/mime/packages/custom.xml
@@ -1,0 +1,82 @@
+<?xml version="1.0"?>
+
+<mime-info xmlns='http://www.freedesktop.org/standards/shared-mime-info'>
+
+  <mime-type type="application/x-bin">
+    <comment>Binary File</comment>
+    <glob pattern="*.bin"/>
+  </mime-type>
+
+  <mime-type type="application/x-dll">
+    <comment>MS Windows DLL</comment>
+    <glob pattern="*.dll"/>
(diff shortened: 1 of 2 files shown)


mavrothal:
Why not just update the shared-mime-info.pet instead.
BTW why "custom" should be by default in puppies?...

wdlkmpx:
It can be updated, I would compile it in Wary, but then I'd have to modify.  

My custom.xml is like an added value..

I just realized there is a puppy.xml in woof-code/packages-templates/shared-mime-info/usr/share/mime/packages

Should I move it to there?

mavrothal:
> My custom.xml is like an added value..

I personally do not see much value in all these windows/game mimes to be in by default when no default app support them. I think these mime types should be added by packages that can handle these extensions rather than by default.

Regarding woof-code/packages-templates/shared-mime-info/usr/share/mime/packages/puppy.xml, indeed this would be the file to edit if needed.

Issue #667, opened 2015-12-10
Misc Issues that probably need addressing


starhawk64:
starhawk from the Murga Puppy Forum here :) Bringing up some issues I've experienced... consider each of these offerings as a note in the Suggestion Box.

(1) savefiles corrupt spontaneously, due to (I suspect) improper unmount at shutdown. fsck at bootup slows this process down but does not solve it completely.

(2) many applications in Puppy have poorly written English in them -- the grub4dos wizard comes to mind here ("Press OK to install. Press Cancel if you do not want to." ;) ).

(3) there is no command-line package installer... and when I showed a Linux Guru friend of mine the part of petget that validates dotPET packages, he very nearly lost both lunch and sanity. I'd asked him to explain to me what it was doing where. He gave up completely midway through. IIRC it was MD5sum-ing two MD5sums and comparing them, about where he stopped... and there was at least one case where a command was routed to a binary that was POSIX-noncompliant -- I want to say that it had the name of a POSIX binary/command but was a Puppy-specific implementation.

(4) we should probably have an archaic-machines branch specifically for dusty old relics brought in by folks who simply don't know any better. No offense to ttuuxxx and goingnuts -- but ClassicPup and pUPnGO are simply not practical offerings for these machines (5x86-class CPUs, max. 256mb RAM). There should be a Pup for these machines... we kind of are the only game in town for them. (DSL is not a serious offering any more, given how outdated its everything is, and Knoppix AFAIK is dead and gone.)

mavrothal:
Welcome.
File corruption may be because of a faulty drive rather than improper shutdown. Fsck can not do miracles in this case though ext4 fs may help a bit.
Regarding english, please you or others, edit as needed.  It does not even need any coding knowledge. Just keep in mind that many apps (like grub4dos) come from pets and are not in woof-CE. Offer your help to the pet maintainer in these cases.
Tell your friend to send some patches over for petget :wink: 
Archaic branch may be a good idea, however I do not think that any Pentium III, 256MB machine can handle a modern browser with any tolerable responsiveness. So I'm not sure that anything "better" than puppy2 can run on this kind of machines (that is the reason I stopped making puppies for the [XO-1](http://laptop.org/en/laptop/hardware/specs.shtml) :crying_cat_face: )

wdlkmpx:
You mean truncate, don't you? I think it's not being used anymore, it was used in /usr/bin/pet2tgz

[...]

PR #670, opened 2015-12-10
Do not include 'truncate' in any WoofCE based Puppy (issue #667)

wdlkmpx:
The truncate utility is deprecated, it was used in /usr/bin/pet2tgz but not anymore
The things is, truncate is a GNU coreutils app, all the more reason to remove this package


mavrothal:
Do we actually *know* that all distros (lucid, magia, whatnot) work fine without the truncate package?

wdlkmpx:
They must...

These are all the 'truncate' string matches in the Woof CE tree
http://pastebin.com/q1bETTzz

I don't see where it's used, there are binary matches in busybox, ntfs-3g, etc

mavrothal:
> They must...

:smirk: 
Maybe so, but I'll leave this for the puppy builders to pull or close.

01micko:
pet2tgz was the only place.

PR #672, opened 2015-12-11
More default apps

wdlkmpx:
This is the first step to make puppy use more default applications in a reasonable way
The predfinied value(s) can be changed anytime to make it them more "generic"


zigbert:
included here
7e27060d8741e65d769246b5e666fc341ba560b9

PR #675, opened 2015-12-14
report-video improvements

wdlkmpx:
http://murga-linux.com/puppy/viewtopic.php?p=876879#876879

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/report-video
+++ b/woof-code/rootfs-skeleton/usr/sbin/report-video
@@ -10,103 +10,201 @@
 export TEXTDOMAIN=report-video
 export OUTPUT_CHARSET=UTF-8
 
-. /etc/DISTRO_SPECS
-
-GRAPHICSCHIP="`lspci | grep -i -E 'graphic|video|display|vga' | cut -f 2- -d ' '`" #130214
-if [ "$GRAPHICSCHIP" = "" ];then
- if [ -f /tmp/ddcprobe.txt ];then
-  GRAPHICSCHIP="`cat /tmp/ddcprobe.txt | grep -E '^oem: |^product: '`"
- else
-  GRAPHICSCHIP="`ddcprobe | grep -E '^oem: |^product: '`"
- fi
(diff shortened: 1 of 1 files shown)


mavrothal:
it is nice but it will break Pup-SysInfo that is using a ```report-video -q 2>/dev/null``` call (I have no idea why '-q') that in your version will produce an invalid argument and no output.  You may want to contact radky (the pup-sysinfo developer) about it.
In general is a good idea when changing a script to check from where and how is called in a running puppy and if it breaks anything

wdlkmpx:
Hahaha report-video -q. It doesn't make sense at all, 

Well, I added the -q option in https://github.com/wdlkmpx/woof-CE/commit/65b956438d1560701bea010fec87edfbce510393 . I don't know, maybe I should allow any value other than the valid options, just in case someone uses other strange parameters, when this script didn't need any:

	#*) echo "Invalid option: $1" >&2 ; exit 1 ;;

mavrothal:
'-q' might be left over from some older version. I do not think should propagate a nonsense if this is indeed the case. Better ask [radky](http://www.murga-linux.com/puppy/viewtopic.php?t=619610) to correct it (not sure if he follows github) if is indeed some leftover.

wdlkmpx:
I think this could be merged as is, I still have pupsysinfo 2.6.0, I just downloaded 2.6.1 and notified radky aboud this, but as of 2015, this -q could be considered canonical, something like "silently print full report"

wdlkmpx:
radky said he's preparing an update, meanwhile I made some changes

mavrothal:
OK. Let's give it few more days till next sysinfo is out.
Plenty of time to work on 0setup :tongue:

wdlkmpx:
radky has released pup-sysinfo 2.6.2, However I will not make further edits to this script to keep backward compatibility... in fact I will, but just to adjust it to arch.

I have confirmed that the output is the same in both original and edited report-video, and I must confess that I have written code that covers most of the same stuff, but I'll probably use report-video instead.

Issue #564, opened 2015-06-27
tahrpup isn't reproducible


dimkr:
It's impossible to build 6.0.2 using:

```
yes "" | ./0setup
./1download
echo "" | ./2createpackages
./3builddistro-Z
```

The process of building tahrpup involves some *mystery* manual tweaking.

I'm currently doing an experiment - trying to build a libre fork of Puppy that is functionally equivalent to tahrpup (Trisquel 7 instead of Ubuntu 14.04, Linux-libre instead of Linux, etc'). It already boots and works, but it's clear that *some parts are missing* (e.g the default wallpaper of tahrpup).

mrfricks:
hi dimkr,

when i built tahrpup i decided not to go with zigberts new ptheme which is now default in woof ...  so i deleted the contents of /woof-out_x86_x86_ubuntu_trusty/rootfs-packages.  this is probably what you're experiencing.

apart from that there is  http://distro.ibiblio.org/puppylinux/pet_packages-tahr/zz_to_add_at_end_of_build.tar.xz which is added just before the iso is created ... this is mainly stuff that woof strips and can't go into the z_tahrfix.pet plus a few extra forgotten things at the time... whatever can be, will be moved to the z_tahrfix next build.   

i just did a test build with the GUI removing the contents of rootfs-packages and it made a perfect tahr.  
wallpaper= vincents-paleblue    gtk & jwm theme = stark-blueish   icons = ardis

dimkr:
Cool stuff! Let me know when the new z_tahr fix is uploaded and the required changes are committed. Looking forward to it.

mavrothal:
Should this still be open?

dimkr:
Until http://distro.ibiblio.org/puppylinux/pet_packages-tahr/zz_to_add_at_end_of_build.tar.xz is gone.

01micko:
Seems like it's gone.

PR #676, opened 2015-12-15
tahrpup 32 & 64bit recipe updates

mrfricks:
issue #564 can now be closed


mavrothal:
I see you put truncate back in. Is there any specific reason for that?

mrfricks:
no an accident... will do another with it removed

PR #678, opened 2015-12-15
sfsget should use sfs_load 

mrfricks:
after sfsget has downloaded an sfs it then opens the bootmanager to load the sfs which requires a reboot.  most modern pups now come with sfs_load so use that 

ignore the double commit :/


mrfricks:
yes that would be better... could you add it, as you know what you're doing ;)

mavrothal:
We'll see about that. :8ball:... 4379c39

PR #679, opened 2015-12-15
fixmenus: speed up

wdlkmpx:
http://murga-linux.com/puppy/viewtopic.php?p=877039#877039


mavrothal:
Indeed :smiley: 
Can you speedup indexgen.sh a bit too?

PR #680, opened 2015-12-16
Implement an 'insmod' facility in 'init' script

gyrog:
This facility provides a workaround when the keyboard does not work during the 'init' script because a kernel module is not available yet.
The module files to 'insmod' can be specified as either a 'pimod' kernel parameter, or the first line of the text file '/mnt/data${PSUBDIR}/${DISTRO_FILE_PREFIX}_${KERNELVER}_init_modules.txt'


gyrog:
mavrothal,

Thanks.

gyro

PR #681, opened 2015-12-16
fixmenus: shorten code

wdlkmpx:
This was
    EXECMENU="`echo -n "$ONELINE" | grep -o '^PUPPYMENU.*' | cut -f 2-5 -d ' '`"

Then it became
    if [[ "$ONELINE" == PUPPYMENU* ]] ; then
        read -r one two three four five <<< "$ONELINE"
        EXECMENU="$two $three $four $five"
    fi

Which is basically the same, but I forgot to add 'six <<<' to make it behave exactly as cut -f 2-5 -d

But I see that this 'cut -f 2-5' is actually *not* needed.

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/fixmenus
+++ b/woof-code/rootfs-skeleton/usr/sbin/fixmenus
@@ -47,10 +47,7 @@ do
  while read ONELINE
  do
    EXECMENU=""
-   if [[ "$ONELINE" == PUPPYMENU* ]] ; then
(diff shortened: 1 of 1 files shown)


wdlkmpx:
What I mean here is that the "^PUPPYMENU " string must be removed, no need to specify a range to be 'cut'. This:

   [[ "$ONELINE" == PUPPYMENU* ]] && EXECMENU="${ONELINE#PUPPYMENU }"

PR #682, opened 2015-12-16
/usr/sbin/indexgen.sh: speed up

wdlkmpx:
http://murga-linux.com/puppy/viewtopic.php?p=877144#877144


mavrothal:
In general script with  loops full of grep, sed. cut etc, processing big files can be benefitted from utilising bash builtins or awk or something. 
0setup processing of the Slackware database (lines 572-619) is one that comes to mind. If you go for it be extra careful to test extensively as 0setup is used both in puppy building and  PPM. If you change more that one thing in it issue different pull requests. Not one big one.

PR #683, opened 2015-12-17
init insmod facility - fix config filename

gyrog:
Specifying the module list by a config file, should behave the same way as specifying the module list by kernel parameter, when handling a different kernel version.
This patch removes the kernel version from the config filename.


gyrog:
mavrothal,

Thanks.

gyro

PR #684, opened 2015-12-17
fixmenus: better logic

wdlkmpx:
It's the same code, but taking this into account:

http://murga-linux.com/puppy/viewtopic.php?p=704831#704831

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/fixmenus
+++ b/woof-code/rootfs-skeleton/usr/sbin/fixmenus
@@ -48,21 +48,22 @@ do
  while read ONELINE
  do
 
-   case $ONELINE in *MENHEIGHT*)
-     MENUHEIGHT=24
-     [ "$MENHEIGHT" ] && MENUHEIGHT=${MENHEIGHT}
-     ## the string 'MENHEIGHT' is replaced by ${MENHEIGHT} or 24 (precaution)
-     ONELINE="${ONELINE//MENHEIGHT/$MENUHEIGHT}" ;;
-   esac
-
    case $ONELINE in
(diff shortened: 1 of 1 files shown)


mavrothal:
Sorry no time for tests. What kind of improvement we are talking about?
Also if you run fixmenus after you shift to a "touch" pTheme, are the menus still OK?
Is it also working in older puppies like Precise?

wdlkmpx:
It also works in precise. Well, I added some comments, it took me a
while to understand this

wdlkmpx:
This is the final version:
https://github.com/wdlkmpx/woof-CE/commit/0f148389ce7c74a2cf756408f066e7b258e3e3cf

wdlkmpx:
This is the "test suite"

http://murga-linux.com/puppy/viewtopic.php?mode=attach&id=92924

just run ./test and check the .diff and .time files

mavrothal:
I see no performance difference what so ever between your previous (pulled) version and this one. At best 2-3%. (In these tests you have to consider that the first time a program runs is always slower 25-50%)
Regarding the "better logic", is there any condition that the old version will fail and the new one will excel?
There are many ways a script can be rewritten, but if there is no tangible benefit (bug, use case, speed. In that order) we are just increasing the chance to introduce issues for negligible,if any, gains.

mavrothal:
I hope we do not dampen your interest.
We are just want to make sure your efforts are going to be as effective as possible :wink:

01micko:
> I hope we do not dampen your interest.
We are just want to make sure your efforts are going to be as effective as possible :wink:

:+1:

PR #686, opened 2015-12-18
partview: center window

wdlkmpx:


mavrothal:
Why?

zigbert:
please use the center attribute carefully. If the user wants to resize the window, it re-centers with a jump. if the gui is set to resizable="false" or for sure not meant to be resized (like a splash screen), the center attribute will work ok.

But still, mavrothal's question is valid - why?
Centering a gui that the user interacted to show, is not the common way to display it (correct me if I'm wrong).

01micko:
Yeah, centering is something for wizards only IMHO.

PR #687, opened 2015-12-18
initmodules - Configure init insmod facility

gyrog:
This script is a GUI to write the config file for the "init insmod facility", (previously committed).
It's based on code "borrowed" from BootManager.


gyrog:
mavrothal,
Thanks.
gyro

PR #689, opened 2015-12-19
probedisk: simplify code

wdlkmpx:
This uses the code posted by jamesbond here:
http://murga-linux.com/puppy/viewtopic.php?t=101946&start=19

And adds puppy code. But I wonder if this /proc/ide stuff really worth being there?
Wary uses a very old kernel and does not have /proc/ide

Also this: usb-drive-log-probedisk

Is this really needed?

### DIFF
--- a/woof-code/rootfs-skeleton/sbin/probedisk
+++ b/woof-code/rootfs-skeleton/sbin/probedisk
@@ -6,27 +6,47 @@
 #v406 support for old kernel, /proc/ide, /dev/hd*.
 # DEVICE|TYPE|INFO
 
-if [ -f $HOME/.usb-drive-log-probedisk ] ; then #force /proc upate mechanism
-  while read ONEUSBDRV ; do
-    dd if=/dev/$ONEUSBDRV of=/dev/null bs=512 count=1 &>/dev/null
-  done < $HOME/.usb-drive-log-probedisk
-fi
+output=1
+case $1 in
+	-alt)  output=2 ; shift ;;
+	-type) output=3 ; shift ;;
+	-desc) output=4 ; shift ;;
+	-size) output=5 ; shift ;;
+esac
 
-# mounted drives/partitions...
-MNTDDEVS="`mount | cut -f 1 -d ' ' | cut -f 3 -d '/' | grep -E '^hd|^sd|^scd|^sr|^mmc' | tr '\n' ' '`"
+device=$1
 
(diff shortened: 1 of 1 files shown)


wdlkmpx:
# 21 Jun 2007 BK: force /proc update for usb drives.
	if [ -f /root/.usb-drive-log-probedisk ];then #force /proc upate mechanism
	  while read ONEUSBDRV ; do
		dd if=/dev/$ONEUSBDRV of=/dev/null bs=512 count=1 &>/dev/null
	  done < /root/.usb-drive-log-probedisk
	fi

wdlkmpx:
Oops I forgot to copy the 1st two comments at least

mavrothal:
Looks good. You may just want to add a comment on top. Something like 
``# major rewrite/cleanup by ...``` :wink:

jamesbond3142:
> the probedisk written by jamesbond is a complete implementation that
discards legacy stuff, so maybe it would be a good idea to look at
fatdog.

Fatdog doesn't use this script. It's there for compatibility purposes; so that .pets converted from puppy will work in Fatdog, and so far it works well for that purpose. That being said, Fatdog's internal scripts use the same method as used in probedisk too: look at /sys/block/* or /sys/class/block/*, and resolve stuff with guess_fstype or blkid when needed for finding disk and partitions.

wdlkmpx:
Now this script has several output options, the -alt option is what I'll use in the future. So I can do this:

read -r dev type size desc <<< "$(probedisk -alt /dev/sda)"

Well, I guess the behavior with zip drives and floppy usbs is still a mystery since I haven't tested myself, I see they are not cheap, and why people should buy that crap nowadays

mavrothal:
No one buys zip and floppy drives now days unless they have some old disk that need to read, because 10 years ago were still fairly popular (their peak was in the 90s!...)
As I said before you may want to make it into a pet and ask people that have zips and floppies to test specifically the said options.

wdlkmpx:
I don't know where to post it, the forum is not responsive to my topics, maybe someone else can post it in the right section, and ask people to paste the output

mavrothal:
I would think in the user section with some catchy title like "probedisk for old hardware (IDE, ZIP drives etc)" :wink:

wdlkmpx:
No one has replied to my topic, I guess this could be probedisk3 or I'll just close this  as I added code the uses /usr/sbin/filesize and basically all the scripts I edit will have my customizations, then real divergence will appear and I'll have to go on on my own

mavrothal:
Is the holiday season. Give it few days.

PR #691, opened 2015-12-20
init insmod facility - support psavemark

gyrog:
These patches enable the "init insmod facilty" config file to work when 'psavemark' is specified.


gyrog:
mavrothal,

Thanks.

I'm always comforted that committing something to woof-CE makes it more
easily modified, not less.

gyro

PR #694, opened 2015-12-22
speed up slackware* processing using mostly shell

wdlkmpx:


wdlkmpx:
I see 13 fields, what I see is that the differences in category are appearing everywhere

Can you upload somewhere your Packages-slackware-*? 

Right now I'm just beginning to analyze what I did hehe

mavrothal:
Actually is the old script that gives 12 fields in salix and patches instead of 13.
I think this change was intentional but I have to look at the logs to see why.

01micko:
>n a rough (test) [#694 (comment)], after the rest of wdlkmpx's changes, about 50-fold faster for the slack dbs . No effect on deb dbs or pet libs.

Yeah.. slackware DB was the only one relying on this script anyway so bash as is to ash is almost irrelevant for others.

wdlkmpx:
I see this code could use the same code in /usr/local/petget/petget

I wrote a function that returns something usable different purposes regarding filenames. Then I added other functions for rpm, archlinux, pet, deb

http://pastebin.com/4iCV5pju

PR #695, opened 2015-12-22
more slackware* processing refinements

wdlkmpx:


wdlkmpx:
Actually it was too rushed, I just checked carefully the diffs. I'll close this, but should I open another one for the sed thing specifically?

mavrothal:
Does it really makes a difference at the end speed wise? Reading a longer file may take a bit more but adding 3 sed calls per package also takes some time.

PR #696, opened 2015-12-23
init insmod facility - improve finding files

gyrog:
'init' now uses the ZDRV variable to decide when to act, and to access the found zdrv...sfs.
'initmodules' and 'bootmanager' now use PDEV1 to find the 'initrd.gz' file.


gyrog:
mavrothal,

Thanks.

gyro

PR #697, opened 2015-12-25
partview: use /usr/sbin/filesize

wdlkmpx:
just execute partview to see what has  changed

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/partview
+++ b/woof-code/rootfs-skeleton/usr/sbin/partview
@@ -63,8 +63,8 @@ do
  
  AFREE=`expr $ASIZE - $AUSED`
  #process size...
- ONEFREE=$(filesize -format $((AFREE*1024)) )
- ONESIZE=$(filesize -format $((ASIZE*1024)) )
+ ONEFREE=$(filesize -bytes=1024 $AFREE)
+ ONESIZE=$(filesize -bytes=1024 $ASIZE)
  
  . /usr/lib/gtkdialog/svg_bar 250 "$(($AUSED*250/$ASIZE))" "${ONESIZE} / ${ONEFREE} $(gettext 'free')"  > /tmp/partview_${COUNT}.svg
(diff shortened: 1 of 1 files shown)


mavrothal:
This pull depends on #693. They should be together.
As mentioned before please provide some reasoning/justification for the changes (Bug, use-case/feature, efficiency)
Also regarding scripts that are called by others  with expected behavior/output try to make sure (as possible) that we do not break anything else and introduce regressions.

wdlkmpx:
filesize is a simple script that use bcs, an arbitraty precision calculator language, which makes it possible to make a floating division.

http://stackoverflow.com/questions/12722095/how-can-i-get-float-division

 filesize can be used to display accurate final results.. that's my idea. 

filesize also comes with a hopefully understandable help message, there's nothing else I can do since Puppy is an 'informal' project 

The next thing I want to do is edit pmount.

mavrothal:
I have no problem with the scrip. Is very nice.
But let me repeat my concerns once more. A lot of the puppy scripts developed in such a convoluted structure because were modified/fixed as issues appeared and features where added. These scripts are tested enough to produce the work they are intended for.
When a script is changed without fixing a bug or adding a feature we fave to verify that all the previous case-specific fixes and additions are not broken in any way. For many of these, we do not even know what they are by now and how to test for them.  As you noticed people are not eager to test whatever comes out either. So we are kind of stuck.
Hopefully we can have soon another merge of testing to master (after Dima and Mick produce their final versions) and then really get wild with testing :wink:

wdlkmpx:
Well, this produces a slightly different output, for example, when I
want to burn a DVD, I just cannot rely or what ROX or the puppy
scripts tell me:

Those utilities might tell me: 4.4G

filesize dir2burn

filesize tells me: 4.25G or something like that

I have to trust filesize since all these linux utilities do not work
the way they should, don't know why

wdlkmpx:
Well, how about we integrate partview into probepart through a special switch, that way partview becomes a pure gui.

device|type|size|freesize

probepart can be altered to use the info of other tools as suggested by gyro to give accurate information about partition size.

These are just ideas of course

wdlkmpx:
Although something more important would be the indent style, somethinkg like 2/3 spaces or a tab

PR #701, opened 2015-12-28
init insmod modules - initmodules - add extra error messages

gyrog:
This adds detection of a CD/DVD boot and provides an explanation message if it can't work with the current setup.
I also adds the standard bootmanager error message if there is no savefile/savefolder defined yet.

### DIFF
--- a/woof-code/huge_extras/initmodules
+++ b/woof-code/huge_extras/initmodules
@@ -2,6 +2,11 @@
 
 . /etc/rc.d/PUPSTATE
 
+if [ "$PUPMODE" = "5" ];then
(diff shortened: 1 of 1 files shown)


gyrog:
mavrothal,

Thanks

gyro

mavrothal:
>Possibly, but the boot command is in 'isolinux.cfg' inside the iso.
That's not so simple to modify

I was referring to the command line boot options that can be typed in. Just a reminder in case the person missed the  boot help menu info.

gyrog:
mavrothal,

Yes, I keep forgetting about that.
I'll take a look at it.

gyro

PR #704, opened 2015-12-29
init insmod facility - change config filename

gyrog:
Removes the "_" from the config filename so it is more like a savefile name.


gyrog:
mavrothal,

thanks

gyro

PR #707, opened 2016-01-01
puppyapps: make_box(): do not allow empty $DEFAULT_EXEC

wdlkmpx:
this fixes the issue i was having in precise


wdlkmpx:
the dialog looks pro, I wonder if you don't mind if I change some stuff to pure shell code, to see whether it loads faster here in this machine..

PR #708, opened 2016-01-02
puppyapps: do not run build_items in the background

wdlkmpx:
This fixes two issues visible in slow machines I suppose

1) The splash window is there after the main dialog appears, for about 3 seconds +/-
Now the splash window disappears just before the main dialog is displayed

--

2) When /tmp/puppyapps* do no exist yet and I run puppyapps, I get random errors like this:

    widget_comboboxtext_input_by_file(): Couldn't open '/tmp/puppyapps_items_spreadsheet' for reading.

All comboboxes are somehow affected and display and blank option first.

After that, if I run puppyapps again there are no errors because all the files are already there

--

In short:

By not backgrounding build_items all files will be created before there is any attempt to use them.

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/puppyapps
+++ b/woof-code/rootfs-skeleton/usr/sbin/puppyapps
@@ -491,7 +491,7 @@ if [ ! "$DEFAULTAPPS" ]; then
 	XPID=$!
 fi
 
-build_items &
+build_items
 
 #build help GUI
 gettext "<b>Helpful Tips</b>


zigbert:
- The startup time is heavily reduced by backgrounding build_items() 
- The splash closes when build_items() is done --> the comboboxlist is refreshed
- the gtk output errors does not affect the use.
- Blank entry in combobox is handled the same way as 'Autodetect'

I can't see any reason for not background this process, but to improve outcome/speed would be great :)

zigbert:
This should fix the blank field issue:
45b28886fe77c6aff09ac5eb95bcd9e2bb0dc2af

wdlkmpx:
Yeah that fixes the issue

zigbert:
In some occasions, I get blank combobox fields on my second test system. Hmmm?
4fc5bfd7bbc2723c9a2bd8e8999a94a5549c9a3d

wdlkmpx:
I was going to say that, but my internet connection went down.

PR #711, opened 2016-01-02
puppyapps: a few edits to improve performance

wdlkmpx:


wdlkmpx:
This file: /tmp/puppyapps/puppyapps_default . I see the only value that is written there is : gettext 'Autodetect'

How about we make a variable instead and place it at the beginnig:
APPS_DEFAULT="$(gettext 'Autodetect')"

wdlkmpx:
This is what I'll do:

APPS_DEFAULT_EXEC="$(gettext 'Autodetect')"

and get_default() will return 5 instead of writing to /tmp/puppyapps/puppyapps_default, then that value (5) will be used later to determine what to do

wdlkmpx:
With these changes, the gui loads 3-6 seconds faster here [randomly], and when the -m switch is used to edit a specifc default app, it loads the gui almost instantly

PR #712, opened 2016-01-04
box_splash: use shell code / assume BOX_SPLASH_BG_GRADIENT=false if n…

wdlkmpx:
…ot set

This takes up 3 seconds to load in a considerably slow machine.

I made several edits to speed up the process, but then I realized that it wasn't enough. This:

    BOX_SPLASH_BG_GRADIENT=true

is what makes it a bit slow. I guess SVG processing is heavy here.

You can use '-bg_gradient true' to get that shiny gradient back.

BOX_SPLASH_BG_GRADIENT can also be set in $HOME/.config/ptheme/gtkdialog_active

    mkdir -p $HOME/.config/ptheme
    echo "BOX_SPLASH_BG_GRADIENT=true" > $HOME/.config/ptheme/gtkdialog_active

    mkdir -p $HOME/.config/ptheme
    echo "BOX_SPLASH_BG=gray
    BOX_SPLASH_BG_GRADIENT=true
    BOX_SPLASH_FG=black" > $HOME/.config/ptheme/gtkdialog_active

So there's nothing to worry about... and puppy will be faster by default

### DIFF
--- a/woof-code/rootfs-skeleton/usr/lib/gtkdialog/box_splash
+++ b/woof-code/rootfs-skeleton/usr/lib/gtkdialog/box_splash
@@ -1,4 +1,4 @@
-#!/bin/ash
+#!/bin/sh
 #splash window - compatible with yaf-splash
 #built upon gtkdialog-splash, written by mave, june 2010.
 #other contributors: BarryK, shinobar, 01micko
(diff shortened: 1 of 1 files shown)


zigbert:
What is the thought of using xrandr instead of xwininfo? (I am blank on this issue)

wdlkmpx:
I've already had written code that uses xrandr... a copy & paste
operation, but both xrandr and xwininfo -root display the same value,
but xrandr gives a very short text, and that is processed by bash
itself way faster than the other method.

To see it yourself, just uncomment the discarded code and add
echo $ROOTX
echo $ROOTY
on both blocks, you will se the same value

I have used the same method to test all the other changes.

I just checked checked the -placement option,
bottom|center|mouse|top-left are working the same.

Regarding the gradient, perhaps when there is no BOX_SPLASH_* set
anywhere else, the script should determine a clearer color for the
background, I see gray90 looks nice, should I do it?

[ ! "$BOX_SPLASH_BG" ]	&&  [ ! "$BOX_SPLASH_BG_GRADIENT" ]	&&  [ !
"$BOX_SPLASH_FG" ]	&& BOX_SPLASH_BG=gray90

That way still looks nice, and loads really fast in a slow machine

01micko:
`xrandr` is an extension to xlib so really shouldn't be used.

@wjaguar 's method is faster than yours, though it isn't shell.

```shell
read ROOTDIMS ROOTHORIZ ROOTVERT << EOF
`xwininfo -root | grep ' \-geometry ' | cut -f 1 -d '+' | tr 'x' ' '`
EOF
```
(see /usr/sbin/background-reshape)

01micko:
BTW, `xrandr` does a lot more in the background than `xwininfo`. You can test the speed of the raw commands without any extra processing to see the difference. 

```
# time xrandr -q 2>&1 >/dev/null

real	0m0.062s
user	0m0.000s
sys	0m0.003s
# time xwininfo -root 2>&1 >/dev/null

real	0m0.012s
user	0m0.017s
sys	0m0.000s
```

wdlkmpx:
OK I'll use wjaguard's method

wdlkmpx:
I added the commits I was talking about. I addressed all the issues I could see. However, if you don't like what you see, I can revert the BG_SPLASH stuff (but after merging the pull request), although it looks nice with a clearlooks feel to it when there is no gradient

zigbert:
There should be no gradient by default. Thank you for your improvements.

wjaguar:
You do not really need grep in there either. Instead of "geometry=$(xwininfo -root | grep ' \-geometry ')", this shell code would do the same thing (as long as xwininfo is present and does what's expected):

geometry=$(xwininfo -root)
geometry=${geometry#*-geometry}

PR #720, opened 2016-01-06
puppyapps: misc edits / gui code relocation / -setdefs option

wdlkmpx:
- misc edits to improve performance
- misc edits to the apps array
- window: execute.svg icon
- window: remove header/dog to make room [i can change it back but check it out first]
- -setdefs, option [this can be useful for testing purposes]
- $IGNORE_ALREADY_SET set by -setdefs and processed by get_default()
- move gui specific code down the command line code block
- possibly other changes not mentioned here

Sorry for cluttering all these changes.. i caught the flu

Now the gui looks ready, but I don't know to make it show up maximized so the user has a clear vision of all apps. Any ideas?
http://s11.postimg.org/k10a1urnl/capture25912.png

Maybe 2 or 3 apps are missing, two of them would be:
defaulttorrent
defaultpresentation [ms powerpoint-like apps]

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/puppyapps
+++ b/woof-code/rootfs-skeleton/usr/sbin/puppyapps
@@ -415,18 +415,23 @@ for arg in "$@" ; do
 		-setdefs)
 			desktop_output
 			IGNORE_ALREADY_SET=1
-			for app in $(default_apps) ; do
-				file=/usr/local/bin/default${app}
-				apps_array $app > /tmp/puppyapps/puppyapps_items_$app
-				RUNAPP_EXEC=$(get_default "$app")
-				if [ $? -eq 1 -o ! "$RUNAPP_EXEC" ] ; then
-					echo -e "#!/bin/sh\nexec puppyapps default${app} \"\$@\"" > $file
-					echo "[autodetect] > $file"
-				else
-					echo -e "#!/bin/sh\nexec $RUNAPP_EXEC \"\$@\"" > $file
-					echo "$RUNAPP_EXEC > $file"
-				fi
-				chmod +x $file
+			shift
+			appz=$(default_apps)
+			[ "$1" ] && appz="$@"
+			for app in ${appz} ; do
+				case "$(default_apps)" in *"${app}"*)
+					file=/usr/local/bin/default${app}
+					apps_array $app > /tmp/puppyapps/puppyapps_items_$app
+					RUNAPP_EXEC=$(get_default "$app")
+					if [ $? -eq 1 -o ! "$RUNAPP_EXEC" ] ; then
+						echo -e "#!/bin/sh\nexec puppyapps default${app} \"\$@\"" > $file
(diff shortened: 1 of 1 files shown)


zigbert:
You are now increasing the size of the window. There has been a ancient understanding that puppy should run ok on a 800x600 screen. Using 3 columns in the gui, it does not. Please revert, and we try to find another solution that won't break the Puppy 'mantra'.

gtkdialog allows only static widget, so it can't ie. switch from 2 to 3 columns if required when maximizing/scaling gui.

zigbert:
```
Xdialog --calendar "" 350x270
```

wdlkmpx:
It still looks good in a 800x600 screen:
http://s12.postimg.org/eyupkxn63/capture19060.png

I'll check in my netbook toshiba nb100

wdlkmpx:
800x600 in a small netbook, slacko 6.3.0
http://s8.postimg.org/hy5hb690l/Screenshot_1.png

but I see that this may not the best approach for mobile devices, just confirm again that you want this reverted to 2 columns..

wdlkmpx:
I see, I added that one, i also I removed ical calcurse, they're using tlc/tk, maybe other more generic apps like osmo can be added, gaby perhaps? (i removed that one before) Even Puppy 4 used osmo I think, but I can't confirm it.

Who can I talk to upload more updated yad pets to the official repos? And even add them to the official puppies

mavrothal:
Any chance that the list could be presented in some sort of order? 
Alphabetical, desktop category, other?
Looking at 20-30 options in one window can be confusing with 3 columns, and partially hidden with 1 or 2 columns.

wdlkmpx:
Yeah, a lot of categories, but if you do:

	puppyapps -m defaultaudioeditor defaultmediaplayer

then the gui will display only those 2 apps
although after commit https://github.com/wdlkmpx/woof-CE/commit/cf83542bbb0a6bf508a90ebddf463a93df686755 you can also do:

	puppyapps -m audioeditor mediaplayer

-

wdlkmpx:
Yeah, I'll see what I can to implement some sort of roder

zigbert:
Regarding the app-array:
One of the ideas is to prefer a user-installed app.
Ie. I see you switched order for Peasydisc and pBurn. This makes pBurn the 'Autodetected' app, even if user installed pEasyDisc. This is not as intended. If user installs pEasyDisc, it should be the preferred one.
My thought is that the most common apps in a standard Puppy should have the lowest priority. Well, the fallbacks should of course be placed at the end of the list.

wdlkmpx:
Your pburn app is well written although a bit heavy, my brother uses it in arch linux, I'd like to see what I can do to make it faster, but OK in future commits peasydisc will be in its previous position, although I also wrote a complete small solution for burning, based on burniso2cd

http://pastebin.com/7TffwnHq

zigbert:
Your help is welcome, but my comment is just a reminder of keeping focus when changing sort order of the app-array.

PR #725, opened 2016-01-09
make updatenetmoduleslist.sh a bit faster

wdlkmpx:
(original)
real    0m42.895s
user    0m8.993s
sys 0m20.629s

(mod)
real    0m7.683s
user    0m1.600s
sys 0m3.386s


mavrothal:
Tested OK

PR #727, opened 2016-01-14
initmodules - look in zdrv partition first

gyrog:
Always look in the zdrv partition for the config file first.
When 'savemark' is active, performance can be improved by copying the config file to be beside the zdrv...sfs, instead of beside the savefile/savefolder. This avoids an extra mount of the savefile/savefolder partition.


gyrog:
mavrothal,

Thanks.

gyro

PR #728, opened 2016-01-14
rc.update/sysinit: use busybox depmod as a last resort

wdlkmpx:
If kmod is there, then busybox depmod  is not used at all
However busybox depmod has a higher priority than depmod itself.
If I remember correctly, depmod was a symlink to busybox in older versions,
 so it makes sense to use depmod as is.

Besides, it seems that busybox needs to be patched for the depmod applet to work with puppy?

I wrote something about this a long time ago:
http://murga-linux.com/puppy/viewtopic.php?t=99162


01micko:
Yeah I had the same problem with slacko back in the 1337 days when kmod was first introduced in slackware.

PR #729, opened 2016-01-16
more options for puppyapps

wdlkmpx:
-dc: replace defaults chooser [if found]
-fddb: download and store latest freedesktop.org.xml
-mimedb: update mime database using the puppyapps xml files
-mimedefs: create mime associations for different standards
-system: -dc + -mimedb + -mimedefs

to test and debug -mimedefs:
    export TEST=1
    puppyapps -mimedefs
    (now go to /tmp/mimedefs and check)
you can do this afterwards:
    puppyapps -mimedefsa &> mimedefs.log

----

These functions use files from /var/lib/puppyapps

Sample files:

mimedb/*.xml - mime packages
dotdesktop - custom .desktop files
exec - app:exec [how to run properly certain apps]
mime_associations - mime associations [mimetype=app;app2;app3]
mime_defaults - mime default associations

Entries/data can be added/altered/removed

Puppyapps can be used to experiment, test new versions of shared-mime-info and can be very useful
to the primitive puppy "desktop environment" making it a bit configurable and expandable
without the need to create more files, all are stored in "dabatases" in /var/lib/puppyapps

I use it with my own "databases" to generate all "my environment" when I do remasters.


wdlkmpx:
Nobody has a word about this pull request, it has been a lot of work. I don't like silence, so I'm closing this.

mavrothal:
> Nobody has a word about this pull request, it has been a lot of work. I don't like silence, so I'm closing this.

As you may have noticed zigbert (the person most relevant to this) is not much around these days.
However, I do not see why the code should be removed because of the delay if you think is valid.
I would suggest to reopen it.

One thing that makes me a bit skeptical about pulls related to puppyapps is that we discussed at some point that puppyapps and the associated files would be moving to rootfs-skeleton and would not be mandatory for puppy to function. Is this still in the cards?

wdlkmpx:
What I think is that things are moving too slowly, I see all this
effort worthless. The way to go is adopting a DE and just get rid of
years of work, LXDE seems to be the right choice.

Whatever choice is right as long as ROX is wiped once and for all for
the almighty's sake

mavrothal:
As I mentioned earlier, build a puppy any way you want it and get it out there :wink:

PR #733, opened 2016-01-27
alsaconf: minor speedup

wdlkmpx:
that is more noticeable in slow pcs

build_card_db

(original)
real    0m30.079s
user    0m5.343s
sys 0m15.482s

(mod)
real    0m5.384s
user    0m0.747s
sys 0m1.197s

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/alsaconf
+++ b/woof-code/rootfs-skeleton/usr/sbin/alsaconf
@@ -485,7 +485,7 @@ build_card_db () {
 
     # list pci cards
     while read driver vendor device dummy; do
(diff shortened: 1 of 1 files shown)


01micko:
`alsaconf` doesn't work in any distro that ships `kmod`. I suppose there should be a condition in `3builddistro(-Z)` that detects `kmod` and acts accordingly. 

NB: that was my one contribution to slackware-14.0 development : http://www.linuxquestions.org/questions/slackware-14/slackware-14-0-rc4-4175425014/page4.html#post4771725

That said, I might merge it anyway.

01micko:
Ok, after consideration it can't hurt to merge this.

PR #735, opened 2016-01-31
Inactivate ddcprobe if broken. Freezes X startup

mavrothal:
ddcprobe just hangs in some distros and hardware (Tahr64 in a MacBookPro in my case) and then the X startup (xwin) as well as xorgwizard{,-automatic,-cli} all freeze. Upon inactivation X proceeds straight to Desktop!
Not sure if such an irreversible measure is wise, or should go through the hassle to test in every script that uses ddcprobe and revert at the script exit. 
Given that we move away from xorgwizard is probably an overkill.
Please review. (Also the code could be improved.)

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/xorgwizard-automatic
+++ b/woof-code/rootfs-skeleton/usr/sbin/xorgwizard-automatic
@@ -14,6 +14,28 @@
 
 . /etc/rc.d/MODULESCONFIG
 
+# ddcprobe hangs in some hardware and the entire X startup freezes.
+# test and inactivate early on
+if [ "$(which ddcprobe)" ]; then
+ for TRY in 1 2 3
+ do
+  ddcprobe &
+  XPID=$!
+  sleep 0.5
+  RUNNING="$(cat /proc/"$XPID"/status | grep -i running)"
+  [ "$RUNNING" = "" ] && break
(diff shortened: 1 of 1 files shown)


01micko:
Does ddcprobe hang in slacko64 on the macbook?

mavrothal:
Yes it does. But 6.3.0.1 goes straight to the desktop :wink:

01micko:
Ok, then the same change should be added to xorgwizard. Macs may not be the only ones suffering this bug. I suspect PowerVR chips might too.

mavrothal:
A new puppy defaults in using xorgwizard-automatic first. No?
So this should deactivate ddcprobe for all cases.
Of course someone could boot with nox and then use xorgxizard or xorgwizard-cli (or one of the couple other scripts that call it). if we want to cover those too is better to make it an independent script and call it from the other places.
But, aren't we moving away from xorgwizard and friends?

01micko:
Well I would like to ditch xorgwizard-automatic (as in `xorg` branch) and since we have been mostly booting to desktop anyway for a few years ddcprobe is pretty much useless. All it is for is to get some modelines (which can be done on cli anyway) and append the monitor name to the xorg.conf in case we switch usb stick installs to other computers.

I suppose we can merge this anyway.

PR #737, opened 2016-02-02
alsaconf: fail if distro is using kmod

wdlkmpx:


01micko:
Not very elegant!

I reckon it should just get deleted in 3builddistro(-Z). When I get some more time I'll take a look.

wdlkmpx:
Well, kmod may be there, but if depmod is not symlink to kmod, then it's not being used.

Yeah I think it's bloated, according to LFS, symlinks are enough
http://www.linuxfromscratch.org/lfs/view/development/chapter06/kmod.html

I was thinking of e2fsprogs I guess, where hard links are applied:
http://www.linuxfromscratch.org/lfs/view/development/chapter06/e2fsprogs.html

It was another rushed decision to make a pull request, but I guess just bored..

01micko:
A distro builder has to take some responsibility. If they include module-init-tools and kmod then that isn't a very sane choice. The existence of kmod in the distro build should be enough just to delete alsaconf. That's my thoughts on the matter anyway.

Bored or not, keep the ideas coming. :smiley:

PR #738, opened 2016-02-03
10alsa: various edits

wdlkmpx:
These changes allow the user to check 10alsa's output in
/tmp/services/10alsa.start.log and /tmp/services/10alsa.stop.log

It also checks for /dev/mixer a little longer if necessary.

### DIFF
--- a/woof-code/rootfs-skeleton/etc/init.d/10alsa
+++ b/woof-code/rootfs-skeleton/etc/init.d/10alsa
@@ -16,20 +16,28 @@ case "$1" in
 	mkdir -p /tmp/services
 	(
 	rm -f /var/lock/subsys/alsasound 2> /dev/null #or alsa will not start.
-	#sometimes these don't all load...
-	modprobe snd-mixer-oss
-	modprobe snd-seq-oss
-	modprobe snd-pcm-oss
+
+	#sometimes these don't all load... oss-compat
+	modprobe snd-mixer-oss #/dev/mixer http://alsa.opensrc.org/Snd-mixer-oss
+	modprobe snd-seq-oss   #oss midi   http://alsa.opensrc.org/Snd-seq-oss
+	modprobe snd-pcm-oss   #/dev/dsp /dev/audio http://alsa.opensrc.org/Snd-pcm-oss
 
 	wt=0.5
 	for I in $(seq 10); do #110111 shinobar: for reliable sound detection on HP Compac nc4010.
-		[ -c /dev/mixer ] && break #note, /dev/mixer is deleted in /etc/rc.d/rc.sysinit.
+		[ -c /dev/snd/controlC0 ] && break #alsa sound interface
+		[ -c /dev/snd/pcmC0D0p ] && break #alsa sound interface
+		#[ -c /dev/mixer ] && break #oss-compat
+		#note, /dev/mixer is deleted in /etc/rc.d/rc.sysinit.
 		echo "Waiting $wt seconds"
 		sleep $wt
 	done
-	if [ ! -c /dev/mixer ] ; then
-		echo "ERROR: /dev/mixer is missing"
-		exit 1
+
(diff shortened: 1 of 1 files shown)


wdlkmpx:
I've been testing these changes in several puppy versions, from puppy431 to slacko 6.3.0/tahrpup... the script works very well in all of them, this machine in particular had "timing" issues with newer kernels that have never been addressed. If net-setup were part of woof-ce, perhaps the circle could be completed..

mavrothal:
Looks OK to me though when I tried it in a problematic hardware did not improve things (but is likely a kernel issue). The log kind of confirmed that (showed no issue). 
Sound is not my thing though so I'll leave it to Mick.
Regarding net_setup, why don't you provide a new pet? The latest one is from 2012. 
It will need extensive testing anyway as it touches some core puppy functions.

wdlkmpx:
Actually there are 2 only signficant changes:
1) output redirection / verbose info
2) wait up to 5 seconds for /dev/mixer to become available [yes, it can take some time in the worst cases]

I've tested in 2 laptos, 1 netbook, 3 pcs.. working as usual, I've been looking at the code in rc.services, it has even more code related to alsa.

However I wonder why the processes are not backgrounded?

	[ -x $service_script ] && $service_script start

--

wdlkmpx:
I also think this code in /usr/sbin/delayedrun is obsolete

	#101121 handle multiple sound cards...
	if [ $PUPMODE -eq 5 ];then
	 if [ -f /tmp/rc_d_rc_services_alsa_ok ];then #see /etc/rc.d/rc.services
	  if [ `find /tmp/pup_event_backend -maxdepth 1 -type f -name 'pup_event_alsa_cnt*' | wc -l` -gt 1 ];then
	   #run kirk's Multiple Sound Card Wizard...
	   Multiple-Sound-Card-Wizard
	  fi
	 fi
	fi

It can replaced with something like this:


	#101121 handle multiple sound cards...
	if [ $PUPMODE -eq 5 ];then
	  if [ `ls -d /proc/asound/card[0-9] | wc -l` -eq 1 ];then
	   #run kirk's Multiple Sound Card Wizard...
	   Multiple-Sound-Card-Wizard
	  fi
	fi

wdlkmpx:
I noticed the codes posted above contained -eq 1 (I fixed that, changed it to -gt 1), which I was using to test if it triggers  Multiple-Sound-Card-Wizard with 1 soundcard

As you can see there are changes to be done

wdlkmpx:
I added a new commit. mavrothal, I'd like to see the log from your problematic hardware

I inserted a pci sound card to test:

This what happens at startup: http://pastebin.com/7YJCqYbT

The onboard soundcard is missing, it gets loaded later

	/etc/init.d/10alsa  start

When I run 10alsa manually: http://pastebin.com/GGiMHuRY

I'll insert another pci sound card

mavrothal:
The updated script indeed provides more info in the logs now, which is nice.
However, as you can probably see from the attached logs and asound.state,  [alsa_logs.zip](https://github.com/puppylinux-woof-CE/woof-CE/files/120617/alsa_logs.zip) alsa can not handle any of the sound cards appropriately in this hardware. I guess is a kernel issue (both Tahr64 and Fatdog are fine with the "analog" card but non can handle HDMI) and I do not think is really related to the pull at hand. 
Looks like Mick is really busy so if he does not appear, I'll probably merge it by tomorrow unless more commits appear :wink:

wdlkmpx:
Uhmm I think that could be solved by adding a rule somewhere to ignore the HDMI output or something like that.

I added a new commit, basically it gets the card driver. I think 3 days could be enough.  I guess that's it, something else... perhaps introduce differentes between start and restart ... later.

kernel 2.6.30: http://pastebin.com/9sNAa3mJ
kernel 3.14.55: http://pastebin.com/D5w8ApxS

PR #739, opened 2016-02-04
initmodules - trim config file line

gyrog:
Trim any residual end-of-line characters from config file line.
The config file can be a dos file or inside an iso, and still work.


mavrothal:
Have you seen #732 ?

gyrog:
mavrothal,

Thanks.

gyro

gyrog:
mavrothal,

I had not.

"edit-initramfs" is a utility written by BK in 2012 to assist developers
in modifying files in "initrd.gz".
This looks like a re-write of this utility.

I don't use this utility, I have my own scripts for extracting and
storing files in "initrd.gz".

I don't see this a interfering with the initmodules code.

gyro

PR #740, opened 2016-02-06
enable psavemark to work with pmedia and psave

gyrog:
Currently, if a SAVEMARK file or psavemark boot parameter is being utilised, and pmedia and pdev1 boot parameters are specified, the psavemark partition is ignored.
This patch ensures that even if pmedia and pdev1 are specified, the psavemark partition remains in the search list.
A save partition specified by a SAVEMARK file only, will still be ignored, but the addition of a psavemark boot parameter will restore the save partition to the search list.


gyrog:
mavrothal,

Thanks.

gyro

PR #741, opened 2016-02-08
remove all references to /tmp/rc_d_rc_services_alsa_ok

wdlkmpx:
/tmp/rc_d_rc_services_alsa_ok is created by rc.services and read
by 10alsa (removed long time ago) and delayedrun

There is a more standard way to handle the situation in delayedrun.

http://murga-linux.com/puppy/viewtopic.php?p=885747#885747

### DIFF
--- a/woof-code/rootfs-skeleton/etc/rc.d/rc.services
+++ b/woof-code/rootfs-skeleton/etc/rc.d/rc.services
@@ -19,7 +19,6 @@ while [ $BRKCNT -lt 7 ];do
   GOODFLAG=1
  fi
 done
-[ $GOODFLAG -eq 1 ] && touch /tmp/rc_d_rc_services_alsa_ok #read by /usr/sbin/delayedrun, /etc/init.d/10alsa.
 
 for service_script in /etc/init.d/*
 do
--- a/woof-code/rootfs-skeleton/usr/sbin/delayedrun
+++ b/woof-code/rootfs-skeleton/usr/sbin/delayedrun
@@ -129,12 +129,8 @@ fi
 
 #101121 handle multiple sound cards...
 if [ $PUPMODE -eq 5 ];then
- if [ -f /tmp/rc_d_rc_services_alsa_ok ];then #see /etc/rc.d/rc.services
-  if [ `find /tmp/pup_event_backend -maxdepth 1 -type f -name 'pup_event_alsa_cnt*' | wc -l` -gt 1 ];then
-   #run kirk's Multiple Sound Card Wizard...
-   Multiple-Sound-Card-Wizard
-  fi
- fi
+  scards=`grep '\[.*\]' /proc/asound/cards | grep -v 'pcsp' | wc -l`
+  [ $scards -qt 1 ] && Multiple-Sound-Card-Wizard #by kirk
 fi
 
 # give xchat unique username based on distro
--- a/woof-code/rootfs-skeleton/usr/sbin/delayedrun
+++ b/woof-code/rootfs-skeleton/usr/sbin/delayedrun
@@ -130,7 +130,7 @@ fi
 #101121 handle multiple sound cards...
 if [ $PUPMODE -eq 5 ];then
   scards=`grep '\[.*\]' /proc/asound/cards | grep -v 'pcsp' | wc -l`
-  [ $scards -qt 1 ] && Multiple-Sound-Card-Wizard #by kirk
+  [ $scards -gt 1 ] && Multiple-Sound-Card-Wizard #by kirk
 fi
 
 # give xchat unique username based on distro


wdlkmpx:
in the forum thread gcmartin's report shows a webcam sound card.. should it also be ignored?

mavrothal:
> in the forum thread gcmartin's report shows a webcam sound card.. should it also be ignored?

Probably not. It may provide the only audio input (mic) in the system

wdlkmpx:
I think it's ready to be merged... I tested in pupmode 5 and other pupmodes to make sure it's correct.

Then I realized I made a silly mistake.. a typo, which is fixed..

I'll be analyzing rc.services

wdlkmpx:
I have more plans for delayedrun. I've edited MSCW to work in conjuction with delayed run. MSCW seems to work fine with its new look, check it out:
http://murga-linux.com/puppy/viewtopic.php?p=888643#888643

This is code I plan to add to delayedrun
http://pastebin.com/aKswXP2w

I wonder if it's possible to just stuff my version of MSCW into woof ce

mavrothal:
> I wonder if it's possible to just stuff my version of MSCW into woof ce

Alsa, even if pulse audio is gaining ground I think will remain in puppy for some time so inclusion could be an option. 
MSCW however is a pet, so why don't we just use yours to update the one in the repo (Tahr is using its own but I think is just Geoffrey's). Send it to Mick or Phil to upload it into the noarch repo.
Alternatively can be placed in rootfs-packages (so will be easier to maintain) and make the relevant changes in DISTRO_PKGS_SPECS-* not to install the pet.
They are both fine with me. What the builders think?

wdlkmpx:
Silence means consent should be the rule here

mavrothal:
Things are moving slowly these days...
However you can still issue a pull request or just make a new mscw pet.
Or you may wand to woof some puppies yourself :wink:

mavrothal:
What is the story here?
Where are we standing with mscw/delayrun and this pull?

wdlkmpx:
I posted a probably final pet here:
http://murga-linux.com/puppy/viewtopic.php?p=889663#889663

Well this can be merged to keep a record. That file is pretty useless...

The next pull request will bring MSCW and delayedrun. It's stable enough, but we can wait a week to test more carefully. I've been testing and added more checks and it works as expected.. from the first boot to the last

Regarding MSCW, I'm thinking of this:
1) change name to /usr/bin/mscw /usr/bin/mscw.alsamixer
2) mscw will automatically replace Multiple-Sound-Card-Mixer (symlink) and alsamixer, that way there is no way it can be removed by installing a MSCW pet

mavrothal:
> Regarding MSCW, I'm thinking of this:
1) change name to /usr/bin/mscw /usr/bin/mscw.alsamixer
2) mscw will automatically replace Multiple-Sound-Card-Mixer (symlink) and alsamixer, that way there is no way it can be removed by installing a MSCW pet

I think we should allow the builder to chose.
I think is better to get it into rootfs-packages and remove the mscw from the different DISTRO_PKGS_SPECS-* . Change also 3builddistro in the relevant segment where it [picks rootfs-packages] (https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-code/3builddistro-Z#L304).

BTW why do you want to remove alsamixer?

wdlkmpx:
I don't want remove alsamixer, mscw.alsamixer would be the equivalent of alsamixer.script

wdlkmpx:
Actually that script is not needed at all, I'll remove it and i'll add to MSCW a setting for alsamixer to pick the correct card

wdlkmpx:
There's a new pet file at the forum, actually I only added one line to write_asound_config. The original MSCW uses alsamixer.script to pick the right soundcard, that's why it has a pinstall.sh file... I removed both files

wdlkmpx:
I'm not sure how to proceed, if another MSCW is to be installed then delayedrun must have code to handle that situation this is quiet mode for my MSCW (the full automatic behavior is disabled by default)

	function start_quiet() { ## PUPMODE=5
		. /etc/rc.d/PUPSTATE
		if [ $PUPMODE -eq 5 ] ; then
			if [ ! -f /tmp/mscw.pm5 ] ; then
				scards=$(ncards)
				#sdevs=$(ndevs)
				HDMI=$(aplay -l | grep 'card 0' | head -1 | grep 'HDMI')
				if [ $scards -gt 1 ] || [ "$HDMI" ] ; then
					touch /tmp/mscw.pm5
					$0 & #launch MSCW GUI
					exit
				fi
			fi
		fi
	}


then delayedrun would look like this:

	if [ $PUPMODE -eq 5 ] ; then
		if [ ! -f /tmp/mscw.pm5 ] ; then
			scards=$(grep '^ [0-9]' /proc/asound/cards | grep -v 'pcsp' | wc -l)
			HDMI=$(aplay -l | grep 'card 0' | head -1 | grep 'HDMI')
			if [ $scards -gt 1 ] || [ "$HDMI" ] ; then
				touch /tmp/mscw.pm5
				Multiple-Sound-Card-Wizard start #parameter only meaningful to WCE MSCW
			fi
		fi
	fi

mavrothal:
I think that is ok to have some code in delayedrun to check for the presence of the specific mscw and if present call it. ```Multiple-Sound-Card-Wizard start``` will open any other mscw so you may need something more specific  (unless you are referring to ```[ $scards -gt 1 ] || [ "$HDMI" ]```

wdlkmpx:
I think I lost control of this branch (testz3), I'll commit changes to delayedrun and we'll discuss this in the other pull request..

PR #743, opened 2016-02-15
box_splash: bugfix

wdlkmpx:
busybox ps: the less digits a pid has, the more leading spaces will be

        1 root       0:01 /bin/busybox init
      635 root       0:00 /tmp/yaf-splash --class=gtkdialog-splash ...
     1172 root       0:00 grep yaf
    30053 root       0:00 /bin/bash

solution:  convert multiple spaces to one space


wdlkmpx:
Yes, that was an actual ps, from time to time I find a window that would not close

mavrothal:
Then I would probably go for this and foget about the spaces 
```[ "$(busybox ps | awk '{print $1}' | grep "${dlgPID}")" != "${dlgPID}"  ] && exit #already killed ```

wdlkmpx:
But what my other commit, give it a look, it's accurate even to get pid 2 or 22 or 222 or 2222

PR #744, opened 2016-02-17
Multiple Sound Card Wizard for rootfs-packages (closes #736)

### LINKED ISSUE #736: Integrate Multiple Sound Card Wizard?
It's about 3kb uncompressed and I find it sad that such a tiny script has to be in repo...

The package name is cryptic: mscw, mscw2 [a sed command can take care of this in woof-distro/]

musher0 shared this link.. I was missing this update:
http://murga-linux.com/puppy/viewtopic.php?p=853357

wdlkmpx:
Originally written by Kirk
Modified and fixed by rg66 and Geoffrey
completely rewritten for Woof CE

Includes code to handle possible issues with soundcards.

### DIFF
--- a/woof-code/rootfs-packages/mscw/pet.specs
+++ b/woof-code/rootfs-packages/mscw/pet.specs
@@ -1,0 +1,1 @@
+mscw-20160217-noarch|mscw|20160217-noarch||Setup|17K||mscw-20160217-noarch.pet||Multiple Sound Card Wizard||||
--- a/woof-code/rootfs-packages/mscw/usr/bin/Multiple-Sound-Card-Wizard
+++ b/woof-code/rootfs-packages/mscw/usr/bin/Multiple-Sound-Card-Wizard
@@ -18,6 +18,9 @@ esac
 #                          FUNCTIONS                          #
 ###############################################################
 
+function ncards() { grep '^ [0-9]' /proc/asound/cards | grep -v 'pcsp' | wc -l ; }
+function ndevs() { grep -v 'pcspeaker' /proc/asound/pcm | wc -l ; }
+
 function is_valid_device() {
 	case $1 in 0|1|2|3|4|5|6|7|8|9) return 0 ;;
 		*) return 1 ;;
@@ -40,28 +43,22 @@ function write_asound_config() { #card device
 	echo "defaults.pcm.card $1"   #line 1
 	echo "defaults.pcm.device $2" #line 2
 	echo "defaults.ctl.card $1"   #line 3
-	) > /etc/asound.conf
-	cp -f /etc/asound.conf $HOME/.asoundrc #make sure both files have the same config
+	) > $HOME/.asoundrc
+	cp -f $HOME/.asoundrc /etc/asound.conf  #make sure both files have the same config
 	write_retrovol_card $1
 }
 
 function asound_conf_values() { #returns: card device cardmixer
-	if [ -f /etc/asound.conf ] ; then
-		xfile="/etc/asound.conf"
-	elif [ -f $HOME/.asoundrc ] ; then
+	if [ -f $HOME/.asoundrc ] ; then
 		xfile="$HOME/.asoundrc"
+	elif [ -f /etc/asound.conf ] ; then
+		xfile="/etc/asound.conf"
 	else
 		return 1
 	fi
-	x=1 ; while read line ; do
-		value=${line#* }
-		[ $x -gt 3 ] && break
-		echo -n "$value "
-		let x++
-	done < $xfile
-	#{
-	#	for x in 1 2 3 ; do read a b ; echo -n "$b " ; done
-	#}< $xfile
+	{
+		for x in 1 2 3 ; do read a b ; echo -n "$b " ; done
+	} < $xfile
 }
 
 function check_asound_conf() {
@@ -80,15 +77,20 @@ function check_asound_conf() {
 }
 
 function print_asound_conf() {
-	if [ -f /etc/asound.conf ] ; then
+	if [ -f $HOME/.asoundrc ] ; then
+		local xfile="$HOME/.asoundrc"
+	elif [ -f /etc/asound.conf ] ; then
+		local xfile="/etc/asound.conf"
+	fi
+	if [ "$xfile" ] ; then
 		echo
-		echo "------------ asound.conf ------------"
-		cat /etc/asound.conf
+		echo "------------- ${xfile##*/} ------------"
+		cat $xfile
 		echo "-------------------------------------"
-		echo
 	fi
 	if [ -f $HOME/.retrovolrc ] ; then
-		echo "---------- HOME/.retrovolrc ---------"
+		echo
+		echo "------------ .retrovolrc ------------"
 		grep 'card=' $HOME/.retrovolrc
 		echo "-------------------------------------"
 		echo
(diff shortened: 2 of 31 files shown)


wdlkmpx:
Ok, I'm not sure what to do next, so awaiting comments.

Here is a pet file:
http://murga-linux.com/puppy/viewtopic.php?p=889679#889679

The automatic stuff is disabled by default... unless you want me to change it back..

mavrothal:
You want also to make changes in 3-bulddistro to include mscw [like is done here] (https://github.com/wdlkmpx/woof-CE/blob/testing/woof-code/3builddistro-Z#L304) for other packages in rootfs-packages and set pet installation to no in the [petspec files]  (https://github.com/wdlkmpx/woof-CE/blob/testing/woof-distro/x86/ubuntu/trusty/DISTRO_PKGS_SPECS-ubuntu-trusty#L636) (look for the other DISTRO_PKGS_SPECS-* files with mscw)

mavrothal:
> The automatic stuff is disabled by default... unless you want me to change it back..

No time to look at it now. Why is it disabled? I think that running on pupmode 5 called by delayedrun is good but does not need to show anything 
a) if it can see alsa configured properly or 
b) configure it itself (ie do a dry run and if whatever is already set agrees with the mscw quit). 
c) If not and mscw can pick another card write this configuration and quit without showing anything. 
If it can not pick a card automatically then should open the GUI.

wdlkmpx:
Actually the automatic stuff is working fine, I'm making some changes to limit its scope according to some rules...

I'll make sure to follow a) b) c)

Meanwhile check this commit that explains even more https://github.com/wdlkmpx/woof-CE/commit/880f6224c8f6fcb31b1e658c06305f987a9b7bf5

mavrothal:
> Meanwhile check this commit that explains even more wdlkmpx@880f622

Looks OK.
This is not part of this pull though. Correct?
>I'll make sure to follow a) b) c)

That would be very nice. 
Also please see about the desktop file.

wdlkmpx:
How about we make it part of this pull? It has do with do MSCW... let's consider this MSCW a branch where everything related to this MSCW will be added in the next commits... if it's not deemed worthy for inclusion, then everything will stay the same as it was before...

Or if you really want to, I can make another pull request for this specific change for delayedrun..

I'm also recieving feedback, see here:
http://murga-linux.com/puppy/viewtopic.php?p=890136#890136

wdlkmpx:
I added NoDisplay=true (desktop), no|mscw (distro specs)... 3builddistro-Z.. it looks like no|mscw is enough

I'm testing a) b) c) d) [i already done most of the work], the next update for MSCW will be huge...

I've been told not to touch .asoundrc, so by using /etc/asound.conf we're playing safe...

wdlkmpx:
For a first release.. I think it's done. This MSCW behaves like the old one, but adds some extra checks for PUPMODE 5... it also does not show the MSCW dialog, it's up to the user (may be changed).

Here is a test in a very old motherboard using the slowest processor I have... somewhere between P III - P IV

I have a pci wireless ethernet card in slot 1 and a pci soundcard in slot 2

1st boot
------------
no internet connection [wired ethernet]... to be solved by net-setup soon

10alsa.start.log... failed http://pastebin.com/Fb4wCQnL
mscw.start.log... ok http://pastebin.com/WtHxz7jz

we can see that mscw can tell new soundcards (2!) were loaded after 10alsa's first run, so it executes 10alsa restart:

10alsa.restart.log http://pastebin.com/1BMGtvnc

mavrothal:
OK.
What about these changes in asla-base.conf? Looked good to me.
Should they be part of this commit. Another? None?

wdlkmpx:
I opened a pull request for that, I think it just prevents HDA HDMI from being the first soundcard/pcm device when thera are other alternatives... then I think I'll comment out the code that calls CheckHDMI in MSCW.. I'm not sure.

I think I'll build a slacko, but I feel lazy.. I'll gather all these files and ask mrfricks to build a tahrpup to see the results... I just want eveything unified ready to use..

mavrothal:
Peebee is also building puppies often enough from the Slackware-current branch. Usually both the LXDE and the JWM versions :wink:

01micko:
@wdlkmpx - you should check out my retrovol fork - speaker branch. (So called because it allows the icon colour to be adjusted, without puppyisms; eg: `root` - writes the icons to `$HOME`. Look for the _colorbutton_ in  the config window > tray tab.)

I added your implementation of mscw to the code but I have a check in the there for its existence so it can still work correctly if mscw is not installed.

See https://github.com/01micko/retrovol/tree/speaker

I've been hoping to get back to my `xorg` branch but still a bit too busy with real life.

wdlkmpx:
01micko, the retrovol window looks ugly, too retro

edit lines 699-702, main.cpp to add 2 leading spaces so there is padding

		} else { //horizontal
			alignment = gtk_alignment_new(1.0,0.5,0,0);
			gtk_box_pack_start(GTK_BOX(vbox), alignment, false, false, 0);
			strcpy(wrapped, "  ");
			strcat(wrapped, list.items[i]->short_name);

--

MSCW can now identify and initialize new hardware, the previous
method relied on 10alsa's log, but this new one is flawless.
1) Source 10alsa (. /etc/init.d/10alsa)
2) Make sure all soundcard IDs are present in /etc/asound.state
3) If it's not the case call set_volume_levels() for every new soundcard found.

The automatic stuff calls this function everytime is triggered
The GUI/cli calls this function when a new card is selected

01micko:
@wdlkmpx 

yes, I see re alignment.

IMHO, a better way,

```c
		} else {
			alignment = gtk_alignment_new(1.0,0.5,0,0);
			gtk_box_pack_start(GTK_BOX(vbox), alignment, false, false, 0);
			snprintf(wrapped, sizeof(wrapped), "  %s", list.items[i]->short_name);
		}
```

wdlkmpx:
Yeah that is a better way. I've never took the time to learn C/C++, I just googled something, but somehow I am able to identify where the code to be modified is, it also happened with deadbeef, when I was able to fix a bug (for myself) located in this file https://github.com/Alexey-Yakovenko/deadbeef/blob/master/playlist.c

mavrothal:
So what is the status of this pull? 
Specially regarding 3-buiddistro-Z, DISTRO_PKGS_SPECS-* and a first boot call.

wdlkmpx:
I tested a tahrpup and it came with the old mscw, so I to have check 3...Z personally.. everything is working fine and better than ever. I think the way things progress, specially where manpower is main lacking thing, is by hijacking apps putting them in repositories, the most important ones directly in the skeleton, removing all references in the DISTRO_PKGS_SPECS, that way development is centralized and focus on all derivatives at once, but I guess even for that it's too late..

mavrothal:
> I tested a tahrpup and it came with the old mscw, so I to have check 3...Z personally.. everything is working fine and better than ever.

:open_mouth:  I'm not sure I get that. 
What about 3builddistro? You did, you will or will not do? 
If you will not, would you rather someone else try it? 

> I think the way things progress, specially where manpower is main lacking thing, is by hijacking apps putting them in repositories, the most important ones directly in the skeleton, removing all references in the DISTRO_PKGS_SPECS, that way development is centralized and focus on all derivatives at once,

Well, that is the idea. But to use the rootfs-packages instead of rootfs skeleton. And instead of hijacking apps we hoped that the original developers will join in but besides zigbert, no one else did so far :cry:

wdlkmpx:
Pushed a new update.

This update takes into account #745 #747 and #751 

adrv for slacko and tahrpup

[adrv_slacko_tahrpup.zip](https://github.com/puppylinux-woof-CE/woof-CE/files/146562/adrv_slacko_tahrpup.zip)

mavrothal:
I'll pull because you'll keep changing it.
You may want to add it in the [default] (https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-code/3builddistro-Z#L307) rootfs-packages

PR #746, opened 2016-02-21
Include /usr/share/alsa from alsa-utils in skeleton

wdlkmpx:
From Ubuntu Xenial [http://packages.ubuntu.com/xenial/alsa-utils]

It looks like 10alsa does not always work the same when there is more
than 1 soundcard, I see in Precise 5.7.1 with k3.14.55 that it fails
to initialize the second card even when the logs say it did.

alsactl init <card> does always initialize any card with no errors..
but requires /usr/share/alsa/init/* from alsa-utils.

There is a totally insane rule in packages-templates/alsa-utils that
prevents /usr/share/alsa from being installed

There is no need to change that because it's coming here and:
1) these are text files
2) can be modified/updated easily to benefit all puppies regardless of version...
3) will be modified to adjust some settings..


mavrothal:
If you want to include more files from alsa-utils you can just modify its [template](https://github.com/wdlkmpx/woof-CE/tree/testing/woof-code/packages-templates/alsa-utils/usr).
However it would appear that puppy is doing mostly fine without it.
So what about if for these rare (?) cases we just instruct the user to install the alsa-utils package with PPM?

wdlkmpx:
This does not affect anything, it's just part of the changes I plan to make to 10alsa, right now the code is in MSCW...it may stay there.

If there is more 1 soundcard, then run alsactl init 1 ... alsactl init 2, etc...  [PUPMODE 5] when the


			if [ $PUPMODE -eq 5 ] ; then
				if [ $scards -ge 2 ] ; then 
					#two or more cards, need to initialize second card, third card...
					newcards=$(grep '^ [0-9]' /proc/asound/cards | grep -v pcsp | tail +2 | grep -o '^ [0-9]')
					for cardx in $newcards ; do
						alsactl init $cardx #-d
					done
				fi
			fi

wdlkmpx:
Initialize means.. set initial volume levels, 10alsa does set them or so it seems, but when I change card with MSCW, the current card is muted.. not initialized, but I can certainly close this and add a rule to the template, although if EVERY puppy is using alsa-base.conf from Ubuntu Precise from the alsa-base package, I see no harm in this...

wdlkmpx:
Actually this can be overwritten when someone install a more complete package, or when someone has already those files ready to be applied, as mainstream puppies come with this feature of lacking some important files.

Now that I explained everything... to avoid long waits... what do you think mav, should I edit the template instead?

mavrothal:
> Now that I explained everything... to avoid long waits... what do you think mav, should I edit the template instead

As  mentioned above without puppy did OK for the most part and with your changes in 10alsa, alsa-base.conf and mscw (you have to finish that pull) is doing even better. I would leave it out, but if you could add a function in mscw to popup a message suggesting to install alsa-utils from the repo when it "sees" such a need would be even better.

PR #747, opened 2016-02-22
10alsa: initialize each soundcard individually

wdlkmpx:
this does seem to work everytime

### DIFF
--- a/woof-code/rootfs-skeleton/etc/init.d/10alsa
+++ b/woof-code/rootfs-skeleton/etc/init.d/10alsa
@@ -12,16 +12,14 @@
 #120226 01micko: added 'Master Front' entries. 120318 bumped it up, 75% to 100%.
 
 function set_volume_levels() { #card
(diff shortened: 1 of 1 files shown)


mavrothal:
This looks good but I have no time to test it.
Does it work OK with your systems?

wdlkmpx:
I actually spend all my free time testing.

Everything that I do in precise in my oldest hardware works the same in newer versions/more modern hardware.

I tested in tahrpup 6.0.5 with adrv_ had to add my xorg.conf-auto-pc. It still has /root/Starup/alsa.. the script I suggested long ago to hide the real problem that I'm solving this year... I regret doing that.

Without adrv_.sfs (with my changes) ... 2nd card muted.

With adrv with all my changes (10alsa, delayedrun, MSCW).. 2nd card initialized just like the first one. On second boot the settings are still there.

mavrothal:
Sold :smile:

PR #748, opened 2016-02-23
initmodules - automatically determine needed modules

gyrog:
A complete rewrite.
It now determines the modules that need to be loaded in "init", including dependencies, by examining the modules that are loaded in the current running system.
Executed with a "-q" parameter, it displays nothing, but writes a config file if needed, and if it can.
Otherwise it is a GUI utility that displays informative messages and it's list of needed modules. The list can then be modified or accepted as is.

### DIFF
--- a/woof-code/huge_extras/initmodules
+++ b/woof-code/huge_extras/initmodules
@@ -1,68 +1,224 @@
 #!/bin/ash
 
+. /etc/DISTRO_SPECS
 . /etc/rc.d/PUPSTATE
 
-if [ "$PUPMODE" = "5" ];then
-	Xdialog --wmclass "module16" --title "$(gettext 'Initmodules: Modules to load in init script')" --msgbox "$(gettext 'You must shutdown and create a personal save file first.')" 0 0
-	exit
-fi
+QUIETMODE=""
+[ "${1}" = "-q" ] && QUIETMODE="yes"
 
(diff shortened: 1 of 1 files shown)


mavrothal:
Is it currently called form anywhere  with the -q argument? 
Maybe add a call in [quicksetup](https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-code/rootfs-skeleton/usr/sbin/quicksetup) when in pupmode 5 ?
Also  if I remember the /tmp/$CFGFN  use by rc.shutdown or shutdownconfig is not implemented yet.
Is it in the planing?

gyrog:
mavrothal,

Thanks.

gyro

gyrog:
mavrothal,

On 23/02/16 23:45, mavrothal wrote:
> Is it currently called form anywhere with the -q argument?
No.
> Maybe add a call in quicksetup
> <https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-code/rootfs-skeleton/usr/sbin/quicksetup>
> when in pupmode 5 ?
Maybe look at this later.
> Also if I remember the /tmp/$CFGFN use by rc.shutdown or shutdownconfig
> is not implemented yet.
> Is it in the planing?
Yes, rc.shutdown immediately after copying files to the new save layer.
> 
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/puppylinux-woof-CE/woof-CE/pull/748#issuecomment-187705281>.
> 

gyro

PR #749, opened 2016-02-24
Add initmodules support for first boot and first shutdown

gyrog:
These patches enable first boot to create an initmodules config file and for first shutdown to copy any initmodules config file found in /tmp into the same directory as the save layer.
(mavrothal is the real author of the quicksetup patch.)


gyrog:
mavrothal,

Thanks.

gyro

PR #757, opened 2016-03-07
SNS: show Access Point MAC address

wdlkmpx:
This helps identifying the correct network.
ex: when the SSID is hidden (#756)

To make room, the 'Strength: XX/XX' string has been removed,
there is already a nice image...


01micko:
Sorry I missed this one

PR #758, opened 2016-03-23
fix psavemark to work with pmedia and psave code

gyrog:
Unfortunately my original version of this patch was the wrong code.
Both versions work, but this makes the code correspond to that agreed in the forum.


gyrog:
Mick,

Thanks.

gyro

PR #775, opened 2016-05-15
0setup: use ./support debdb2pupdb|find_cat if building from woof

wdlkmpx:
otherwise /usr/local/petget/* to update dbs in a running pup

the files in ./support are supposed to be always the latest version

### DIFF
--- a/woof-code/0setup
+++ b/woof-code/0setup
@@ -89,8 +89,13 @@ if [ -f ./DISTRO_SPECS ];then
       echo -e "\n\nlib64 directories created in packages-templates."
     fi
   fi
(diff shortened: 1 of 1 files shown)


01micko:
I think it would be safer to use the absolute path.

wdlkmpx:
I redid the commit: $(pwd)/ instead of ./, also placed the variables a few lines above.

I tested with the ubuntu xenial files provided by mrfricks. I have an old version of debdb2pudb in /usr/local/petget... the difference is this:

old /usr/local/petget/debdb2pupdb .. incredibly slow ... after 2 minutes freezed the system 
./support/debdb2pupdb... it took a few seconds

01micko:
Thanks

PR #776, opened 2016-05-15
Minor UI enhancements for sns

Mr-Kumar-Abhishek:
:neckbeard: sorry, but I am getting a bit interested in UI's of *gtkdialog* thought I might contribute a bit while learning it .. between where are the `primary-icon-name`s are stored ?? 

sns

Hopefully everything is okay, I have tested them in puppy Linux - **tahrpup**

### DIFF
--- a/woof-code/rootfs-skeleton/usr/local/simple_network_setup/sns
+++ b/woof-code/rootfs-skeleton/usr/local/simple_network_setup/sns
@@ -515,7 +515,7 @@ if [ "$IF_INTTYPE" == "Wireless" ];then
          <vbox space-expand="true" space-fill="true">
            <hbox space-expand="true" space-fill="true">
               <vbox>
-                <text xalign="0" space-expand="true" space-fill="true"><label>'$(gettext 'Key:')'</label></text>
+                <text xalign="0" space-expand="true" space-fill="true"><label>'$(gettext 'Password:')'</label></text>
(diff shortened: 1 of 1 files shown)


wdlkmpx:
Looks goo to me. 

gtkdialog takes icons from current icon theme which should be in /usr/share/icons/theme, well at least that's the way it is in my system.

Also the custom icons that are widely used in woof ce are in /usr/share/pixmaps/puppy, all of them are .svg icons, there are scripts in /usr/lib/gtkdialog you might want to take a look at

Mr-Kumar-Abhishek:
amended  commit message 47b53c7. :smile: , Thanks for the info @wdlkmpx , I will take a look at those scripts.

zigbert:
Great to see your gtkdialog interest.
The tooltip is not gettexted, and because of NLS, I suggest gtk should handle alignment. Just put all text in one line, and if required gtk will split it.

Markups are cool, but personally, I think tooltip markups should be used where _really_ important stuff is to say. Here it is just repeating the obvious, and markups is a bit overkill. Most important is that it reduces the effect where we actually need it. Also, markups clutters the use of gettext. Either we have to include markup code in the gettext string, or we have to split the string...

I suggest:
tooltip-markup="'$(gettext 'Enter your network password')'"

Mr-Kumar-Abhishek:
@zigbert  done.. 

sns2

01micko:
Ok :smile_cat:

PR #779, opened 2016-05-25
Correct search for top menu category

rserwin1:
The desktop-file category, "Utility", erroneously finds "Setup" as its top menu category because of the entry, "X-SetupUtility".  This change corrects the TOPCAT grep patterns to find only complete category text strings, for hierachy files with and without comments on each line.


rserwin1:
Add frisbee package to rootfs-packages

wdlkmpx:
You can't. You have to create another branch and add commits. Learning how git works takes time, it's very confusing at first, and it's still confusing once you reach a decent level and want do more complex stuff. It's particularly confusing once your branch is behind and has diverged to do unmerged commits... that is time to use the cli to force undo commits and pull the new ones from "upstream".

Adding and manipulating stuff works better in cli, you add a whole directory and then in one commit you add all your changes.

Something like this:
1) add frisbee directory to rootfs-packages
2) terminal window in woofce root dir: git add .
3) git commit -a -m add frisbee
4) git push origin testing [you have to setup stuff first]
5) Go to the github.com/<yourusername> go to your woofce repo
6) You'll see that you're 1 commit ahead and you can create a pull request..

This takes time to learn, and once you learn you'll end up typing git commands very often to the point that often it takes longer to add and push a commit than the changes you've made. That's why I think there should be a dedicated topic in the forum where people can request to add changes, etc. Some like "woof-ce commit requests, that way those of us who have already wasted a lot of time on git can make pull requests as fast as possible and possibly the team itself may add the commits

01micko:
I can fully understand why anyone would _not_ want to learn the ins and outs of git CLI!

@zigbert has addressed this somewhat [here](http://blog.puppylinux.com/?viewDetailed=00023)

01micko:
@rserwin1 the commits look ok from a quick glance. I haven't much time in the next 48 hrs so I'll take a closer look then.

Thanks!

rserwin1:
Mick,
Thanks for the link to zigbert's explanation.  He mentions a  "*Propose 
file change* button.", which I have not seen on the edit or upload pages.

I used a similar technique, though, with frisbee and pgprs, except that 
I clicked the "Commit" button (the only other button being "Cancel").  
And I avoided having to paste content into the initial file by leaving 
it empty, committing it, then uploading the correct file, to replace 
it.  That seems less error-exposed than possibly pasting the wrong content.

FWIW, the frisbee example is erroneous in that the file belongs in 
/etc/frisbee/, not /etc/.  Please remove those ".tmp" files from the 
testing branch, because it conflicts with the actual structure of frisbee.


Mick Amadio wrote:

zigbert:
Commit changes / Propose file change is the same button - depending on your rights.
I have updated the blogpost

zigbert:
frisbee tmp files is removed

PR #781, opened 2016-05-28
Add new pgprs package; correct connectwizard test; remove old pgprs API

rserwin1:
Adds pgprs-2.0 package, removes test for presence of /usr/sbin/pgprs-shell from connectwizard and deletes pgprs-shell altogether.
I could not find the old pgprs directory which used to be in /lib/modules/all-firmware.  I see that all-firmware became "modem-firmware", but I could not find that, either.  That directory should be removed, wherever it is.  However, to preserve the old/original version, it could be made into a pet package.  Let me know if you want me to make such a pet package, if one has not already been made.

BTW, this includes the first 3 commits on May 27, which should be removed or ignored.  I inadvertently tried to update connectwwizard in kernel-skeleton instead of root-skeleton.  I created an empty file, uploaded my new version, then deleted it.
R


01micko:
Never mind about those deleted files, they show up in the commit log anyway; doesn't matter.

I'll merge this but what do think about changing the icon to /usr/share/pixmaps/puppy/phone_mobile.svg or wireless.svg?

The reason is that mini-icons look terrible if the user has a touch screen theme and the jwm menu is enlarged to 32px (or more).

rserwin1:
I am open to that.  I will add that to my branch and create a new puull request, now that you have merged the this pull.  The "wireless" icon looks good.  That will be consistent with the icon in connectwizard. 
R

01micko:
It is best to use the full path for the svg icons. :smiley:

PR #783, opened 2016-05-28
Change pgprs window/menu icon to svg; fix modem detection test

rserwin1:
Previous message: "My pgprs package includes a relative link to the wireless.svg icon.  I am not allowed to upload that link file.  The message begins with "Something went really wrong ... cannot upload file" (or such).  If that cannot be done in the web interface, how should I do it another way?"

Recoded icon link generation,  Ready for inspection and consideration for merge.
The wireless.svg is used if present, otherwise the Pwireless.png icon is used.


01micko:
> My pgprs package includes a relative link to the wireless.svg icon. I am not allowed to upload that link file.

I guess this is a limitation with the GUI interface. IIRC @zigbert has had the same issue. Can you create the link via the pinstall.sh?

Otherwise, I can merge this PR and create the link on CLI then push the change.

Your call.

rserwin1:
Thanks, Mick.  I think I will make a null-file placeholder and change it to a link in the pinstall script.  I want some sort of file in place so that an entry is created in the packages pgprs...files file.

PR #800, opened 2016-06-10
Update AppRun and rc.sysinit

rserwin1:
AppRun was left out of the interface updates related to connectwizard.
Rc.sysinit is updated to support USB3 driver module and to check for frisbee as the active network manager.
AppRun requires execute permission, but rc.sysinit does not.

### DIFF
--- a/woof-code/rootfs-skeleton/etc/rc.d/rc.sysinit
+++ b/woof-code/rootfs-skeleton/etc/rc.d/rc.sysinit
@@ -91,6 +91,8 @@
 #130618 3builddistro sets 'DEVTMPFSFLG' variable.
 #130629 need directory /tmp/pup_event_ipc, to support new pup_event IPC.
 #130630 bring back tmpfs on /tmp for full HD installation.
+#160609 rerwin: Add wait for USB3 driver.
+#160609 rerwin: ignore frisbee as default if not active.
 
 #unset TZ #100319 busybox hwclock gives priority to this (rather than /etc/localtime) and 'init' has set it wrong.
 #...comment-out for now. note, TZ now set in rc.country.
@@ -498,7 +500,7 @@ if [ "$PCIPCMCIA" != "" ];then #this may be slow to respond.
   echo -n " $WAITCNT" >/dev/console
  done
 fi
-PCIUSB="`elspci -l | grep -o -E '0C0300|0C0310|0C0320'`"
+PCIUSB="`elspci -l | grep -o -E '0C0300|0C0310|0C0320|0C0330'`" #160609
 #note, if initrd then usb-storage will have already loaded...
 USBBUILTIN='no'
 if [ "`modinfo ehci_hcd 2>/dev/null`" = "" ];then #110712 130201 shut this up.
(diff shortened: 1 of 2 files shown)


wdlkmpx:
Some radical changes are happening in another branch ( https://github.com/puppylinux-woof-CE/woof-CE/tree/rationalise ) where rc.sysinit has been replaced with the contents of woof-code/huge_extras/rc.sysinit ( https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-code/huge_extras/rc.sysinit )

This situation led to many confusion in the past, so it's being addressed but the outcome will take some time, some breakage is about to happen thanks to me.

Also there is a blog post 
http://blog.puppylinux.com/

micko should open a thread about this on the forum

wdlkmpx:
It merged into "rationalise" quite well..

The git commit history shows the author, so there's no need add extra comments, also to prevent the huge wall of comments about every change change at the beginning of every puppy script, a comment should be added to the corresponding code.. but to explain what it does, or clarify, something like this:

	#                               unknown|unknow|unknow|usb3
	PCIUSB="`elspci -l | grep -o -E '0C0300|0C0310|0C0320|0C0330'`"

rserwin1:
Regarding the history comments at the beginning of some scripts, are we 
to no longer track history there?  That does make sense if we now 
consider the git repository as maintaining the history.

As far as embedding explanatory comments within the code, I have been 
taking my cue from Barry's apparently minimalist approach to that, 
probably to keep scripts from getting too large, with comments.  I am 
fine, with commenting in place.

Regarding commit comments, I would add them only if I needed further 
explanation pertinent to integrating a fix.  But for the pull request, 
that is where I try to "sell".the fix or to add detail for the request.


wdlkmpx wrote:

wdlkmpx:
Well, when I read a script, i rarely look at the comments unless there
is some kind of design explanation, I just try to understand it as is,
and I  usually also delete commented out code to make the screen look
cleaner, change identantion, and it's ready to be read by me.

And when I write code, I rarely comment.. of course that's my style.
When I comment I only do it in place, most code doesn't need
explanation, it just needs to be well written and formatted, for some
obscure details... comments are needed, and that's when I comment.

Of course that's my style, woof-CE hasn't defined a coding style yet..
and I doubt it will ever do..

01micko:
Well I like everything nested tab (4 spaces) indented which I believe is fairly common.

As for comments, Barry used to put the changelog at the top of the script. This goes back to the 'unleashed' build system which was not in a VCS. Also Barry had the habit of just commenting out old code and replacing it with new code in the next block - not ideal. It just adds bloat.

As for general comments, here's my take - (and my memory isn't the best!)

> Comments are a Good Thing. You may have understood the script perfectly when you wrote it. But six months from now it could look like modem noise.

by [Bruce Barnet](http://www.grymoire.com/Unix/Sed.html)

rserwin1:
I concur with your style, completely.

wdlkmpx wrote:

rserwin1:
I will keep in mind the need for "memory joggers".  I see and have used 
indentations of both one and four spaces.  I think four is better 
because that leaves single-space indentation for continuation of 
statements using "\".  I wil go that way.


Mick Amadio wrote:

PR #802, opened 2016-06-12
Kicked RHEL and Scientific Linux support

dimkr:
Anything I've missed?


wdlkmpx:
It looks good, if there is anything else.. it's probably in packages-templates, but that requires a closer look, I removed some stuff with the mageia label though

PR #808, opened 2016-06-15
Update Packages-puppy-noarch-official

peabee:
get_libreoffice-0.40


01micko:
Thanks! In future though (after we sort out `rationalise` branch) I think it should go to `rootfs-packages`.

peabee:
If rootfs-packages continues to grow there needs to be a better 
mechanism in 3builddistro-Z to select items from the multiple choices

Best would be an extension to the syntax of DISTRO_PKGS_SPECS-** to 
specify packages are obtained from rootfs-packages

2nd best would be to have a separate config file with choices in it

3rd best would be to have existing check box mechanism but have all 
items selected by default...

PR #810, opened 2016-06-15
Delete pinstall.sh

peabee:
pinstall cannot have exit 0 as this kills combined sandbox3/pinstall.sh


01micko:
Yes these are fatal bugs but deleting the pinstall isn't the answer. Same with #811

01micko:
Ok, here is a work around which seems like a reasonable solution

```diff
--- ../woof-CE/woof-code/3builddistro-Z	2016-06-16 09:02:54.677229957 +1000
+++ 3builddistro-Z	2016-06-16 09:32:50.440504871 +1000
@@ -403,8 +403,10 @@ if [ "$EXIT" = "OK" ];then
 		if echo $line|grep -q "true" ;then
 			echo "copying ${line%%=*}"
 			cp -a rootfs-packages/${line%%=*}/* sandbox3/rootfs-complete/
-			[ -f rootfs-packages/${line%%=*}/pinstall.sh ] &&\
-				cat rootfs-packages/${line%%=*}/pinstall.sh >> /tmp/rootfs_pkgs_pinstall.sh
+			if [ -f rootfs-packages/${line%%=*}/pinstall.sh ];then
+				grep -qw 'exit' rootfs-packages/${line%%=*}/pinstall.sh ||\
+					cat rootfs-packages/${line%%=*}/pinstall.sh >> /tmp/rootfs_pkgs_pinstall.sh
+			fi
 			cat sandbox3/rootfs-complete/pet.specs >> /tmp/rootfs-packages.specs
 		else
 			continue


```

If 'exit' is present in the pinstall.sh it doesn't get copied. We should use this mechanism for _any_ pinstall just in case the casual woof user isn't aware of that.

PR #813, opened 2016-06-15
Delete puninstall.sh

peabee:
not required in woof-ce


wdlkmpx:
This is correct, however rerwin might up uploading that file again, so we should communicate this to him

01micko:
The thing is, the packages should reflect actual pets, so pinstalls and puninstalls should be the same as what is found outside woof; 'puninstalls' get deleted anyway.

rserwin1:
Thank you, Mick, for confirming what I have said in a PM to peebee, that the entire source of a pet package should be maintained as the official and complete source.

PR #818, opened 2016-06-16
Update Packages-puppy-slacko14.1-official

ninaholic:
ppm2 fails to find the mplayer2 when selected, but I found it manually in older folder :santa: 

http://distro.ibiblio.org/puppylinux/pet_packages-slacko/
http://distro.ibiblio.org/puppylinux/pet_packages-slacko14.1/


01micko:
That package was compiled for 13.37! (slacko-5.3x)

ninaholic:
haha yes 01micko! the problem is on Slacko 6.3.0.6 you can click it in ppm2, but then it gives error because it's in the list for 14.1 but not in the repository :D

wdlkmpx:
mplayer2 is a dead project.

The best media player for gnu/linux is mpv which is the successor of mplayer/mplayer2, it works flawlessly with old hardware and new... or at least up to certain commit, i used it with different kernels/drivers and setups and it always worked perfectly... mplayer comes second and VLC being the worst of the 3 when it comes to adaptativeness.

I might add a petbuild, which is about 300 lines for a fairly static, functional and portable mpv... if compiled in precise.

PR #834, opened 2016-08-03
Dead link in PPM help

ninaholic:
Is linking to old version wise? Or maybe creating a newer info page on puppylinux.com would be more appropriate? :octocat:


wdlkmpx:
That is really old. It still contains some relevant info though. I doubt anyone is writing docs nowadays

01micko:
@ninaholic - sorry off topic but I can't find the conversation where you
pointed out that xarchive and xdelta packages in 700 series were broken.
I know it isn't best practice but I have reuploaded the offending packages
with fixed packages (32 and only xdelta in 64 as xarchive had the correct
tar wrapper); overwriting the old versions so the version number remains
the same. I haven't released yet so I figure this is ok :P. Just deleted
the old ones, run 1download and 2createpackages on those and you should be
good to go.

The build bugs are also fixed at petbuilds - slacko_142 branch.

01micko

ninaholic:
Thanks 01micko! I use your slacko-14.2 mixed with some 14.1 (alsamixer) and 14.0 (jwm, xchat) things as my main OS for three weeks now. Working well on my Pentium M laptop :sunny:

PR #864, opened 2016-08-26
Revert hostname-set

ninaholic:
Is there any benefit to it? reverting fixes Rox bugs around since Slacko 5.7

http://www.murga-linux.com/puppy/viewtopic.php?p=906343#906343
http://www.murga-linux.com/puppy/viewtopic.php?p=908737#908737

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/hostname-set
+++ b/woof-code/rootfs-skeleton/usr/sbin/hostname-set
@@ -83,7 +83,7 @@ $(gettext '\ZbNOTE:\ZB There did appear to be an active network connection, that
 
  hostname $NEW_HOSTNAME
  echo -n $NEW_HOSTNAME > /etc/hostname
- echo "127.0.0.1 $NEW_HOSTNAME localhost" > /tmp/hostname-set-hosts
+ echo "127.0.0.1 localhost $NEW_HOSTNAME" > /tmp/hostname-set-hosts
  grep -vw 'localhost'  /etc/hosts >> /tmp/hostname-set-hosts
  [ -s /tmp/hostname-set-hosts ] && mv -f /tmp/hostname-set-hosts /etc/hosts
  M_h1="$(gettext 'Set Hostname: done')"


wdlkmpx:
People contradict each other in that thread and it looks like there is no "consensus"

ninaholic:
This is the interpretation I get:
- No known bugs happen with the original
- Bugs were found from changing it
- Perhaps these new bugs are re-producible on all systems, or maybe not?

It would seem to make more sense to revert it unless there is an actual bug found with the original then, based on this? :laughing:

wdlkmpx:
so by reading rcnscs51's comment at least i learned something today.

maybe hostname-set should offer to restart X? it also needs a woofce look, very easy to apply.

01micko:
I meant to write to that thread but I have been side tracked by real life for a while. IDK why mav pulled in the first one. Seems to me in /etc/hosts it should be ipaddress > hostname > alias 
In the case of 127.0.0.1 'localhost' is the hostname and the actual hostname is an alias. There are tons of references on the web to /etc/hosts.

Probably the other crud in the shipped /etc/hosts file should be removed; the pc1, pc2, pc3 stuff. You can actually map these to addresses on your lan and ping them by hostname then.

eg I add 192.168.1.20 slacko32-intel

```
# ping -c3 slacko32-intel
PING slacko32-intel (192.168.1.20): 56 data bytes
64 bytes from 192.168.1.20: seq=0 ttl=64 time=0.763 ms
64 bytes from 192.168.1.20: seq=1 ttl=64 time=0.472 ms
64 bytes from 192.168.1.20: seq=2 ttl=64 time=0.293 ms

--- slacko32-intel ping statistics ---
3 packets transmitted, 3 packets received, 0% packet loss
round-trip min/avg/max = 0.293/0.509/0.763 ms
```

In first run X is supposed to be restarted, but it isn't forced so if a user doesn't then they'll get the bug - in any system if they change the hostname.

01micko:
Bingo!

X gets the hostname from utsname (see /usr/include/sys/utsname.h and /usr/include/X11/Xtrans/xtrans.c) so this "bug" is actually a faux bug caused by not restarting X. Nobody ever disclosed whether they actually _did_ restart X in that thread. It has absolutely nothing to do with /etc/hosts. The file /etc/hosts is only to do with networking. (That said, I believe this pull request should be pulled anyway because I reckon it's wrong and pending further investigations I think the whole /etc/hosts thing could be improved dramatically).

ninaholic:
Thanks @01micko for the detailed explanation! :smile:

01micko:
I'll probably cherry pick this for `rationalise` branch

Issue #873, opened 2016-09-05
Slacko woof-out explosion from glibc


ninaholic:
(thanks for reminding me about this one with https://github.com/puppylinux-woof-CE/woof-CE/issues/872 report 01micko! :bulb: )

The end of this post ("note:") explains what happens in more detail:
http://www.murga-linux.com/puppy/viewtopic.php?p=896200#896200

Before:
http://oi65.tinypic.com/2vkwyol.jpg

After:
http://oi63.tinypic.com/2dw59xw.jpg

Can other Slacko builders see these? Just deleting the red circles after seems to work ok.

wdlkmpx:
I have something in mind to address this.. basically a _PRE file and a custom hack-install.sh_${DISTRO_FILE_PREFIX}_${DISTRO_COMPAT_VERSION} to replace the generated hack-install.sh..

01micko:
@ninaholic yes that glib bug in woof has existed for years. I've been meaning to look at it but haven't as yet. I'll see if https://github.com/puppylinux-woof-CE/woof-CE/commit/51f124607c58033572fa4ef3d27bf4d36adff088 fixes it

ninaholic:
Thanks @wdlkmpx and @01micko ! Yes the bug was around since my first build attempt haha. I changed mv to cp on lines 727 and 729 on https://github.com/puppylinux-woof-CE/woof-CE/commit/dfba4ecf80838886ffaa61272ec87842c81fe48b in case I want to rebuild again with newer zoneinfo. It seems to work ? :smile:

ninaholic:
Thanks @wdlkmpx and @01micko :smile: The PRE file idea is working for me. :+1:

PR #874, opened 2016-09-05
Update 2createpackages

ninaholic:
In case you want to rebuild glibc more than once


01micko:
Is this still wanted since #873 is closed?

ninaholic:
Yes @01micko , sorry for bad communication. I think @wdlkmpx fixed the problem, but we can reopen https://github.com/puppylinux-woof-CE/woof-CE/issues/873 if we discover a bug with it :laughing:  mv deletes the file after first compile so I think cp makes more sense.

01micko:
Ok thanks @ninaholic

wdlkmpx:
cp is the command. i thought i had updated the file before pushing

PR #875, opened 2016-09-07
Make wpa default in sns

peabee:
wpa is these days the default for wifi so make it default in sns


zigbert:
I suggest we keep the original sort order (of the radiobuttons)
Use instead <default>true</default> for wpa to make this one the defualt

PR #876, opened 2016-09-21
dir2pet: Add user control of compression default; align & edit help output

rserwin1:
…lp output


wdlkmpx:
i tested this and it works as advertised..

PR #886, opened 2016-10-10
xrandr and xorg compatibility fix

ninaholic:
If the user uses xrandr to change resolution, sometimes magic is needed for xorg changes to stick. Feel free to push and adjust better. I also change xorg icon to the most appropriate one and for better variety (do you agree? :D )

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/xorgwizard
+++ b/woof-code/rootfs-skeleton/usr/sbin/xorgwizard
@@ -186,6 +186,8 @@ fi
 
 if [ "$CLI" = "1" ] ; then
 
+rm -f /etc/xrandrindex
+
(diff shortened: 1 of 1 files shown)


wdlkmpx:
I think the xrandrshell code was not a good idea. A tiny and faster app that can handle even multiple monitors was always at hand: LXrandr

# filesize /usr/bin/lxrandr 
21K

dependencies: gtk2 and x libs

https://wiki.lxde.org/en/LXRandR

I'll take a look at the changes..

ninaholic:
Thanks @wdlkmpx , if xorg is used after xrandr I found that /etc/xrandrindex needs to be deleted or sometimes xorg won't work. Maybe there is a better way to keep both in sync.

PR #890, opened 2016-10-16
"Darn Adobe at it again with their URL changes"

ninaholic:
Courtesy of 01micko :laughing:
http://www.murga-linux.com/puppy/viewtopic.php?p=928449#928449

Might change again soon if Flash 23 npapi goes stable?

### DIFF
--- a/woof-code/rootfs-packages/getflash/root/.getflash/getflash-conf
+++ b/woof-code/rootfs-packages/getflash/root/.getflash/getflash-conf
@@ -23,8 +23,8 @@ _V11_64_size="6.9"
 _V11_64_url="http://sourceforge.net/projects/asriedu/files/asriedu_packages/asriedu_packages_31x_base_puppyprecise5xx"
 
 ## Adobe URLs
-CHECKURL_32='://get.adobe.com/flashplayer/download/?installer=FP_11.2_for_other_Linux_32-bit_(.tar.gz)_-_NPAPI&standalone=1' 
-CHECKURL_64='://get.adobe.com/flashplayer/download/?installer=FP_11.2_for_other_Linux_64-bit_(.tar.gz)_-_NPAPI&standalone=1' 
+CHECKURL_32='://get.adobe.com/flashplayer/download/?installer=FP_11.2_for_other_Linux_32-bit_(.tar.gz)_-_NPAPI&sType=2721&standalone=1' 
+CHECKURL_64='://get.adobe.com/flashplayer/download/?installer=FP_11.2_for_other_Linux_64-bit_(.tar.gz)_-_NPAPI&sType=2721&standalone=1' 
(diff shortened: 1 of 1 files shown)


ninaholic:
With .643 it is now Type=2732. I'm just using @rserwin1 pet instead
http://www.murga-linux.com/puppy/viewtopic.php?p=929390#929390

wdlkmpx:
http://arstechnica.com/information-technology/2015/10/firefox-dropping-npapi-plugins-by-the-end-of-2016except-for-flash/

At this point it might be better to send that file to this location:
http://distro.ibiblio.org/puppylinux/pet_packages-noarch/

... and delete the woofce ''getflash" rootfs-package

wdlkmpx:
peebee or somebody should upload rerwin's pet pkg to noarch and add a corresponding entry

i'll delete the rootfs-package if micko doesn't mind.. unless there is another plan..

rserwin1:
I can do that, unless someone else wants to.  Just give me the go-ahead.
rserwin1


wdlkmpx wrote:

rserwin1:
I misunderstood what was being requested.  I cannot do what I offered.
rserwin1

wdlkmpx wrote:

wdlkmpx:
well, the pkg is available for new builds: https://github.com/puppylinux-woof-CE/woof-CE/commit/8f2fd6062e71a8bad1334e3306c2668034868cd7

if the rootfs-pkg is going to stay, maybe an hybrid between the 2 the makes sense, or just replace it with the pet pkg from the repo

peabee:
I think I'm going to replace getflash in my build with a prompt to download the latest .pet from noarch......

wdlkmpx:
so what does micko think.. should i replace the rootfs-pkg with
rerwin's code or maybe should i delete it?...

peabee:
getflash in rootfs-pkg was already at version 1.5.8 so the new version really needs to be 1.5.9.....

rserwin1:
I bumped the version to 1.6 because the method of accessing the adobe 
site got changed to Geoffrey's technique.  I think the 1.5-? format has 
gotten overworked.  My version is based on 1.5-6, without the 
enhancement of the automatic bootup-time that seems to be treated as a 
minor fix, in 1.5-7, which it isn't.

I recommend that we wait for Argolance's integration of the automatic 
feature and new flash into 1.6 and call it 1.7. because it is a 
significant upgrade from 1.5-6 and 1.6.
rserwin1


PB wrote:

ninaholic:
The v1.6.1 pet is working for me, and downloaded and installed flashplayer-24.0.0.194 successfully.
http://www.murga-linux.com/puppy/viewtopic.php?p=936597#936597

I think replacing v1.5.8 one with v1.6.1 in Woof-CE is a good idea now that 24 is "stable", so that when you select "getflash" in ./3builddistro-Z it actually works. :laughing:

PR #900, opened 2016-11-08
Raspberry Pi Build

woodenshoe-wi:
Its not very polished, 3builddistro-Z will insist on downloading one of the "Huge_Kernel" pets, but it boots to desktop on both my pi1 and pi2.

I haven't done any x86 test build, and if you want to wait until one of your test builds I'm not in any hurry.  I saw the dates on some of my old commits and I haven't worked on this for almost a year.


wdlkmpx:
i'll merge first and test later.

i think i'll move the code to download/use huge kernel to 'support/huge_kernel' or something like that. and if BOOT_BOARD='raspi' then do something else..

PR #924, opened 2016-12-22
Update raspbian/jessie to 6.0.4

woodenshoe-wi:
Add firmware for Pi3 wifi.
raspbian/jessie/patches/no-delete-brcm_pi3.patch is only needed
for raspberry pi builds because they do not have a z drive.

I did a test build and everything seems to work fine with xorg_base_new except I can no longer CTRL+ALT+BACKSPACE to exit X.  I can still CTRL+ALT+F2 to get to the command line so it is not a problem if this is intentional, but /etc/X11/xorg.conf still has `Option "DontZap" "false"`


wdlkmpx:
indeed CTRL-ALT-BACKSPACE was not working, but commit https://github.com/puppylinux-woof-CE/woof-CE/commit/a21839617d6557fd5fc6ba26a8b956eee36b7cfc fixes the issue for me..

PR #929, opened 2016-12-28
fix by radky for /usr/sbin/bootmanager .... as wasn't working

mrfricks:


wdlkmpx:
I just noticed you're releasing betas, then i'll merge rationalise into testing. So that unstable development continues in rationalise and patches go to testing, don't forget to update the upup puppies "build recipes" ..

I think the code that handles desktop files in 2createpackages (woof) and installpkg.sh (petget) needs some updating. Oh and tell newpet that the encrypted savefiles are no longer supported, there is an explanation in the slacko 700 beta thread.. but i guess you can't reason with a puppy user, that's why i feel a little guilty and want to go my own way..

PR #939, opened 2017-01-08
Super Key + L to lock computer?

ninaholic:
I use this "windows refugees" shortcut sometimes to lock my laptop in public places, especially now as some newer Slacko pthemes don't have the lock icon on desktop any more.


ninaholic:
Trying to find how to update rootfs-packages/jwm_config/root/.jwm/backup/jwmrc-personal
too in the same commit, I know the option to edit more files is hiding somewhere! :rofl:

PR #949, opened 2017-01-27
remove libtsan0 from x86 xenialpup?

ninaholic:
Apparently this cannot be found in 32-bit builds because it doesn't exist
http://www.murga-linux.com/puppy/viewtopic.php?p=918353#918353
http://packages.ubuntu.com/search?arch=any&searchon=names&keywords=libtsan


wdlkmpx:
I appreciate that you spend some time building pups and opening pull requests. This is how I imagine the CE part.

I think I'm going to bring back older slackware versions. Although applying the reversion kits (1,2,3) for slacko 700 - slackware 14.2, it's possible to do basically everything that's needed for older hardware. I can do this and provide sfs's or pets, but of course this will take me some time, and I don't have enough time.

Some people might want older slackware versions because  they're quite smaller or something. Don't know.. but you might want to do test builds and release them in the wild, but first we have to try and solve all the major issues directly in woof. What do you think?

PR #960, opened 2017-02-10
aufs-utils url change

peabee:


kernel-kit/build.sh
```diff
@@ -197,7 +197,7 @@ FW_URL=${FW_URL:-http://distro.ibiblio.org/puppylinux/firmware}
 # $package_name_suffix $custom_suffix $kernel_ver
 kernel_version_full=${kernel_version}${custom_suffix}
 kernel_srcsfs_version=${kernel_version}
-aufs_utils_git="git://git.code.sf.net/p/aufs/aufs-util.git"
+aufs_utils_git="https://git.code.sf.net/p/aufs/aufs-util"
```
mavrothal: Why? https works but the original works fine here too (and is the dedicated one for git).
 What is the error that you get  with ``` git clone git://git.code.sf.net/p/aufs/aufs-util.git```?

ninaholic: @mavrothal: It was not working yesterday until now. The error was "Connection Refused". I guess they fixed the git link exactly when you checked :laughing:

mavrothal: For these rare chases may be better to add ```aufs_utils_git_alt="https://git.code.sf.net/p/aufs/aufs-util"``` and then change the code [later](https://github.com/peabee/woof-CE/blob/f630ebd1248bdb7720cda73824f3015f43f484f9/kernel-kit/build.sh#L645) to
```
if [ ! -d aufs-util ] ; then
    git clone ${aufs_utils_git} aufs-util
        if [ $? -ne 0 ] ; then
            git clone ${aufs_utils_git_alt} aufs-util
                 if [ $? -ne 0 ] ; then
			rm -rf aufs-util
			exit_error "Failed to get aufs-util from git, do it manually. Kernel is compiled OK :)"
		fi
	fi
fi
```
The only problem in this case is that you may end up with a different aufs-utils if there is any other reason for git to be inaccessible.

wdlkmpx: mav, i think it's a good idea to apply that patch.

peabee:
What really could do with fixing is to move the test as to whether the aufs-utils url is active to before the kernel is compiled - you currently have to wait for the kernel compile to finish (90mins) before being told that the aufs-utils url is down. Simply having the alternative urls in build.sh but commented out would be sufficient.

wdlkmpx:
I was editing the script to clone the aufs stuff just once and then just synchronize it with the remote repo. But just like many things i've done, it got lost.. I'll make the script more reliable.

PR #975, opened 2017-03-06
glib and zzz_glib conflict

ninaholic:
Things seem to work better (but still a few terminal warnings) without debian version of glib. Thanks to pemasu
http://murga-linux.com/puppy/viewtopic.php?p=946215#946215


wdlkmpx:
I also identified that issue.. i just send everything to NULL instead, so all the debian glib packages appear to be installed. I'm finishing building, then i'll push some fixes

PR #977, opened 2017-03-07
add jwm-xdgmenu for fixmenus

ninaholic:
"xdg_puppy" downloads:
http://distro.ibiblio.org/puppylinux/pet_packages-common/xdg_puppy-0.7.6-15-binaries_no-JWM-p4.pet
"xdg_puppy_jwm" downloads:
http://distro.ibiblio.org/puppylinux/pet_packages-common/xdg_puppy_jwm-0.7.6-16-i486.pet

2nd one has jwm-xdgmenu (needed for fixmenus by default), so I changed to xdg_puppy_jwm.
1st has fvwm-xdgmenu, icewm-xdgmenu and wm-xdgmenu only. Should that be kept too? :D


mavrothal:
Well, slackware builds are using xdg_puppy_jwm while Ubuntu xdg_puppy. Phill however does not incorporate all of @zigbert changes in the desktop/jwm environment, so my inclination would be to pull. However, I would rather wait (a day) to hear from @01micko, @mrfricks (Phil) or Zig first. 
Is this change working for you OK?

mavrothal:
... After further review the call is reversed. 😸

PR #1012, opened 2017-05-15
Prevent multiple network manager dialogs; package SNS

rserwin1:
Terminates old dialogs or avoids starting second copy of same dialog.  Moves SNS into its own package.


rserwin1:
Please ensure that the 5 new files in rootfs-package/simple_network_setup/usr/local/simple_network_setup are given execute permission.
Please add package simple_network_setup to your list of packages available for inclusion in a distro -- however you do that -- or tell me how to do it myself.
Also, can I help resolve the apparent conflict regarding sns?

01micko:
Hi Richard,

I haven't had time to make a full review but I suspect the conflict is only minor (probably from  3a556ad )

As for:

> Please add package simple_network_setup to your list of packages available for inclusion in a distro -- however you do that

Take a look at `3buildistro-Z` .. I think it will become apparent that it will be in there 'by default' but you may feel that some rules should apply. Take a look at the other cases.

On a more separate note, and referring to my recent slacko beta, (http://www.murga-linux.com/puppy/viewtopic.php?t=110510) there is now a problem with auto connecting (on a discovered interface) since @wdlkmpx shifted something (not sure what just yet).

Since SNS is now proposed to go to a package then we need a generic mechanism to connect a wired connection on first boot - which I think is desirable (however could be debatable). Of course this brings `netmon_wce` (the current 'status icon' - which I maintain) into play as well. While connecting to a network if it is a wired connection is relatively easy the problem is the exact place to do it. Sure, this isn't part of the current PR and should be separate, but I thought I would mention this to bring it to your attention.

Cheers!

01micko:
Ok, so the conflict turned out to be a 'false flag'. Can happen with the git web gui.

I'll close this and leave my other comments for consideration.

rserwin1:
I added a commit to merge the ".svg" fix into the sns package but see 
that you already made that change.  Can you delete that commit for me?  
Or should I take some other action to nullify it?

Also, #1012 seems to detect an eth0 connection on a first boot in 
xenialpup, so might fix the slacko problem.  I have posted the upgrade 
packages in the forum (additional software > network > 
network_connect_update), in case you want to try it easily.

BTW, in that situation, it is the network wizard rc.network that starts 
the connection.  So, we need to decide what to do if NetWiz is left out 
of a build -- maybe just not support the automatic default connection, 
without NetWiz.  Or expand /etc/rc.d/rc.network-basic to handle that case.


Mick Amadio wrote:

PR #1053, opened 2017-06-21
Fix #907 for Rox

ninaholic:
I think this works? Thanks to radky
http://www.murga-linux.com/puppy/viewtopic.php?p=958263#958263


ninaholic:
Tagging https://github.com/puppylinux-woof-CE/woof-CE/pull/907 in the title has failed, but I know it is possible... somehow...

PR #1054, opened 2017-06-21
sns: display connections by signal strength, best first

rserwin1:


rserwin1:
Also protects against double quote in SSID or password, when deleting a profile.

PR #1063, opened 2017-07-14
slacko 14.2 14.1 14.0 remove sfs-convert, update

ninaholic:
sfs-converter needed mksquashfs3 and unsquashfs3 to convert
http://www.murga-linux.com/puppy/viewtopic.php?p=960716#960716

Also, the newer initrd_progs did not boot with older default mksquashed files (14.0)


wdlkmpx:
I clicked the wrong button but the changes look good so i'll merge them..

PR #1071, opened 2017-07-29
merge2out: make sure woof-out is owned by root

woodenshoe-wi:
If running woof-CE in a multi-user environment using run_woof
it is possible that the woof-CE repo has been cloned
as an ordinary user.

### DIFF
--- a/merge2out
+++ b/merge2out
@@ -216,6 +216,10 @@ ls rootfs-packages|while read pkg;do
 	echo -e "$pkg=\"true\"" >> support/rootfs-packages.conf-backup
 done)
 
+if [ "`stat -c %U:%G $0`" != 'root:root' ] ; then
+	[ "${WOOF_OUT}" != "/" ] && chown -R root:root ${WOOF_OUT}
+fi
+
 echo -e "\nUndoing VCS-friendly stuff..."
 (
 cd ${WOOF_OUT}


wdlkmpx:
Looks like a heavy operation..

Git doesn't track file ownership, that's why all files are owned by the user cloning the repo.

It also means 'merge2out' itself is owned by the wrong user... so this might be a better solution:

	if [ "`ls -l $0 | grep root`" = "" ] ; then
		chown -R root:root .
	fi

woodenshoe-wi:
I don't want to change the ownership of the git repo, just the woof-out_* directory that merge2out is creating before any ownerships are set later in merge2out.

Would something like this be better?
```
if [ "`stat -c %U:%G .`" != 'root:root' ]; then
  chown -R root:root ${WOOF_OUT}
fi
```
I would need to test it before I would make it into a commit.

wdlkmpx:
woof is designed to work properly only with root. making root the
owner of the woofce dir is a one time operation. once that happens you
need to be root to update the local git repo, etc... thus making
things smooth.

i removed a lot of entries from VCSMETADATA because most of them were
with removed files or files that are already owned by root with the
proper file permissions.. only files owned by spot or that require
special permissions should be kept there.

you might want to test something like this and then push it to testing...

	if [ "`whoami`" != "root" ] ; then
		echo "Must be root"
		exit 1
	fi

	if [ "`stat -c %U:%G $0`" != 'root:root' ] ; then
		chown -R root:root $(dirname "$0")
	fi

woodenshoe-wi:
Since I have been receiving e-mails from puppylinux-woof-CE there have already been two people that ran into trouble trying to use woof-CE from Ubuntu.

Regular linux users might not be comfortable having to make all changes and updates to woof-CE as root.  If people have git configured to their liking as their normal user, as soon as they change to root user all the configuration would have to be redone.

I think run_woof can work to allow people to do woof-CE builds from whatever version of linux that they want, and merge2out is the perfect place to switch from a standard multiuser linux environment to a Puppy everything as root environment in the woof-out_* directory.

Of course merge2out needs to be run as root, but I think it still works fine when it is not owned by root, and all the other scripts are run from the woof-out_* directory, not woof-CE.

wdlkmpx:
I agree.

woodenshoe-wi:
Ok, I pushed an amended commit.

This should only do the chown -R if needed.  I tested it with my woof-CE repo which is not owned by root, and woof-out is owned by root as expected.

PR #1076, opened 2017-08-30
Record packages downloaded

peabee:
Ready for a new multiple-package 2createpackages tool.

### DIFF
--- a/woof-code/1download
+++ b/woof-code/1download
@@ -153,6 +153,7 @@ else
 	XFILE=status/findpkgs_FINAL_PKGS-${DISTRO_BINARY_COMPAT}-${DISTRO_COMPAT_VERSION}
 fi
 
+echo "" > $MWD/status/download_list
 LINECNT=1
 cat $XFILE | while read ONEDBENTRY
 do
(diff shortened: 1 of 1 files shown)


wiake:
I was wondering if option CHANGED_ONLY for 2createpackages doesn't handle this need proposed by peabee anyway? i.e. only packages that have been changed get rebuilt. Both 1download and 2createpackages seem to use run_findpkgs which determines what has been changed via md5sums according to my reading? I may be wrong and haven't had time to test CHANGED_ONLY but plan to tomorrow.

peabee:
I prefer to have a list that I can review and that can be used to explicitly do 2createpackages - also don't know where the md5sums for each package are stored??

wdlkmpx:
Currently 1download does not store md5sums. pet packages already have a md5sum at the end of the file. Currently 1download only checks md5sums in pet pkgs, verifies pkg integrity for other formats and that's it. 1download should store the md5sum after a pkg is successfully download and verified.. and
an extra md5sum check can be added and triggered with a param like "./1download check" or something..

PR #1085, opened 2017-09-11
Create 2create_download_list_pkgs

peabee:


wdlkmpx:
This looks like something that can go in 2createpackages.. with a
param like downloaded or downloaded_only. I can merge the pr as is or
i can integrate the script into 2createpackages. WhatsApp do you think

peabee:
I'm happy for it to go into 2createpackages as an option - but 2createpackages needs some --help to explain what the options do??

peabee:
Wow - 2createpackages makes my brain hurt! ;-)
However there is a loop at lines 935-940 that could do my selective build....
"Just" need to get the contents of /status/download_list into /tmp/2createpackages-redo in the correct format and then make sure that nothing else is run if a parameter of downloaded is given....

PR #1093, opened 2017-10-15
Update Original Pup

peabee:
To ensure switching from and to other themes is consistent


zigbert:
No big deal, but the aero cursor theme has never been used in the old Puppies. Keeping this empty will use the already defined cursor theme.

What we need is an option to revert to the original X theme.
I'll look at that...

PR #1096, opened 2017-10-31
bring back the woofwoof,

mrfricks:
it's part of puppy's identity! plus the new sound really is rubbish!!!
shame on you wdlkmpx :p


wdlkmpx:
image

There were better sounds than that barking piano, but that one happened to be small enough. How an annoying sound can be so important?

The distro should be renamed to Easy OS or something, or yet another fork should fix this, then we won't have to worry about really childish stuff. Funny one would think that most p***y users are children, but according to what i read in the p****y forum, most are grandpas.. that should know better.

Do you realize how awful a bark is.. Think of death metal and hear the growls. I downloaded a technical death metal album expecting the best, but damn, the constant and rabid growls and barkz didn't let me hear the instruments. Even a cookie-monster voice is not as bad as rabid growls and barks from the underworld, in fact it even sounds melodic such as in 'Prelude to a suicide' https://m.youtube.com/watch?v=_qJFWKEn99w

In contrast, 'Angel' [ https://m.youtube.com/watch?v=aAkH2RP-FLM ] is
a good example of how to sing with power without trying to sound like a wild animal..

PR #1098, opened 2017-11-06
Update DISTRO_SPECS

peabee:
Remove old BUILD_FROM_WOOF= as ./merge2out does not update it

### DIFF
--- a/woof-distro/x86/ubuntu/xenial/DISTRO_SPECS
+++ b/woof-distro/x86/ubuntu/xenial/DISTRO_SPECS
@@ -10,7 +10,6 @@ DISTRO_FILE_PREFIX='xenial'
 DISTRO_COMPAT_VERSION='xenial'
 #read by /usr/bin/xwin to bypass Xorg Wizard at first boot:
 DISTRO_XORG_AUTO='yes'
-BUILD_FROM_WOOF='testing;61d7002;2015-11-03 06:44:22 +1000'
 DISTRO_KERNEL_PET='Huge_Kernel'
 DISTRO_DB_SUBNAME='xenial'
 DISTRO_ARCHDIR='i386-linux-gnu'


mavrothal:
The BUILD_FROM_WOOF info is automatically edited by merge2out (#L286-291). 
Unless a specific puppy is  edited manually outside the woof-CE git the actual entries in the DISTRO_SPECS files in woof do not matter. 
In puppies edited outside woof, no mater what the "fixed" entry is, it  will be misleading in the sense that the build  is  not reproducible from that git version. 
So better be an ancient one, or maybe remove that entry totally in these cases.

peabee:
Maybe #L289 that deletes any BUILD_FROM existing should be outside the if clause - I believe this is the only aspect of the build process that requires git and therefore devx - so if devx is not loaded the old BUILD_FROM line does not get deleted and no error is thrown so the builder is not alerted there is a "problem" as the build completes successfully.

ninaholic:
I saw that xenialpup-7.0.8.6.iso showed up with "testing;61d7002" in /etc/DISTRO_SPECS too and when I tried to find that commit in woof-CE it was from pre-Master and even pre-Legacy:

https://github.com/puppylinux-woof-CE/woof-CE/commit/61d7002 https://github.com/puppylinux-woof-CE/woof-CE/tree/61d7002

This was quite confusing. Is newest xenialpup really made from woof-CE from November 2015? I think this BUILD_FROM_WOOF line should be removed too so it can be replaced with real commit.

mavrothal:
The BUILD_FROM line only requires git which presumably you have to clone the woof git. 
I think in the Xenial case Phil may never cloned woof but just downloaded the tarbal. Thus the failure. 
Instead of taking it out maybe add and else in the clause 
```
else
      echo "BUILD_FROM_WOOF=No woof-CE-git info available" >> ${WOOF_OUT}/DISTRO_SPECS
```
or something

mavrothal:
I committed d5db90b
change/edit as needed

PR #1100, opened 2017-11-08
slacko 14.1: fix/revert dbus_slack

ninaholic:
The "machine-id" part of this patch has never worked for me in 14.1: https://github.com/puppylinux-woof-CE/woof-CE/commit/d23a3ed987320f8bde804c47530e1d2e2b6a1270

https://ibb.co/d0W3Lw

Ignoring the code for 14.1 with if statement (behavior before patch) is one way that seems to fix.

### DIFF
--- a/woof-code/packages-templates/dbus_slack/FIXUPHACK
+++ b/woof-code/packages-templates/dbus_slack/FIXUPHACK
@@ -4,6 +4,9 @@
 mkdir -p etc/init.d
 mv etc/rc.d/* etc/init.d/
 find etc/init.d -type f -name '*.new' -delete # may or may not be there
-mkdir -p var/lib/dbus/ 
-touch var/lib/dbus/machine-id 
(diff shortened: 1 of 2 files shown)


mavrothal:
It would be a good idea not to put 2 unrelated commits in the same pull. 
If in doubt for one will stall the other too.

ninaholic:
Thanks @mavrothal I will try to separate them now :)

ninaholic:
The strangest part is that 6.3.2 was released after this patch (I think) yet it somehow works... will have to investigate why as I always got dbus error in 14.1 with this new template :thinking:

ninaholic:
Another way to fix seems to be to run **dbus-uuidgen > /etc/machine-id** at startup after, but I think this "blank uuid" should be removed in 14.1 anyway because of fatal bugs in screenshot :thinking: 

screenshot

PR #1111, opened 2017-12-06
Kernel kit changes

woodenshoe-wi:
Over the summer I got kernel-kit to cross compile kernels with AUFS for Raspberry Pi.

I finally made the changes into proper commits.


wdlkmpx:
I read the whole diffs.. and the changes look good. But i can't test and look out for bugs atm.. but i think i'll merge them anyway

woodenshoe-wi:
There is no hurry, so take as long as you want.

PR #1112, opened 2017-12-06
Zip file frugal install for Raspberry Pi

woodenshoe-wi:


initrd-progs/0initrd/init
```diff
@@ -1317,7 +1317,11 @@ if [ "$PUPSAVE" ];then #refine pupmode
  #   we must check if it's located in a flash storage device
  #   in order to change the $SAVE_LAYER to /pup_ro1
  PUPSAVE_DRV=$(echo $PUPSAVE | cut -f1 -d ',')
- PUPSAVE_DRV=${PUPSAVE_DRV:0:3} #sda1 -> sda TODO: needs a proper fix
+ if [ "${PUPSAVE_DRV:0:3}" = "mmc" ];then
```
mavrothal: Any chance that this might affect a laptop/chromebook booting from mmc?

woodenshoe-wi: Absolutely.

The problem was that when the Raspberry Pi boots from mmcblk0p1 PUPSAVE_DRV gets set to 'mmc' which doesn't exist, and PUPSAVE_MEDIA would not be set to usbflash.

If laptops/chromebooks boot from mmcblk0p1 they would have been experiencing the same problem, if they boot from a device literally called 'mmc' then I don't know what the effects are, but I can't test it because I don't have an affected laptop/chromebook.

wdlkmpx: That is actually a bugfix. I had identified that issue but never bothered to fix it myself, like some other bugs i know about

gyrog: Perhaps a more generalised solution might be code like the following:
if [ ! -d "/sys/block/$PUPSAVE_DRV" ];then #not device e.g. sr0, assume partition
 PUPSAVE_DRV="${PUPSAVE_DRV%[0-9]}"; PUPSAVE_DRV="${PUPSAVE_DRV%[0-9]}"; PUPSAVE_DRV="${PUPSAVE_DRV%[0-9]}" #remove upto 3 trailing digits
 [ "${PUPSAVE_DRV:0:3}" = "mmc" ] && PUPSAVE_DRV="${PUPSAVE_DRV%p}" #remove any trailing 'p'
fi

woodenshoe-wi: If someone wants to fix the init script in a different way than 43c95aa let me know when it's fixed and I will cherry pick the other commits and reissue the pull request without 43c95aa

Until this is fixed I would not be comfortable merging the zip file builds anyway.

gyrog: I suggest that you don't wait.
Rather change the line:
if [ "${PUPSAVE_DRV:0:3}" = "mmc" ];then
to:
if [ "${PUPSAVE_DRV:0:6}" = "mmcblk" ];then
Then there is less chance of clashing with some other completely different device that happens to start with "mmc".
Since the way partitions are named for "mmc" devices is different from "sd" devices, there may be no simple clean code that works for all devices.
Despite it's limitations, the current code has been working reliably for most folk.
I think your patch wouild work just as reliably in the real world, unless you can imagine there being a machine with a "mmcblk10" device (i.e. at least 11 "mmc" devices).

woodenshoe-wi: I am not in any hurry, and I discovered that I have to add gdb-python to the devx or the debugger won't work, so I will try another build with your suggested solution and if I don't have any typos I will reopen the pull request then.

PR #1114, opened 2017-12-09
slacko 14.0 - _00build.conf with extra commands

ninaholic:
I notice this option when looking at mrfricks xenialpup commits, seems like a great idea, which makes easier to fix some things. I started with these:

1. Some programs/libraries have dependencies that are not met, sometimes because lib is wrong version, the biggest offender probably being the 1.2MB x264 binary: http://murga-linux.com/puppy/viewtopic.php?p=976194#976194

2. Pmusic is already part of OpenWith from the PMUSIC_TRAY pet, this deletes duplicate OpenWith entry, and adds one for gnome-mplayer

3. Some programs (xarchive) look for 7za instead of 7z, and will not open 7z archive without finding it. Also added an icon for 7z files so it doesn't show up as unknown

4. the e3 binary can't be stripped or you get "segmentation error", this was fixed then removed from woof-CE so I bring it back for 14.0 specifically

Might add more later or try to fix some things in Devuan next. Tested a build with this and the EXTRA_COMMANDS did work as expected :smile:

### DIFF
--- a/woof-distro/x86/slackware/14.0/_00build.conf
+++ b/woof-distro/x86/slackware/14.0/_00build.conf
@@ -1,0 +1,244 @@
+#
+#  persistent configuration options
+#
+#  see also DISTRO_SPECS DISTRO_PET_REPOS DISTRO_COMPAT_REPOS-*
+#
+#  **NOTE**: check the original file every once in a while
+#            settings might be added or removed...
+#
+
+#----------------
+#support/findpkgs (called by most scripts)
+#----------------
+## set to 'yes' to speed up process..
+USE_FINDPKGS_SEARCH_HELPER=yes
+USE_FINDPKGS_DEP_HELPER=yes
+
+#----------------
+# 2createpackages
+#----------------
+# binaries are usually already stripped. set to 'no' to speed up process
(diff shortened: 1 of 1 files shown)


mavrothal:
I think this may not be the best approach because then someone has to maintain 3-13 _00build.conf files or they'll stay behind. Might be better to use a patch file for the specific puppy version instead of the full _00build.conf and add the patching command merge2out.
Let's think of a it for a couple of days and if nobody else pulls/closes this or comment, I'll pull it

wdlkmpx:
Slacko 14.0 is not supported anyway so anything that goes in there doesnt really matter. I think i deleted it twice before bringing it back again. It's a good idea to open a thread about the new slacko 14.0 for retro machines, with xorg XAA  accelaration. With dpup having precise's as an alt xorg server by default i couldnt care less about the rest.

Regarding devuan, im not sure youd want to maintain that thing, but it can also be possible. Then there is dpup which i'll keep updating to come with lxde as the only possible option with my favorite apps wiping most puppy stuff, but i'll do it in a w subdir or something, then someone could invite radky to place his stuff in the default dpup build.

PR #1125, opened 2018-01-07
stretch and puduan: fix some menu entries

ninaholic:
to remove % and bad xcalc


wdlkmpx:
jwm should ignore %f %F %u %U, it's really simple.

that way there's no need to alter desktop files. pcmanfm recognizes all those %* and can work without them. it's getting harder and harder to keep the fondness for jwm.

key scripts are 2createpackages and installpkg.sh

PR #1127, opened 2018-01-09
Update probepart

peabee:
See: http://www.murga-linux.com/puppy/viewtopic.php?p=979709#979709
corrects problem with partview after creating an encrypted savefile

### DIFF
--- a/woof-code/rootfs-skeleton/sbin/probepart
+++ b/woof-code/rootfs-skeleton/sbin/probepart
@@ -91,7 +91,7 @@ else
 	fi
 	#devices that have partitions... 
 	#130127 screen out sr0-sr9. early kernels do not have these in /proc/partitions, 3.2+ do, which causes desktop icon to not appear when audio-cd inserted.
-	PARTITIONS="`cat /proc/partitions | tr -s ' ' | cut -f 4-5 -d ' ' | grep -vE ' loop| ram| zram| sr'`" #each line ex: 16076800 sda5
+	PARTITIONS="`cat /proc/partitions | tr -s ' ' | cut -f 4-5 -d ' ' | grep -vE ' loop| ram| zram| sr| dm-'`" #each line ex: 16076800 sda5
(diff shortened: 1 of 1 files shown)


mavrothal:
Come to think of it I believe this is the wrong way around it.
We do not want to hide luks encrypted volumes. We have to teach apps how to handle them.
Not only partview but also pmount and pup_event to show them on the desktop and (un)mount as needed.
Here is a fast patch for partview.
Analogous thinks can be done for pMount/pup_event and friends
```
diff --git a/woof-code/rootfs-skeleton/usr/sbin/partview b/woof-code/rootfs-skeleton/usr/sbin/partview
index cb531b7..0728c29 100755
--- a/woof-code/rootfs-skeleton/usr/sbin/partview
+++ b/woof-code/rootfs-skeleton/usr/sbin/partview
@@ -33,10 +33,12 @@ do
    if probedisk $ATAG | grep -q '|optical' ; then
      ASIZE=4300000
      [ "`dvd+rw-mediainfo /dev/${ATAG} 2>&1 | grep 'non-DVD'`" != "" ] && ASIZE=700000
-   fi
+   fi ;;
+   dm-*) ATAG="$(df |grep \/luks| cut -f1 -d ' '| cut -f 3-5 -d /)" ;;
  esac
  AFPATTERN="^/dev/$ATAG "
  AUSED=`df -k | grep "$AFPATTERN" | head -n 1 | tr -s " " | cut -f 3 -d " "`
+ [ "$AUSED" ] && [ "$(echo -n "$AFPATTERN" | grep \/luks)" != "" ]  && ATAG="luks-fs"
  if [ ! "$AUSED" ];then
   MNTSTATUS='(not mounted)'
   mkdir -p /mnt/$ATAG
```

wdlkmpx:
With this very probepart script, I saw a lot of dm- devices on android.

I think the p-v-m misbehaves when it has to deal with luks-encrypted partitions, but the rox desktop works ok. It's a while since i had a savefile to test with.

PR #1131, opened 2018-01-22
slacko 14.0: optimize

ninaholic:
A few larger things that were removed in slacko 5.7 package-templates to save size


ninaholic:
actually all these together only saved about 625K as super xz compression sfs, probably not worth it. Someone can reopen this if they want  :laughing:

PR #1132, opened 2018-01-23
merge2out-gui

woodenshoe-wi:
Completed a modified xenial64 build and ran merge2out/0setup several times without noticing any problems.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+woof.conf
--- a/merge2out
+++ b/merge2out
@@ -301,7 +301,11 @@ sed -i '/BUILD_FROM_WOOF=/d' ${WOOF_OUT}/DISTRO_SPECS
 if which git &>/dev/null && [ "`git log -n 1`" != "" ]; then
 	GITBRANCH=$(git rev-parse --abbrev-ref HEAD)
 	GITHEAD=$(git log -n 1 --pretty=format:"%h;%ci") # git rev-parse HEAD
-	echo "BUILD_FROM_WOOF='${GITBRANCH};${GITHEAD}'" >> ${WOOF_OUT}/DISTRO_SPECS
+	if [ "$ALTERNATE_BUILD" != "" -o "$EXTRA_MERGE" != "" ];then
+		echo "BUILD_FROM_WOOF='${GITBRANCH};${GITHEAD};ALTERNATE_BUILD=${ALTERNATE_BUILD##*/};EXTRA_MERGE=${EXTRA_MERGE##*/}'" >> ${WOOF_OUT}/DISTRO_SPECS
+	else
+		echo "BUILD_FROM_WOOF='${GITBRANCH};${GITHEAD}'" >> ${WOOF_OUT}/DISTRO_SPECS
+	fi
 else
 	echo "BUILD_FROM_WOOF=\"No woof-CE-git info available\"" >> ${WOOF_OUT}/DISTRO_SPECS
 fi
--- a/merge2out
+++ b/merge2out
@@ -228,6 +228,95 @@ if [ -d "woof-distro/${TARGETARCH}/${COMPATDISTRO}/${COMPATVERSION}/patches" ];t
 		patch --unified --forward --silent -p 3 --input="../woof-CE/$PATCHFILE" --directory=${WOOF_OUT}
 	done
 fi
+if [ "$EXTRA_MERGE" != "" ];then
+	if [ -d "${EXTRA_MERGE}/packages-templates" ];then
+		echo "Copying ${EXTRA_MERGE}/packages-templates..."
+		for ONEDIRECTORY in \
+		`find ${EXTRA_MERGE}/packages-templates -mindepth 1 -maxdepth 1 -type d | tr '\n' ' '`
+		do
+			TEMPLATEDIR=`expr "$ONEDIRECTORY" : "${EXTRA_MERGE}/packages-templates/\(.*\)"`
+			echo "    $TEMPLATEDIR"
+			if [ -n "$TEMPLATEDIR" -a -d "${WOOF_OUT}/packages-templates/${TEMPLATEDIR}" ];then
+				rm -r ${WOOF_OUT}/packages-templates/${TEMPLATEDIR}
+			fi
+			cp -a ${ONEDIRECTORY} ${WOOF_OUT}/packages-templates/
+		done
+	fi
+	if [ -d "${EXTRA_MERGE}/rootfs-packages" ];then
+		echo "Copying ${EXTRA_MERGE}/rootfs-packages..."
+		for ONEDIRECTORY in \
+		`find ${EXTRA_MERGE}/rootfs-packages -mindepth 1 -maxdepth 1 -type d | tr '\n' ' '`
+		do
+			PACKAGEDIR=`expr "$ONEDIRECTORY" : "${EXTRA_MERGE}/rootfs-packages/\(.*\)"`
+			echo "    $PACKAGEDIR"
+			if [ -n "$PACKAGEDIR" -a -d "${WOOF_OUT}/rootfs-packages/${PACKAGEDIR}" ];then
+				rm -r ${WOOF_OUT}/rootfs-packages/${PACKAGEDIR}
+			fi
+			cp -a ${ONEDIRECTORY} ${WOOF_OUT}/rootfs-packages/
+		done
+	fi
+	if [ -d "${EXTRA_MERGE}/rootfs-skeleton" ];then
+		echo "Copying ${EXTRA_MERGE}/rootfs-skeleton..."
+		ROOTFS_SKELETONFILES=`find ${EXTRA_MERGE}/rootfs-skeleton/* \( -type f -or -type d -or -type l \) -printf '    %f\n'`
+		if [ -n "$ROOTFS_SKELETONFILES" ];then
+			echo "$ROOTFS_SKELETONFILES"
+			cp -a -f --remove-destination ${EXTRA_MERGE}/rootfs-skeleton/* ${WOOF_OUT}/rootfs-skeleton/
+		fi
+	fi
+	if [ -d "${EXTRA_MERGE}/patches" ];then
+		echo "Applying patches from ${EXTRA_MERGE}/patches..."
+		for PATCHFILE in \
+		`find ${EXTRA_MERGE}/patches -mindepth 1 -maxdepth 1 -type f -iname '*.patch' | tr '\n' ' '`
+		do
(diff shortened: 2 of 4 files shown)


mavrothal:
Did not really looked carefully through it, but may not be a good idea if there is not any record of what is actually merged and where, whatever is merged, is coming from.
It may simply result in non-reproducible builds.

woodenshoe-wi:
merge2out will output the names of the merged files to the terminal, if it would log things like kernel-kit that would help with keeping track of what was done.

As for reproducible builds, there is currently nothing that stops people from making their own changes to woof-CE and not sharing them.  The GPL2 only requires the "source code" to be redistributed if the "binaries" are redistributed.  As far as I know, this is the closest thing to leverage that you can use to force people to merge their changes into woof-CE.

I was working on making a build of xenial64 with valgrind, but I had to modify some of the scripts and debdb2pupdb.  If you want the modifications it is easier for me to tar up a dictionary and put it in Dropbox than to make commits of quick hacks that shouldn't be directly merged into woof-CE anyway.

mavrothal:
Obviously anyone can make any changes and not share or even locally track. 
The question is if this is a good idea and should it be facilitated by the puppy build system. 
Obviously I have no issue with adding some EXTRAS folder, I just do not think that the additions should be unknown to the user and soon forgotten by the builder.

woodenshoe-wi:
If you don't want this change merged then close the pull request.

I already have the changes in a tar file in Dropbox and if someone else wants to use them they can add them to their own woof-CE directory.

I will continue to use the changes myself.  I had been keeping changes in my woof-CE directory without making them into commits, but running `git status` would give me a wall of modified and untracked files.

You did not say if you wanted my modifications to add valgrind to the devx.  I have no plans to release the iso, it was only for my own use.  Since valgrind is only useful for someone writing or debugging code it will not be generally useful to average Puppy users and I have no plans to try to merge the changes into woof-CE because I already have my working build.

mavrothal:
I would rather had the pull amended/resubmitted with code that will pass the EXTRAS info (including version of binaries and of possible patch/config options or something like a petbuild) into the finished build (somewhere in $HOME/.packages). If it is pointing to a public source for those, even better!

Regarding valgrind is a fairly big package and I think is not even a handful of people reporting using it in puppy but I do not have strong feelings. Is an excellent debugging tool. Besides the puppy builder can include anything they see fit in their offerings.

gyrog:
Regarding valgrind, large, limited number of users, surely it could be a separate sfs file?
Even if it needed to be loaded before the devx.

Regarding EXTRAS:
If you push mods to  bits and peices from time to time, it's important to keep you local clone "clean".
If you are also working on a significant project that you are not yet ready to commit to woof-ce, then the EXTRAS feature allows you to use your local clone of woof-ce for both.
It's the sort of thing that in a few situations would be just the thing, if it's there.
Even if it is not frequently used, someone (other than woddenshoe-wi) might find it useful sometimes.

woodenshoe-wi:
I'm not suggesting valgrind be added to builds by default, but just adding it to my own build required me to recompile debdb2pupdb because valgrind depends on the debugging info for libc6 and debdb2pupdb skips all debugging packages.  Then I had to fix 2createpackages so it would know what to do with the debugging info.

If you want to install valgrind into an existing Puppy, you have to download the debugging info for libc6 and if it is not an exact match to your current libc6 you will also have to download and install the matching libc6.  Add to that Puppy doesn't keep the libraries in the same place as the compat distro so the debugging info has to be moved to match.

I think I found a solution to make sure the builds are reproducible, just make a tar file of the ALTERNATE_BUILD and EXTRA_MERGE directories in /root/.packages

The ALTERNATE_BUILD and EXTRA_MERGE directories can even be git repos themselves, with readme files that can have a link to the online location.

mavrothal:
As gyro suggested and sfs with valgrind and matching libc6 maybe the way to go on that.

Regarding the info on the EXTRAS a tarball maybe an unnecessary duplication and is not going to add much more than a file list. 
If the extra files are scripts/text/xml etc, can be easily found in the build. But if they are binaries, without the source or a build-script/config are as good as their listing. So a ```find /$EXTRAS/ > sandbox3/rootfs-complete/root/.packages/EXTRAS_file_list``` should be OK to identify (and recover as needed) the added files from the build but still leaves a black box in the case of binaries.
If there is a private git repo that also contains build scripts etc then a tarball would make sense.

I should just mention that I have rediscovered the wheel several times (and witnessed it many more times)  just because or limited recording. It looks unnecessary when you are working on something and you know exactly what is going on and where things come from and go to at the time but a couple of months latter you may have no idea what the project was about... 😸

woodenshoe-wi:
Maybe an example will help,  
https://github.com/woodenshoe-wi/xenial64-valgrind

The only binary is the modified debdb2petdb2 (I did include the source code, modified Make file and petbuild)

The modified scripts are part of woof-CE and won't be in the final build, so a file list won't work.

If the EXTRAS tarball is included in /root/.packages if someone wants to duplicate a build they can clone woof-CE, copy the EXTRAS tarball to the parent directory of woof-CE, unpack it and add it as an EXTRAS during merge2out and as long as nothing in EXTRAS breaks the build they have a brand new build with all the latest changes from woof-CE.

woodenshoe-wi:
Good point about BUILD_FROM_WOOF

Another thing I want to point out is that this doesn't _need_ to be merged into woof-CE.

If it is not wanted in woof-CE I can modify the scripts slightly and it can be used as an outside add on.

The only disadvantage of keeping it as an outside add on is making sure it stays in sync with the merge2out in woof-CE.

mavrothal:
Is there any additional reason that this is merged?
It would appear that was in discussion with additional changes pending.
If deemed fine as is, some justification should be added as to why we "officially" facilitate addition of untracable/hidden build options and elements in puppies. 
Please provide some reasoning for this policy change **before** you pull.

wdlkmpx:
The reason is simple, it's the same reason micko added a bunch of
custom core scripts to the slacko 13.13  build as pet pkgs

You can always revert everything i push or change, its really up to
you, just like when you brought back those annoying barks that i hate
with passion. i just  think of the work and time wasted by the
contributors to make the project more friendly

mavrothal:
I do not really care much to revert the change before complains for unreproducible puppies appear and the "explanation" is the EXTRAS folder (or even then...). So I will not revert as I didn't pull but did not close it either.  Someone's works on something per se can not be enough reason to pull it. Pull means you reviewed and find it valid/useful/interesting/"something". 
So the question remains if there is a policy change in facilitating untracable build options in puppy? 

BTW z_hacks pets (like Micko's)  is the "usual" final customization of puppies and so far are publicly available. EXTRAS by design, is not. So the parallel does not really address the issue if we want to facilitate untracable elements in puppies. 
Do we?

Just to repeat that I think the EXTRAS is a fine idea, but the current implementation introduces (more) black boxes in the released puppies. Thus my concern.

wdlkmpx:
Reading the commits i see it doesn't introduce a black box, it
properly logs all the extra files.

It's useful indeed, let's think about this: someone finally produces
official reproducible dpup builds with their own stuff now officially
available by default... would it be somethind i'd use? Probably, but
most likely i won't.

Should i be willing to build a default tahrpup, xenialpup or slacko..
definitely not, that's why i never build one of those puppies and
pontetially identify some important issues.

Meanwhile I'll edit the script to make the extra stuff hidden by
default and post here the cli arg env var to trigger it

wdlkmpx:
Ok to enable the extra_merge alt build stuff, export ENABLE_EXTRA_MERGE=1

Very few people will be aware of this, those who are watching the repo
an receiving emails (i am not) and those who are already hacking
woofce. But to be able to use this stuff, you already have to have a
knowledge that is beyond most puppy users can understand in their
whole lives and beyond.. because you must understand how woofce works
and the directory tree.

woodenshoe-wi:
> Is there any additional reason that this is merged?  

Because I told you to close the pull request if you didn't want it to be merged, but you didn't.  

> It would appear that was in discussion with additional changes pending.  

What changes?  I already added d9d8484, 7921154 and e4518e3.  

> some justification should be added as to why we "officially" facilitate addition of untracable/hidden build options and elements in puppies.  

7921154 was supposed to have fixed that.  At least it is better than making changes directly in woof-out which aren't documented and are easy to accidentally overwrite with merge2out.

mavrothal:
> it properly logs all the extra files.

I may have missed that but I can not find where in the build puppy (not the local log) this is happening

woodenshoe-wi:
The way I was envisioning it the ALTERNATE_BUILD functionality was to be equivalent to the builds in woof-distro, and should not be used to add binaries to the builds unless those binaries are part of woof-CE.

The EXTRA_MERGE functionality was to make changes to the woof-CE build process that you wanted to be able to use with any of the builds in woof-distro, also not intended to be used to add non woof-CE binaries.

Although it looks like duplication to you, if the ALTERNATE_BUILD or EXTRA_MERGE directories contain modified versions of woof-CE build scripts, these scripts would not be included in the final build without the tarball.

Hopefully adding a tarball of all the changes to the final build will discourage people from adding anything too large to the EXTRA_MERGE directories.

wdlkmpx:
I have a usecase for this. To add my stuff and commands to complete
the build and boot to a system in which i do things at the speed of
light.

So i guess this does qualify as the tool..

I don't know but maybe copying merge2out.log to
rootfs-complete/root/.packages should be enough? (in line 918)

mavrothal:
The entire merge2out.log maybe an overkill and even confusing to the user (though potentially useful to aspiring builders).
As mentioned earlier if sources or compile options/scripts are not included, a file list is as good as the tarball.  So maybe change [L241](https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/merge2out#L241) of merge2out to ```find . > ${WOOF_OUT}/EXTRAS_file_list```  and  add this latter into rootfs-complete/root/.packages/woof-build.conf in 3builddistro-Z

woodenshoe-wi:
Hmm, merge2out.log is created in ${WOOF_OUT} and stays there.  It is meant for builders to be able to see what happened when merge2out was run after the terminal window it was run in has been closed.  I don't understand why it would be confusing.

I know you keep pushing for a file list instead of a tarball, and I keep saying that the tarball is needed for any changes to woof-CE scripts that don't end up in the final build.

If you want to add binary packages to the build, shouldn't they be added as pets and included in the proper pet package list?

Adding pets can't change the woof-CE build scripts, that was why I wanted the ALTERNATE_BUILD/EXTRA_MERGE features.  If improvements are needed in the pet handling parts of woof-CE to facilitate the addition of locally built pets I would be open to suggestions, but I don't understand why that should be related to ALTERNATE_BUILD/EXTRA_MERGE.

PR #1136, opened 2018-02-05
Fix for mozilla browsers run-as-spot under LXDE

peabee:
See:
http://murga-linux.com/puppy/viewtopic.php?p=982274#982274

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/run-as-spot
+++ b/woof-code/rootfs-skeleton/usr/sbin/run-as-spot
@@ -19,7 +19,7 @@ if [ $(id -u) -eq 0 ]; then
  export XDG_CACHE_HOME=/root/spot/.cache
  export XDG_DATA_HOME=/root/spot/.local/share
  #exec su spot -s /bin/sh -c "cd \"$CWD\"; \"$APP\" $ARGS"
- exec su spot -s /bin/sh -c "\"$APP\" $ARGS"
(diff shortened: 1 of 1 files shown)


wdlkmpx:
just by reading a few lines i knew it had to with the way you start dbus with the lxde stuff in .xinitrc. i have a partial lxde, which means no lxsession. and an untouched .xinitrc. and i guess i don't have that issue.

i have a complete DE with the woofce core stuff +  pup-volume-monitor, pcmanfm (desktop app+filemanager), openbox (wm), lxpanel(menu, tray), my engrampa gtk2 fork (winrar/winzip) and gmrun/lxrandr/gpicview/lxtask/lxappearence/obconf/lxinput. nothing else...  it has a wXP feel and it just feels right.

the latest version of all those apps compile even in slacko 13,37, which is older than ubuntu precise. lxde consists mostly of generic apps that can work independently.

lxde is too underrated... the perfect base for the OS and a project we all should contribute to.

hmm also reading a couple topics in the bugs section, i have a fix or hack for:
http://murga-linux.com/puppy/viewtopic.php?t=112670

this is something anybody can apply, but nobody cares
http://murga-linux.com/puppy/viewtopic.php?t=112368

this looks like something easy to fix
http://murga-linux.com/puppy/viewtopic.php?t=112108

will push some fixes sometime in the future.

funny how i must be the one to do it, because nobody cares.. neither do i haha..

peabee:
Thanks.
FYI....
.xinitrc is unchanged from woof-ce.
startlxde is where dbus is started.

PR #1141, opened 2018-02-15
Improve checkdeps for 64bit

peabee:


wdlkmpx:
You should never use your 'testing' branch to issue a pull request, as there are different ways to merge your commit or in case it's rejected (which i was going to do now), and that's when you'll have trouble syncing your fork with "upstream". Unless you know how to deal with such situations..

PR #1155, opened 2018-03-13
Stop Frisbee entries being corrupted

peabee:
by errors


mavrothal:
I wonder if the other two instances of frisbee/entry redirections on WEP type (open/shared) and the set AP scan mode are also affected by GLib errors.
@rserwin1  ?

rserwin1:
I assume you refer to the --radiolist  Xdialogs.  Yes, they should be changed, too.  Also, the version number in /usr/local/bin/frisbee should be bumped to 1.4.8.

mavrothal:
@rserwin1 
You may want to issue a pull request with these (or additional) changes. 😉

PR #1158, opened 2018-03-22
SNS rc.network: detect modules with hyphens in name

rserwin1:


wdlkmpx:
I'm closing this. The reason i mentioned you is to let you know that it's already fixed with a link to the file with the latest changes.. it was not an issue but a commit. I sometimes do that, since i don't find other way to communicate with others. There's also my forum.. http://wmsgs.freeforums.net/

That is how git works, people can edit stuff, i can also edit your frisbee stuff and the next time you open a pull request there will be a merge conflict just like it's happening now with SNS. But i won't do that of course..

Issue #1133, opened 2018-01-23
Common repos for compiled apps with 32 and 64 bit variants?


peabee:
The repo structure on Ibiblio does not currently (Jan 2018) cater for common compiled apps with both 32 and 64 bit variants.
Such apps have to be placed into the compat-distro specific repos (as they could have lib dependencies which are specific to the compat-distro structure - particularly for 64-bit apps) and therefore can become "out-of-step" over time (yad has been recently cited as an example).
There are a very few apps already in both packages-common and packages-noarch with bit-length parts to their names....even though noarch is meant to be bit-length agnostic (isn't it?).
This issue is raised to stimulate debate as to whether anything should be done?

gyrog:
The elephant in the room is that we don't have a proper repo for Puppy pacakages, the repo on Ibiblio is not really up to the task.
It should be something on git-hub and updateable by at lest the folk that have direct access to woof-ce.
I wish I could say, "here's my sugestion on how to set this up", unfortunately I can't because I have no idea.
And my "todo" list is still quite long.

woodenshoe-wi:
Trying to keep a repo in git will not allow old versions to be deleted.  If you delete something it remains in the history and the git repository just keeps growing.

gyrog:
I thought keeping all the history was the whole idea of git.
And this is supposed to be very good for source, but I can see that this is neither efficient nor necessarily helpful with binaries.

mavrothal:
> The repo structure on Ibiblio does not currently (Jan 2018) cater for common compiled apps with both 32 and 64 bit variants.

Not clear to me what "common compiled apps" means. All flavors (i486, i686, arm7h, x86_64 etc) of a specific app in the same directory?
To have both 32 and 64 bit version of an app in packages-common (or equivalent)?
Something else?
What is the goal here? To have updates versions of a compiled binary app in all ARCH variants? Other?

 > There are a very few apps already in both packages-common and packages-noarch with bit-length parts to their names.

"Noarch" suppose to have only packages with text files and maybe images. "Common" does not make sense anymore, unless we talk about statically build binaries. Is a relic of the i486/i686-only era of puppy.

wdlkmpx:
A 'universal' petbuilds repo is needed.. apparently. A repo with the most basic woofce apps, including ddrescue, exfat... building blocks to be shared by all woofce pups. I have a list of those apps.. somewhere.

[...]

PR #1165, opened 2018-04-02
14.0 - Partially revert "noarch pet merge"

ninaholic:
These 3 work better from 14.0 version (noarch updates_mgr-0.21 does not work at all in it, maybe in 14.1 too)
https://github.com/puppylinux-woof-CE/woof-CE/issues/1133#issuecomment-377586760

The others I did not test fully but are maybe ok/good. Loading devx then making a savefile appeared to load the SFS on next bootup but really didn't too, perhaps a different sfs_load is needed or it was a strange one-off thing :laughing:


wdlkmpx:
I built a slacko with the 2014 jwm_config and it became even more extremely hard to use it without my keybindings, it was really extremely hard before that. I'll add these changes to my next test build. I guess i'll delete the jwm_config pkg templates which also affect the jwm tray.

This slacko is the only woofce pup with xorg XAA acceleration and working xorg drivers for really old hardware, it's my intention to use it, i have to compile and install all my stuff so i can use it.

PR #1166, opened 2018-04-02
findpkgs: Increase priority of common32/64

woodenshoe-wi:
Some Packages-puppy-*-official package lists have pets marked as
noarch, even though they are not noarch.  @peabee asked for this.


wdlkmpx:
The upups usually have some weird stuff, i guess that's why i don't use them

PR #1168, opened 2018-04-09
slacko 14.0 - remove duplicate FFConvert in OpenWith

ninaholic:
There seems to be two entries in right-click OpenWith list that point to the same place

Screenshot: https://ibb.co/cQHmcH

### DIFF
--- a/woof-distro/x86/slackware/14.0/_00build.conf
+++ b/woof-distro/x86/slackware/14.0/_00build.conf
@@ -243,6 +243,7 @@ rm usr/bin/rsvg-view-3
 rm usr/bin/x264
 rm -rf usr/lib/xmms
 rm root/.config/rox.sourceforge.net/OpenWith/pMusic
+rm root/.config/rox.sourceforge.net/OpenWith/FFConvert
 ln -rs usr/local/bin/7z usr/local/bin/7za
 ln -rs usr/share/applications/gnome-mplayer.desktop root/.config/rox.sourceforge.net/OpenWith/gnome-mplayer
 mkdir root/.config/rox.sourceforge.net/MIME-icons


ninaholic:
after fix

screenshot

mavrothal:
Do we know where the second one is coming from and if this or the other must be removed preferentially?
What about if there is only one in some build instance? Maybe add a condition to run only if more than one are detected?

ninaholic:
Hi mavrothal!

The "ffconvert" one (all lowercase) comes from the ffconvert pet here:
http://distro.ibiblio.org/puppylinux/pet_packages-slacko14/ffconvert-1.4.2.pet

I didn't know where the "FFConvert" one came from but observed that it was created around the middle of 3builddistro-Z, so I searched inside files for "FFConvert" in woof-out* folder and now think the 2nd one with different capitalization probably comes from the pinstall.sh for roxfiler:
http://distro.ibiblio.org/puppylinux/pet_packages-slacko14/rox_filer-130126-s.pet

screenshot 2

mavrothal:
Looks like it.
So if you use another pet for ffconvert you may have no entry. 
You may want to delete the pet link instead. Or maybe something like that  ```[ "$(ls | grep -i root/.config/rox.sourceforge.net/OpenWith/ffconvert | wc -l )" -gt 1 ] && rm ....``` to check if a removal is needed at all.

wdlkmpx:
I get this, and it also has the ffconvert symlink

:ffconvert:|pet|Packages-puppy-noarch-official|ffconvert-1.4.3|ffconvert|1.4.3||Multimedia|208K||ffconvert-1.4.3.pet|+ffmpeg|FFConvert multimedia converter||||

This requires fixing either the ffconvert or rox pet... but who will do that?

Reminds me i might have broken tahrpup's jwm keybindings (evdev driver) when i removed the jwmconfig3..pet pkg template

woodenshoe-wi:
Maybe the rox pinstall.sh should be in a woof-CE pet template instead of in the petbuild.
https://github.com/puppylinux-woof-CE/petbuilds/blob/generic/basic/pkgs/rox-filer/rox-filer-template/pinstall.sh
That way the pets don't have to be changed every time someone wants to tweak the pinstall script.

mavrothal:
I think would be better to have better packages (pets) rather than trying to fix everything in woof.
I think changing the symlink in the ffconvert pet to FFConvert should work either if the pet is installed before or after the rox pinstall script runs.
I believe Peebee, 666Phil and Micko could upload a corrected pet in ibiblio.

woodenshoe-wi:
One complaint I have with putting everything in the pet is that configuration files are not treated specially.  I have been testing a new ROX pet and every time I re-install it my pinboard gets reset and any custom settings are lost.  Because of this I do not recommend people wanting to test the modified version of ROX use the pet but instead just copy the ROX-Filer executable over the one in /usr/local/apps/ROX-Filer/

Maybe the pets need to be split into the part that is safe to install into an existing system and the part that is only safe to use with woof.

wdlkmpx:
I always thought that if there is a jwm_config, there must be a
rox_config. i don't care about it, just saying.

Talking about uploading pets, i've been taking a look at ptheme and
it's not that bad. Adding themes and fixing a few issues here and
there and now i like it. The current global theme i'm using looks like
if it was configured with jwmconfig3 or something, depends on the
ptheme, so i think these 2 packages should become 1

I'll add commits with fixes and enhancements, although jwm itself does
not look better than older versions... but it's what there is.

I see the jwmthememaker can be tweaked a bit to produce stuff for the
woofce jwm_config.

So everybody should prepare to switch to the woofce jwm_config which
will default to "Original Pup", but that doesn't mean i'll make all
pups use the woofce jwm_config, most pups are already using it.

I remember someone said the current jwm freezes their system, if there
are patches then we should place them in a future generic .petbuild
for it.

But jwm cannot be a common app, it links to libjpg, libpng, etc...
these libs change sonames very often. The only solution is to use
technosaurus' patched version with tiny image loaders that link
statically or something... his pull request was not merged by Joe and
i doubt it ever will.. the patched version needs to be patched to
resolve conflicts with newer jwm code, i can do that.

But the problem is always uploading pets and updating the pkgs dbs to
reflect the changes.

ninaholic:
Nice to see more ideas/discussion here. I try to use simplest approach, so I vote just merge this to get rid of duplicate entry, this can be reverted if someone wants to try something else later :laughing:

PR #1169, opened 2018-04-12
Latest kernel 4.16.1 for 64bit cpu AMD/Intel

stemsee:
Tested on i7 5500U and Atom x5. 1000mhz.


mavrothal:
That is a really big file to look at... can you offer some info as to how does it compare with older configs? ie what modules existed in previous kernels but not built in older kernels are now are now build, any old one not built anymore, any differences in built-ins vs modules, the size of vmlinuz and the overall size etc. Also which puppy was tested against.
You may also want to put it out there for some more testing by users before it becomes "puppy official"
Please do not close this request, till more info or opinions are added (or maybe someone pulls it)

wdlkmpx:
The file does not comply with the "naming convention" the other kernel configs follow.. but that's not reason it's not getting merged (by me at least).

Currently the kernel-kit has basically a limited collection of LTS kernel configs. And other special configs, i.e.: stemsee's config for cherry trail tablets: #1144

This pull request should be kept open forever, because i'm writing this little guide here.

Currently missing kernel configs:
- The latest working config for LTS kernel 3.18 (missing)
- A more suitable config for LTS kernel 4.9 (to replace the current one) (disable some aufs features)
- The latest working config for kernel 4.14 (probably steal from fatdog)

As you can see, the current kernel configs show an outdated revision, but that's fixed just by renaming the file, in some cases aufs might be broken with latest LTS revision, so that's why i haven't  renamed them yet, because i haven't tested the results, and replaced the config with the one generated by 'make oldconfig'.

However having a repository with extra compatible kernel configs for woofce would be the right thing to do. So a bunch of kernel configs can be added.

But until then, as git is hard to master, and we have a single kernel config for each LTS release... and  I don't want to discourage @stemsee ... maybe he should add all of his kernel configs to his smokey01 account or something (in a special directory), i will place a link to it in the kernel-kit's README.

I'd like to see an updated version of his cherry trail config, he might want to notify us here or through email (my username at gmale) when a new kernel config is added or something.

Or open a pull request to rename ```DOTconfig-4.14.20-Atom-cherrytrail``` -> ```DOTconfig-4.16.2-Atom-cherrytrail``` with updated stuff..

As usual, if someone adds a special config or updates/add an lts config, then that's a strong candidate for merging, but when adding, a little explanation about the original source must be given. .i.e: All the kernels i added were based on older woofce/fatdog kernel configs.

As usual, if you have improvements or something that should be disabled or enabled in the current configs, just open a pull request modifying one or more of the existing files or let us know what SPECIFIC_CONFIG must be enabled or disabled.. then maybe someone with commit access can do that..

stemsee:
Ahh! ... OK. Well, without commenting on a complete diff comparison with fatdog kernels, or Easy- OS kernel, (which I compared yesterday!). Fatdog kernel 4.14.12, does not support overlayfs builtin, and does not have mtd block device builtin, so cannot find files at boot time located on such devices (mmcblk0p0/1/2).  Also for those kernels, there seems to be a conflict with how the usb1/2/3 are arranged; on my hp tablet with one usb port ... I found that usb3 builtin while 1 and 2 are modules works best. But anyway, in my experience, having usb 2 and 3 builtin, when device has one physical port, caused problems. Fatdog kernel also uses ext4 for ext2, and misses full ext3 support (or handles ext3 poorly), Easy-OS kernel keeps them separate as does mine ext 2/3/4 builtin. My config excludes the older/rarer filesystem types, jffs, reiserfs, minix, amiga,  and others, basically supporting builtins ext, btrfs, NtfS, fat, udf, overlayfs, aufs. Also, more i2c and hid support is mostly builtin.  One could say this is definetely a config for newer intel/amd hardware.

wdlkmpx:
Yeah i'm aware fatdog does not have the ext2 module. I think i added
proper ext2 and ext3 support for the 4.4 x86 kernel configs when i
added those configs to the git repo (which i extracted from one fatdog
sfs). I also enabled a few more modules.

For the 4.9 kernel config, i just took a config from lxpup, which now
i see has the complete set of aufs features enabled, this is not
advisable..

When a new feature is needed or unneeded by the project to support/fix
stuff i enable it in all the kernel configs that support it. That's my
unification approach.

The officially supported woofce kernel configs should be compatible
with new and old hardware although this has some drawbacks, but it
seems to work for the most part.. at least in my machines.

The woofce approach is more or less archeological.

I rarely build kernels by the way, so the project needs the 'kernel
guy', but as far as i know you don't use the kernel kit, all the
kernel configs in the kernel kit can be compiled in an automated way
or through steps to choose stuff, well at least the x86 stuff. Unless
there are bugs in the build script, haven't use it in some time.. it's
a bit complex.

Now, again, i invite you to create a page with all the kernel configs
you produce (properly tagged).. and explaining the features that make
then unique. I'll take some of them, the special ones which will be
updated in the git repo too... you can also let me know about a
special config if you wish, i'll take a look and add/update it myself
so you don't have to go trough the git hell...

stemsee:
Actually I enjoy the prestige of having thousands of lines of code attributed to my handle! But fine, Please specify the config naming algorithm, instead of letting me work it out. There is also a lot more to the kernel configs I make, which include exposing hardware irq to virtual OS, to assist near native speed emulation on bare metal. 

SUKK (stems unattended kernel kit) started with kernel kit, original by Iguleder, then re-written by 01micko; I wrapped the existing code in automation. It can build three kernels consecutively, i386 no-pae for single thread cpu, 300mhz, configured for older hardware; and i686-pae, and x86_64 kernels, unattended. 

The present kernel kit has also undergone lots of development and improvements, I tested it out a while ago. But yes I agree with your logical praactice of providing configs for LTS kernels, makes sense, and also configs for special cases.

I don't have an acount with smokey01 ... yet! But the woof-ce kernel compiling thread should suffice.

ninaholic:
Hi wdlkmpx,

The 3.18.22 kernel config worked for me in 3.18.104 when I renamed it to 3.18.104, and has always been the one I've used for 3.18.x since:
https://github.com/puppylinux-woof-CE/woof-CE/blob/legacy/kernel-kit/configs_extra/DOTconfig-3.18.22-slacko32

I've always been curious if there was anything that should added or removed since .22 in it. I can try it again with 3.18.105 then upload this config with that name if successful, or just add it as 3.18.104 for now and we can change it later as new point versions are tested if that's better. I like the idea of having all the LTS kernels as options too. :)

wdlkmpx:
@stemsee i can add you as the author of the commit and all the stats
go to your account, it's no big deal for me. although when renaming a
kernel config i see the github stats dont reflect the change, it's
like adding a new big text file, that's why i was thinking of another
method to update configs... i.e: a config file that only contains
kernel versions, that way the kernel configs remain with a static name
and it's easy to trace all the changes... i think this must happen. It
must happen.

Well, i'm interested in specialized kernel configs for modern x86
tablets and other specialized configs probably, i know that support
gets better with newer kernel versions, i.e: compatible sound drivers
may appear in kernel 3.17, so the kernel config is updated until
(almost) everything works ok..

I foresee a lack of communication in the near future, we can work with
this method: you open a pull request, i close it, and add/update the
file with the proper name and stuff impersonating you as the author,
so everything goes to your account.

@ninaholic in LTS kernels in some cases some features are added at
some point, and there might be some slight differences and changes
that break the AUFS patches, it only takes a few minutes to produce
the kernel config with the latest revision, as long as it compiles and
works, that's enough for the 3.18 kernel. Aufs is broken for lts
kernels 3.4, 3.2 as far as i know. it only takes a few minutes to
produce the updated kernel configs for i686 and x86_64.

You can add the i686 or i486 version, whatever, just run build.sh and
wait until it starts compiling the kernel... ctrl-c and grab the
.config, rename it following the "naming conventions" and that's it.

01micko:
Sorry guys, this is old and probably no longer relevant. Please correct me if I'm wrong.

Issue #1142, opened 2018-02-17
Packages-templates is unmaintainable


wdlkmpx:
At least for 1 person.

I was looking in the woofce tree to see what's there. And i thought hmm /usr/share/examples should be deleted, really what is it doing there.. hmm. There are broken .xpm symlinks that should be deleted. Hmmm ... Hmm.

Then i tried to open colorcir.ps with ghostscript. ghostscript segfaulted

The ghostscript pkg template is messing up the package. Both processed and unprocessed pkgs have basically the same size, the unprocessed pkg is slightly bigger, like 0.2MB for debian stretch.

Even though i fixed a lot of pkgs templates, it just upsets me that there are many more to fix.

This is the fix i propose: 

1) delete all pkgs templates that are currently not being used by the default builds (also delete belpup if no one is maintaining it)
2) heavily simplify all the remaining pkg templates
3) only allow _PRE & FIXUPHACK, icons and stuff should to rootfs-skeleton.. seriously.

FIXUPHACK should have a fixed list of files to delete. This "dynamic" stuff is really dangerous, and 2createdpackages should be heavily simplified as well, it has broken code and very unreliable code and logic.

woodenshoe-wi:
About point 3.

Adding stuff to  rootfs-skeleton means it is included whether or not that package is in the build.

Wouldn't it make more sense to add PLUSEXTRAFILES and PLUSEXTRADIRS to the templates, or even change 2createdpackages to only add files from the templates and not delete anything except using the FIXUPHACK script?

wdlkmpx:
Hmm well, PLUSEXTRAFILES & PLUSEXTRADIRS should be default behavior with or without those files. If pkg template dir exists, copy everything to the pkg overwriting existing files.. then post-process with FIXUPHACK.

FIXUPHACK can be inside or outside the pkg template ```pkg_FIXUPHACK``` or ```pkg/FIXUPHACK``` .. this is already implemented, that's why i deleted a lot of dirs, otherwise some dangerous code gets executed..

wdlkmpx:
There's something that worries me and is  the fact that the system does not process debian control files. Many packages installed from the ppm will be broken. It's a miracle that the thing works, in slacko all the doinstall.sh files are executed that ensures everything is in place.

A puppy install requires the whole packages-templates dir for deb based pups and some essential fixes for the slackos, so that puppy can become more scalable with very barebones installs.  i think this design flaw is not fixable.

[...]

PR #1172, opened 2018-04-16
simplified 2createpackages & packages-templates (beta)

wdlkmpx:
This is a work in progress that will take some time to finish.. and requires testing in all the woofce pups, so far i tested in dpup and a modded slacko14.0

Related discussion in: #1142 

The changes are explained here: https://github.com/puppylinux-woof-CE/woof-CE/issues/1142#issuecomment-370258764

This is a preview for everyone who wants to test. I build dpups with these changes applied, i fixed the most important bugs but there are more. I think these changes are also required to add stuff for dpup rc-1.. can't remember.

How to test:
```git clone -b temp --depth 1 https://github.com/wdlkmpx/woof-CE woof-CE_w```
and follow these instructions: https://github.com/puppylinux-woof-CE/woof-CE/wiki/Git---users (starting from "Choose target distro")

The new packages-templates stuff will be more 'static' and more portable to other build systems (i.e; woof-next)

I'll update the pr when the time is right..


wdlkmpx:
I updated the pull request.

It's a big change, and you are not supposed to notice it, unless you have custom stuff. After this, 2createpackages will be less likely to create broken packages, now the openssl binaries and the dpkg binaries for deb-based builds are also included. so the build might be 1mb bigger. it it's 3-5mb bigger then there's something to fix.. or maybe not.

It's a big change, but in the grand scheme of things, it's one of the small changes towards a greater goal. Small change.

I uploaded a snapshot to:
www.01micko.com/wdlkmpx/woof-CE-2cp_pt-20180511.tar.xz

Feedback is needed, build any pup and post the iso here if you wish. So far i don't see any breakage in dpup.. if there's no feedback i'll push this to testing in one or two or three weeks anyway.. causing trouble probably.

PR #1174, opened 2018-04-17
Re-add 3.18.x kernel

ninaholic:
3.18.22 tested and works for .105. This was supposed to be EOL at 3.18.47 but it just keeps going :D


ninaholic:
More conversation here: https://github.com/puppylinux-woof-CE/woof-CE/pull/1169#issuecomment-381420159

PR #1179, opened 2018-04-24
Transfer missing package-db's from ibiblio

peabee:
Initial part of changing location of package-db's from ibiblio to github for new woof-ce builds


wdlkmpx:
This is not mergeable.. all the files are in the wrong place.

It's kind of pointless to have them in git anyway.

wdlkmpx:
Unless no dbs are downloaded from ibiblio or somewhere else then it makes sense. Otherwise we'll be discussing how to get rid (some) of the pkgs dbs from git again.

But i dont see anyone trying to discuss that. This is too random, like 10100% more random than what it should be, trying to fix this might be seen as act of war so we better do nothing..

woodenshoe-wi:
For one thing I keep the common32 and common64 builds on my own GitHub page,  
https://github.com/woodenshoe-wi/common32
https://github.com/woodenshoe-wi/common64

There are no woof-CE common32 and common64 builds.
The files in woof-distro are only merged by merge2out for the build chosen.

If you are sure you want to keep these files in woof-CE I think you should think about adding a new directory somewhere else, not putting them into woof-distro.

woodenshoe-wi:
Maybe the "noarch" one could be directly in woof-distro/ and the "common32" and "common64" ones could be in woof-distro/x86/ and /woof-distro/x86_64/ respectively and then merge2out would merge them automatically.

wdlkmpx:
Ah, downloading them from git.. i wonder if this has any impact in merge2out. i'd like to explain how it was done before, th4 keyword is hierarchy or something.

woodenshoe-wi:
In its current form this pull request would probably make merge2out think that the valid arch choices are:

- arm
- common
- x86
- x86_64

And that common32 and common64 and valid distros.

I think merge2out was designed to merge extra files at each level of the hierarchy, so the following could work:

* woof-distro  
  - noarch package list 
  * arm (directory)
  * x86 (directory)
    - common package list
    - common32 package list
  * x86_64 (directory)
    - common64 package list 

An extra complication is that there could be multiple versions of the package lists, the ones in the master branch, the ones in the testing branch, the ones in any other branches...

I wish woof-distro was its own repo with only one branch, or that each distro was its own branch but there would not be a "master" and "testing" version of each distro.

peabee:
I don't mind where the 5 package db's go as long as they are available on github to be downloaded and also to be maintained:
common, noarch, common32, common64 could be in a new woof-ce branch?
wary5 goes into ubuntu-x86

wdlkmpx:
by the way @peabee

i found the reason free_coord() (frontend_funcs) fails sometimes: 

this is the message sent to rox. what it expects to find in the iteration for next drv:
```
...
   <Path>/root/.pup_event/drive_sdc19</Path>
   <X>249</X>
   <Y>678</Y>
...
```

and this is the result

```
  <icon x="250" y="678" label="sdb2" args="drive ext3">/root/.pup_event/drive_sdb2</icon>
.
.
.
 <icon x="250" y="678" label="sdc19" args="usbdrv ext2">/root/.pup_event/drive_sdc19</icon>
```

249 is not equal to 250, the script needs to see 24* actually (in order to continue iterating), so that's where it stops and all the following drives appear in the same location..

This is why the bug appears to be somewhat random...

peabee:
> I can put all the files in their correct place, but are you sure you want this?

I vote YES - github is a better place to maintain the package db's and to record any changes

Many thanks for the rox disk icon bug fix.

wdlkmpx:
does that mean that i have to block certain pkg dbs from being
downloaded from ibiblio or all of them? or should i just place the
files there and let the script tell you the pkg dbs at ibiblio are
different and offer a merge operation between the 2 or something?

there are people here that might want to say something. keeping an eye
on github and participating is a must for every "builder" or
developer, if they don't do that i'm not going to reach out to them
privately.

i follow my own agenda, and i struggled way too much to keep it in
sync with woofce, i lost way too many code lines because of that, and
that won't happen again..

peabee:
That all sounds far too complicated.....
"Old" systems already released and in use will go on using the ibiblio package db's.....
"New" systems will use the github package db's for both build and PPM by altering their DISTRO_PET_REPOS
Changes will be made accountably and reversibly if necessary to the github versions.
If significant changes are made to the github versions which need to be made available to "old" systems then the ibiblio versions will be brought into sync
but in general, the ibiblio versions will no longer be maintained and will only be for "old" systems.....
Thats how I see it anyway.

wdlkmpx:
i get it, it's your wish. let it be known that i didn't plan that..

peabee:
So.............
how about making a new branch "pkgs_common" with just the 4 common package db's:
common, noarch, common32 and common64
and add wary5 to the woof_distro/ubuntu folder.
Then I will make pull requests for selected DISTRO_PET_REPOS amendments.....
Thanks

wdlkmpx:
You shouldn't make pull requests, @mavrothal or @01micko should have
granted you commit access long ago.

do you know how to deal with branches?, i guess you don't. a linear
and only active branch is the only solution for woofce. i don't like
merging endless commits to another branch, i prefer rebasing,
everything looks clean that way.

other branches than testing i don't manually update become outdated quickly

i really doubt git is even the right tool for the project, everything
should go back to the primitive old days.

woodenshoe-wi:
I don't think woof-CE should have any more branches than it has already.  I don't even know if it is possible to build from the master branch because all the bug fixes go into the testing branch along with the new changes.

If you mean a separate repo at the puppylinux-woof-CE organization, that would allow making changes to the package lists without touching the woof-CE repo, but then there shouldn't be duplicate copies in the woof-CE repo.

wdlkmpx:
i will the add files and those files only, will point directly to the raw version of the github pkgs dbs. so update are possible from the ppm.

the same will happen with some other pkgs dbs that are not likely to get updated by the original creator of the file..

PR #1197, opened 2018-05-08
shared-mime-info 2

peabee:


wdlkmpx:
by the way, i see suspicious stuff in the upupbb pkgs dbs, pet packages that should be compat distro packages (gtk, ffmpeg,etc).

i haven't tested but ubuntu is not a barebones distro like slackware, there is a dependency hell that is very easy to break

PR #1202, opened 2018-05-14
pup_event_frontend: remove free_func() and free_flash_func()

wdlkmpx:
there is the freememapplet_tray app running in every pup
so you have an idea about storage usage.

if the icon turns red, it means the personal storage is getting full

frontend_timeout already runs a lot of apps every 6 seconds, including ps, grep
and cddetect_quick needed for an optical drv + systemd-udevd
greps /proc/mounts

this extra code is overkill

@gyrog has patches for these functions for his overlayfs stuff

if this is merged those patches are probably not needed anymore...

maybe these functions should be moved to somewhere else?

### DIFF
--- a/woof-code/rootfs-skeleton/usr/local/pup_event/frontend_funcs
+++ b/woof-code/rootfs-skeleton/usr/local/pup_event/frontend_funcs
@@ -324,59 +324,7 @@ remove_pinboard_func() { #needs DRV_NAME (name of entire drive)
  done
 }
 
-free_func() { #called every 4 seconds.
- case $PUPMODE in
-  6|12)
-   pup_rw="$(stat -c %m $(readlink -f /initrd/pup_rw))" #get mountpoint regardless of filetype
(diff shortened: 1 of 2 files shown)


wdlkmpx:
The question is, does freememapplet work properly in all pupmodes?

The current version is http://distro.ibiblio.org/puppylinux/sources/f/freememapplet-2.8.6.tar.bz2

is this code in rc.sysinit good or bad?

```
[ -s /initrd/tmp/sfs_ram_sizek ] && ALLOCK=$(($ALLOCK + $(cat /initrd/tmp/sfs_ram_sizek))) #proper size
```

wdlkmpx:
I briefly tested freememapplet. When you hover over the freememapplet tray icon you'll see numbers, accurate numbers.

Pupmodes 12,13, with savefiles/savefolders: 100% accurate info
Pupmode 2: 100% accurate info
Pupmode 5: it's ok. i remember a discussion about this

So this is safe to merge, i was worried my reasoning was failing me again ahaha..

PR #1204, opened 2018-05-16
add gtk-update-icon-cache

peabee:
looks like was added as a separate package in artful
http://www.murga-linux.com/puppy/viewtopic.php?p=992169#992169

### DIFF
--- a/woof-distro/x86/ubuntu/upupbb/DISTRO_PKGS_SPECS-ubuntu-bionic
+++ b/woof-distro/x86/ubuntu/upupbb/DISTRO_PKGS_SPECS-ubuntu-bionic
@@ -244,6 +244,7 @@ yes|gtk_theme_flat_grey_rounded||exe
 yes|gtk_theme_stardust_zigbert||exe
 yes|gtk_theme_stark-blueish||exe
 yes|gtk_theme_stark||exe
+yes|gtk-update-icon-cache|gtk-update-icon-cache|exe,dev,doc,nls
 yes|guess_fs||exe
 yes|gutenprint|libgutenprint2,libgutenprint-dev,libgutenprintui2-1,libgutenprintui2-dev,printer-driver-gutenprint|exe,dev,doc,nls
 yes|gview||exe


peabee:
having to update both testing and master is a pain.....which should we now update and how will any changes get merged into the other branch?

wdlkmpx:
@peabee i don't think this is right:

yes|gtk+3_dev|libgtk-3-0,libgtk-3-dev,libgtk-3-common|exe>dev,dev,doc,nls

The same applies to the ffmpeg app.

This issue is what it's probably causing comments about how the ppm is failing to install many packages in upupbb.

Puppy is not good at scaling from small to big due to flaws in the build system/package_management/rootfs design that requires some serious changes to even make sense a bit.

Well at least the rootfs is becoming more "agnostic", that was my main focus since the beginning. And that's the only really important part.

In short, to make a reasonably stable build with DEB-based distros, you must use all default libs and most essential apps (ffmpeg) and override them with a zz package if you know that it won't cause any problems.

Just add some extra updated gtk apps and other apps that don't interfere..

I haven't tested upupbb... i was just looking at the PKGS_SPECS. I'm planning to perform more cleanups on the PKGS_SPECS and pkgs DB's following order 66. But this doesn't feel right, events must be happening at a faster and more brutal rate.

I might try to fix some issues in upupbb, but for one person alone to test and fix everything, hmmm something is not right...

Do you have this problem in upupbb:

After installing a package in the ppm, a splash pops up time and again (while some heavy processing is happening), at least 6 times.

wdlkmpx:
Hmm i read it somewhere else then, it was sc0ttman saying that i
think. And others probably saying something installs fine in XX deb
distro, but not in upupbb... along those lines

Then the ppm issue is a dpup specific issue. Looks like signals
getting triggered when they shouldn't.. i agree with some nice looking
and mildly complex gtk dialogs, but the ppm and a few other apps have
unreasonably complex gtk dialogs, only fixable by a guru of the
telecommunications industry.. and gettext'ing the return 'strings'
makes the case even worse.. for other locales than EN

So if it's not a problem in the slackos or the upups, then i think
i'll just leave that bug floating around

wdlkmpx:
I installed gnome-mpv in upupbb. No issues to report.

Packages succesfully Installed or Downloaded 
- adwaita-icon-theme_3.28.0
- gnome-mpv_0.13
- humanity-icon-theme_0.6.15
- libass9_0.14.0-1
- libatk-bridge2.0-0_2.26.2-1
- libatspi2.0-0_2.28.0-1
- libcolord2_1.3.3
- libgtk-3-0_3.22.30
- libgtk-3-common_3.22.30
- libjack-jackd2-0_1.9.12
- liblua5.2-0_5.2.4
- libmpv1_0.27.2
- libpulse0_11.1
- librest-0.7-0_0.8.0-2
- librubberband2_1.8.1
- libsdl2-2.0-0_2.0.8+dfsg1
- libuchardet0_0.0.6-2
- libva-drm2_2.1.0-3
- libva-wayland2_2.1.0-3
- libva-x11-2_2.1.0-3
- libwayland-client0_1.14.0-2
- libwayland-cursor0_1.14.0-2
- ubuntu-mono_16.10+18.04.20180421.1

Notice these 2 pkgs, a gtk3 pet pkg is already installed.
- libgtk-3-0_3.22.30
- libgtk-3-common_3.22.30

libgtk-3-common_3.22.30 is a big package, but 2createpackages discards 99% of it, but here we have the full package installed.

No matter how you look at it, the pet pkg does not make sense, you only need this line

```
yes|gtk+3|libgtk-3-0,libgtk-3-dev,libgtk-3-common|exe,dev,doc,nls
```

libs and important stuff go to _EXE
_DOC and _NLS are discarded (libgtk-3-common)
-dev goes to _DEV

This was just a test to make sure my changes in dpup don't cause issues in other builds. 

I waste my time explaining all this stuff so i hope it gets understood..

PR #1207, opened 2018-05-22
Remaster - copy all *drv .sfs

peabee:
see:
http://www.murga-linux.com/puppy/viewtopic.php?p=992793


wdlkmpx:
The usb-modeswitch pet pkg is not compatible with deb-based pups i guess. It has a tcl script.

This fixes the issue
https://github.com/puppylinux-woof-CE/petbuilds/commit/0e9b6bf7f53bbaf2d4bd5bc31bff70a271f73232

So the pet pkg must be recompiled and reuploaded. But the resulting pet pkg must be tested, that's the difficult part.

I think i'll just make all the deb-based pups use the compat distro DEB pkg..

Issue #1043, opened 2017-06-09
petget in pupmode 13 is broken


01micko:
Here is the problem:

The default for petget is to save the package directly to the save file.
This is bad and broken.
* a pinstall.sh, doinst.sh (and probably debian equivalent) is executed at '/' - which goes to the tmpfs - that makes it broken.
* it is the understanding of most users on the forum that if they install a package, it gets nuked if they don't save the session - it doesn't. That's bad.

Everything works correctly if you check the box  `install to tmpfs instead of savefile, till session saved`

IDK why we even save it directly to the savefile in the first place :confused: . IMHO that option should be removed and if not, any post install script should be executed at the place where the install happens.

ninaholic:
@01micko I notice this as well with pets, and I think you described both issues perfectly. Seems like "install to tmpfs instead of savefile, till session saved" should be default?

wdlkmpx:
#boot from flash: bypass tmpfs top layer, install direct to pup_save file...

That code has been there for years. I'll take a look

jamesbond3142:
On Fri, 09 Jun 2017 14:25:26 -0700
Mick Amadio <notifications@github.com> wrote:

> 
> IDK why we even save it directly to the savefile in the first place :confused: . IMHO that option should be removed and if not, any post install script should be executed at the place where the install happens.
> 

Historical perspective: 
---
When this was devised, target machine was 128MB machine, with perhaps 64MB absolute minimum. The amount of RAM reserved as tmpfs layer was miniscule, perhaps 64MB at most or may be less. If you install directly to root ("/"), which means it gets installed to that tmpfs, installing a few large packages can easily overwhelm it. It was confusing when people created a 1GB savefile and yet installing a few large pets suddenly popped up warning that the savefile is full.

The workaround this was to install one pet at a time, and then forcing a snapmergepuppy. Another more drastic workaround is to reboot after installing large package. But this of course don't work if the size of the pet is larger than tmpfs to begin with (see below for the size requirement calculation).

Because of that, Barry conceived the idea that pet installation should go directly to savefile (because that's where we want it to end up anyway) to avoid clogging that tmpfs.

PS: The required storage for package installation isn't the size of expanded pet. It's the total size of the following:
a) size of pet itself 
b) size of expanded pet
c) Depending on how you do it, you may need another copy of (b)

[...]

Issue #822, opened 2016-06-30
Overwriting /lib/$ARCHDIR breaks puppy


mavrothal:
[3builddistro](https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-code/3builddistro-Z#L495) generates symlinks for /lib/$ARCHDIR /usr/lib/$ARCHDIR /usr/bin/$ARCHDIR that link to upstream directory. When a package with such a folder is installed, will overwrite the link and break the built.
Test case install libnih1 in Xenialpup64. Will overwrite the  /lib/x86_64-linux-gnu symlink and the build will break till you remove the /initrd/pup_rw/lib/x86_64-linux-gnu folder.
As a matter of fact as soon as you overwrite or delete the symlink, puppy will break.
PPM could be modified to force pet packages or slackware packages to respect symlinks (use the -h option in tar) but dpkg-deb or dpkg, do not provide such an option.

It might be better to have  /lib/$ARCHDIR as a directory and link its content to /lib/<filename>

mavrothal:
Here is a workaround patch to extract in /tmp and copy each file and link avoiding overwriting the $ARCHDIR symlink.

```
--- a/usr/local/petget/installpkg.sh
+++ b/usr/local/petget/installpkg.sh
@@ -261,7 +261,24 @@
   [ $? -ne 0 ] && exit 1
   echo "$PFILES" > /root/.packages/${DLPKG_NAME}.files
   install_path_check
-  dpkg-deb -x $DLPKG_BASE ${DIRECTSAVEPATH}/
+  # Workaround for overwriting x86_64-linux-gnu symlink
+  # Should also be done for i386-linux-gnu too
+  mkdir -p /tmp/$DLPKG_BASE
+  rm -rf /tmp/$DLPKG_BASE/*
+  dpkg-deb -x $DLPKG_BASE /tmp/$DLPKG_BASE/
+  for f in $(find /tmp/$DLPKG_BASE \( -type f -o -type l \))
+  do
+	xf=$(echo $f | cut  -f 4-30 -d "/")
+	if [ "$(echo $xf | grep x86_64\-linux\-gnu)" = "" ]; then
+		mkdir -p /$(dirname $xf)
+		cp -a $f /$(dirname $xf)/
+	else
+		xpath=$(echo $xf | sed 's/x86_64\-linux\-gnu\///')
+		mkdir -p $(dirname $xpath)
+		cp -a $f /$(dirname $xpath)/
[...]

PR #1224, opened 2018-07-07
get rid of the multiarch symlink

wdlkmpx:


wdlkmpx:
This is a pull request and an issue at the same time.

This looks like small change - and it is, thanks to important changes in 2createpackages and packages-templates.. i built a dpup, installed and compiled a few apps.. no issues so far. 

But this change probably breaks all existing installations and alternatives to petget that also implement the symlink. Apparently some people also added the symlink to some sfs's and pet pkgs..

For deb-based pups: many package templates no longer apply 'their logic' to the pkg.. many pkg templates basically are there to create symlinks for older libs... some old apps may no longer run, the upups have many old apps from the 'common' repo.

But this is more a call for a discussion.. is it morally, politically, spiritually wrong to get rid of the multiarch symlink after so many years?

peabee:
I don't think I know what "the multiarch symlink" is..... sorry! Can you say what it is and what are the consequences of removing it? And why do you want to remove it? Is it causing problems? If so, where have they been reported?

"the upups have many old apps" - is this necessarily a bad thing? If the apps work and there is no need to update them they can still be used?

"this change probably breaks all existing installations" - this sounds bad!

At the moment, not understanding, I vote to not do this change - unless you can say why it is needed.

wdlkmpx:
The multiarch symlink, also known as the 'multiarch hack' or the 'hack for multiarch' is a hack to alter the compat distro directory structure replacing the multiarch directory with a circular symlink to avoid breaking stuff.

- /lib/i386-linux-gnu
- /usr/lib/i386-linux-gnu
- /usr/include/i386-linux-gnu
etc

These are actual directories with many subdirectories.

2createpackages is faster and even the system seems to boot and work faster (might be the placebo effect)

In the past this was not feasable because 2createpackages and packages-templates created a bunch of broken packages because of the design itself. It's been an insane amount of work to revert or delete most of the poisonous stuff to get basic things to work. And this is just the tip of the iceberg.

peabee:
Still not sure what problem you are trying to fix - and for such a fundamental and disruptive change it must be a BIG problem to justify......I haven't seen any such problem reports.....
"And this is just the tip of the iceberg." is worrying.....

wdlkmpx:
I was just adding some suspense and drama to make a point.

In the end this change may not be that dramatic.. by creating symlinks
to some really old libs the pkgs templates might actually introduce
bugs due to incompatibility, might be better to add the actual old
libs that were in puppy4, wary .. e.g: openssl

The few libs that are backward compatible are basically libc, zlib?,
gtk2 and other libs, but when you add openssl, and some other libs
that change api (too) often you have an incompatible old app.

For exisiting installations, i don't know what happens, i never do
upgrades, but i assume something bad will happen. It doesn't bother
me, i follow my own timing

The logic is simple, the more compatible puppy with the compat distro
is, the less hacks are needed and thus stuff works faster, more stuff
works out of the box, etc ...

That's why i said 'it's the tip of the iceberg', there was too much
incompatibility and this is just the beginning and it doesn't mean
something will follow, there's a limit..

But in my mind i see the big project supporting and working in 2
modes: with and without the compat distro package manager... for the
'with' part there's a long and winding road. By installing debian and
see that i basically don't know what to do with it, it came to my mind
that i need a standard, known desktop in all my installs, a
lightweight one, thus making me regret about many things i did.. and
all the time i wasted

peabee:
UpupBB has no components coming from -puppy-common-
I venture to suggest that package-templates is not the place to be creating symlinks to satisfy old lib requirements - these should be created in EXTRA_COMMANDS= in _00build.conf as they are specific to whats included in a particular build
Some way of letting builders know what the dependencies are would of course be useful.

I still vote against "get rid of the multiarch symlink" as it is of limited benefit for the amount of disruption it might cause; puppy has existed for a long time with it in place without any major problem emerging; and everybody knows this is the puppy model and lives with it.

Would be nice to hear from others though.....

woodenshoe-wi:
> Would be nice to hear from others though.....

I'm not going to vote for or against the changes, but if you were wondering what problems the changes were trying to fix here are two links.

https://github.com/puppylinux-woof-CE/woof-CE/issues/1043#issuecomment-307569077
https://github.com/puppylinux-woof-CE/woof-CE/issues/822

The problems have already been fixed in other ways, but I guess they got some people (including 01micko) thinking that the symlink was causing more problems than it was fixing.

Theoretically as long as the multiarch directories are listed in /etc/ld.so.conf everything _should_ keep working, but without testing you can't be sure.

I guess that is the reason that at least two branches are needed in woof-CE, one to build test builds and another to build stable releases.

If problems show up maybe the changes aren't worth it, but you can't know what the problems are without testing.

wdlkmpx:
This looks like a major change, a bit disruptive. The current changes in testing do need some testing and feedback.

It's just doesn't get any better than using git and a platform like github to make everything possible and easy to test and provide feedback. It's easy to spot issues, mistakes and meaningless changes.

For example the first changed line in https://github.com/puppylinux-woof-CE/woof-CE/commit/422e17d505fe081fde1fde3654828593bbe4727e is completely meaningless, what does that mean?.. well it means that i was trying to do something but didn't get around to actually do it, or that the file is not that latest revision.

Hmm i don't know if this change made it possible.. but for the 1st time i'm seeing pup-volume-monitor work with the compat distro glib in a deb-based pup. I don't know if it's a glib update/fix.. hmm... what can it be.. i remember fixing warnings and stuff but i think it wasn't enough to make p-v-m work in a deb-based pups.

I removed the pet pkg and built a new dpup iso just to make sure, and... it's true, it's working fine.

wdlkmpx:
* This is something that should be announced in the blog and on the forum

I posted patches for puppy-get and Pkg (on the forum), i guess that's enough. I'm not aware of other petget/PPM alternatives. Whether my patches are ignored is not my business... woofce is just one of several forks

The more this is delayed, the more innocent souls will perish in the abyss of distant memory.. and it's already done and my time already wasted

After the pr is merged i'm willing to test and try to fix issues in isos provided by anybody as long as there's an accurate bug report, however fixes may happen on weekends..

* In this pr i explained what the issue is about, so you can reuse my words or you should probably use your own wording and add this:

DISCLAIMER: You’re not getting anything. You understand? Nothing. You’re getting nothing.

wdlkmpx:
This is the final version of the pr. I noticed github has been improving a lot lately, you can see details that were not available before

wdlkmpx:
Last time i tested this, it was all working "right". But that was only in dpup stretch.

By merging this  some work must be done to fix new issues.. ain't it fun.

It's been almost a year since i opened the pull request.

Now that puppy 8 has been released without any sort of coordination, would it be ok to merge this, maybe next month?

The master branch is up-to-date, only the testing branch would continue for a while. Basically some of the changes i have in mind introduce incompatibilities

So, should this be merged? Should i force my way? Should the deb-based pups be retired if this this is not merged? Should

peabee:
This change should be done in a separate branch - but we then need builds to be made from that branch and isos issued so that testing can be done and impact of "incompatibilities" assessed.
If the impact is that many existing pets become unusable then we are talking about starting a complete new repo which only contains pets that are compatible with this new architecture - that sounds like a very big job?
Retiring deb-based pups would be detrimental to Puppy.
I don't think you should force your way on this - some strategy is needed to progress slowly and carefully.
We seem to be living fairly successfully with the current architecture - are the benefits of doing a disruptive change so big that we want to face the disruption? I currently vote "No".

wdlkmpx:
I created the experimental branch, an unstable branch where things may be discarded and forced updates may happen

This change is already there

https://github.com/puppylinux-woof-CE/woof-CE/commits/experimental

This is a disrupting change because it simply does not force a certain directory layout. All the changes in packages-templates are a prelude to this simple and disrupting change.

Hundreds of files were deleted so that this simple change can happen.

Imagine /usr/lib64 is a symlink to /usr/lib in slacko.. shouldn't it be fixed?

peabee:
So what do you now want to happen? Presumably a 64-bit Ubuntu or Debian build? As my only Woof-CE 64-bit builds are Slackware based presumably these wouldn't help test?? And we don't have any recent Debian 64-bit builds (I think...) so looks like it's @mrfricks BionicPup64 that has to be the guinea pig??

wdlkmpx:
This change affects all 32bit/64bit deb-based builds.

Guess what? Petget doesn't run .deb post-install scripts. Woof doesn't
run .deb post-install scripts.

Compatibility with debian/ubuntu is just a massive hack, there's
really no compatibility with debian/ubuntu... at all.

On the other hand, Slackware is highly compatible with puppy, packages
are properly installed by woof and by petget. We should also add
Fatdog as compat distro, it's basically a slackware

Debian requires a chroot, it's mandatory for a proper build, the
current build stuff (system) provides a build based mostly on hacks..
the hacks replace the post-install scripts. Petget needs hundreds of
hacks then.

Well, i don't expect anything, i only expect this to be merged.

wdlkmpx:
In woof-next, rootfs-skeleton is copied after all the packages have
been processed, it overrides files and configs from the compat
distro.. requiring less hacks.

In testing, rootfs-skeleton is copied before all the packages are
processed, that is wrong, maybe it should be copied twice, before &
after.

This and many more changes should happen, but i no longer feel
confident to apply them. It's about evolution. Evilution.

wdlkmpx:
I noticed i said "disrupting" instead of "disruptive" many times. It's
not my fault.

I'll try to explain again why this change is disruptive.

Because pupsave upgrades will become particularly problematic after this change.

I don't know, but apparently some people have created SFS files that
include the symlink, i don't know much about the pet packages in the
upups, but i can certainly assume some pkgs contain forbidden stuff.
But i guess it doesn't really break the system.

That's why it's disruptive, pupsave upgrades will be way more problematic.

There is a change in the directory structure.

It took a huge amount of work to unlock this level. Level 2000

Other similar projects start with level 9000 because they start from
scratch and don't have to deal with a really bad design and slowly and
painfully attempt to right the wrongs..

peabee:
> pupsave upgrades will be way more problematic
If you are able to describe why this will be, I for one would find that useful.....
How will we go about testing this???

peabee:
First trial build of UPupDD-19.04 with EXPERIMENTAL.......
Good news - it boots ;-)
BUT
LOTS of build problems to fix because legacy links created both in zz_fix pets and in _00build.conf EXTRA_COMMANDS are now wrong

peabee:
Legacy links libcrypto and libssl seem to be created in the wrong place in sandbox3/rootfs-complete/lib but not strangely in packages-upupdd/openssl
Will try to track down where they are being moved...may be my fault?

peabee:
Here you go.....
[experimental iso](http://www.smokey01.com/peebee/downloads/experimental/upupdd-19.04+0-uefi-EXP.iso)
Now what should happen??

wdlkmpx:
Of course it boots to desktop and all the hacks in packages-templates
are applied correctly.. i tested this stuff for months. But when
something is not merged i just forget about it and do something else.

Other custom hacks outside packages-templates may not apply, since
some libs that were in /lib/xx,.so are now in /lib/*-linux-gnu/xx.so

I never do pupsave upgrades, but after upgrades issues with the
symlink and the ppm may arise.

There are other implications, but it's better for people to find out.

I never do upgrades, so people will find out what happens, there is a
reason i tried to completely block upgrades from different
WOOF_VERSIONS.

Currently petget doesn't run .deb post-install scripts, and the reason
is... the control files are not extracted from the deb package. I will
add that change to the experimental branch

Everybody who reads the code or installs .deb packages with petget or
the ppm knows how awful the situation is when it comes to deb
packages.

There is also a reason i don't add new distro compat versions and
totally discourage such practice unless it's a new LTS release.

Imagine peebee releasing upuptt or LXpup upuptt, or LXPUP tt xx... and
people thanking him seeing  how updated and functional those pups have
become.

Ponce.cc or arch linux mean nothing to me, when i want to use lxde in
a pup i compile the latest version of the whole desktop environment.
It compiles fine even in slacko 14.0. gvfs is not needed of course.

wdlkmpx:
This is funny, petget does extract the debian control files, slackware
doinst.sh, etc

it runs postinst in the background, and it immediately deletes the
whole directory, i don't know, but this may cause some packages to
fail to install properly

woof-code/packages-templates/dpkg-deb_FIXUPHACK will be deleted, as it
deletes a binary that is called by some debian postinstall scripts.

The command that is run by dhcpcd6/DEBIAN/postinst does work, if a
debian puppy is created using a chroot, many other hacks can be
deleted. But they won't because they're still useful for cross-builds.

update-alternatives --quiet --install /sbin/dhcpcd dhcpcd
/sbin/dhcpcd5 50 .......

I could explain many things, but i guess it's easier to just do things.

wdlkmpx:
I added 24 commits to the 'experimental' branch. I should have named
it "staging"

https://github.com/puppylinux-woof-CE/woof-CE/commits/experimental

Everything i do is experimental since i only test in 1 pup.

The changes include fixes for bugs mentioned above, better support for
the future zwoof-next directory, random gui changes and stuff.

I maintain several local branches and i decided to merge some of them,
i'll build a new iso. There's plenty of time for people to read the
commit messages, read the commit diffs, test the changes in new
builds, etc.

There's plenty of time to identify issues and revert unwanted stuff
before it's merged to testing.

wdlkmpx:
This is an example of how to provide feedback, i really don't expect
anyone to do something like this, with the possible exception of
ninaholic. This can also happen in wmsgs.

Using the changes in the experimental branch.

2createpackages is _faster_ for deb-based builds.

Build: no errors, no regressions.

I have a pupsave, i have stuff and configs. I previously installed and
compile a few apps.

I replaced the sfs files and rebooted, an update was no triggered,
because i didn't change the DISTRO_VERSION. But things happen anyway,
the update process is to apply hacks, now without the hacks.

The system is stable, there are duplicate libs, as there is a change
in the directory structure, the multiarch syslimk doesn't not exist

Installing a pet pkg, it wiped the multiarch directory and restored
the symlink, but the system is still running fine, duplicate libs.

It's the outdated installpkg.sh from the pupsave. The patched version
after the [BRUTAL] commit does not override the symlink, the huge and
silly code blocks to enforce symlinks were deleted.

So this change wasn't as dramatic as i thought it would be.

It has my seal of approval.

wdlkmpx:
This can be considered "merged" if we assume that the experimental branch is indeed the future of this project..

jamesbond3142:
> In woof-next, rootfs-skeleton is copied after all the packages have been processed, it overrides files and configs from the compat distro.. requiring less hacks. 

Basically the hacks are all done in one place: rootfs-skeleton. Instead of spread over everywhere. So it's simpler to manage and maintain. You may still need distro-specific hacks though.

> I never do pupsave upgrades, but after upgrades issues with the symlink and the ppm may arise.

We don't do upgrades Fatdog. It's too hard.

PR #1233, opened 2018-08-03
Seed new pupdial package directory with current file versions

rserwin1:
Create a package for pupdial and copy the current versions of its files to the package into a dedicated directory, as the basis for updates to pupdial 2.0.


rserwin1:
I discovered that I left out a file but cannot add it to the commit.  Please also copy file /usr/sbin/get_bluetooth_connection to directory /usr/local/pupdial/.

PR #1254, opened 2018-09-27
Network connect: Move network wizard; add connect support for peasywifi

rserwin1:
Connect wizard will offer peasywifi as a wireless option if installed, such that it co-exists with other tools without interference and allows switching between it and other network managers.

### DIFF
--- a/woof-code/rootfs-packages/network_wizard/pet.specs
+++ b/woof-code/rootfs-packages/network_wizard/pet.specs
@@ -1,1 +1,1 @@
-network_wizard-20170623|network_wizard|20170623||Network|376K||network_wizard-20170623.pet|+gtkdialog|Dougal's Network Wizard||||
+network_wizard-2.0|network_wizard|2.0||Network|388K||network_wizard-2.0.pet|+gtkdialog|Dougal's Network Wizard||||
--- a/woof-code/rootfs-packages/network_wizard/pinstall.sh
+++ b/woof-code/rootfs-packages/network_wizard/pinstall.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+ln -snf /usr/local/network-wizard/net-setup.sh usr/local/bin/
+
+if [ "`pwd`" = '/' ];then
+ rm -f etc/rc.d/rc.network
+ rm -f usr/sbin/net-setup.sh
+ rm -f usr/sbin/wag-profiles
+ rm -f usr/sbin/ndiswrapperGUI.sh
+fi
--- a/woof-code/rootfs-packages/network_wizard/usr/local/bin/net-setup.sh
+++ b/woof-code/rootfs-packages/network_wizard/usr/local/bin/net-setup.sh
@@ -1,0 +1,1 @@
+#Placeholder for link to /usr/local/network-wizard/net-setup.sh, to include in remasters.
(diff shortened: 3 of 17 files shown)


wdlkmpx:
new iso for rerwin

www.01micko.com/wdlkmpx/slacko_retro/slacko-5.8.2-g4dos.iso
www.01micko.com/wdlkmpx/slacko_retro/slacko-5.8.2-g4dos.iso.md5.txt

www.01micko.com/wdlkmpx/slacko_retro/devx_slacko_5.8.2.sfs
www.01micko.com/wdlkmpx/slacko_retro/devx_slacko_5.8.2.sfs.md5.txt

rserwin1:
I think we should start a thread somewhere about the slacko 5.8 ISOs, 
ideally in the forum.  That way we can work out the problems together in 
public.

I tried 5.8.2 but soon found that ROX Filer segfaults when I attempt to 
change to the list mode of display.  The screen goes black but with the 
panel still there.   The log message:
Sep 29 11:53:45 puppypc22194 kern.info kernel: ROX-Filer[5078]: segfault 
at 0 ip 0806802e sp bfb4db30 error 4 in ROX-Filer[8048000+9e000]

I see, too, that its ath9k driver sometimes does not connect, as is also 
the case in tahrpup (using essentially the same kernel). Xenialpup's and 
peebee's Bionic pup's ath9k drivers connect.  I use SNS in all of those 
cases.  With 5.8 I tried only pfix=ram (actually pupsave option 0).  
Maybe there is a later version of ath9k that would work better in 
tahrpup and slacko 5.8.  I see in the message log, driver cfg80211 
"Calling CRDA for country: CO".   But CRDA components are not in 5.8.2.  
Most of the tries find the command "ifplugstatus wlan0" returning "wlan0 
unplugged" even though it is a built-in PCI card.  On the rare occasions 
where it detects an interface, the attempt to connect times out and 
fails.  Go figure.

Regarding the analog modem drivers I created and provided a link to, it 
seems to make sense, to me, that they be included in the ISO. The 
supporting "firmware" package could be placed in the repo for slacko 5.8 
or simply be included in slacko 5.8, since analog modem support is one 
reason for users to adopt the otherwise old 5.8.  The 4 drivers are here:
http://www.mediafire.com/folder/g7b4os9oob9k0/Kernel_3.2.102
The "firmware":
http://www.mediafire.com/file/oiv03fxqgwg5blf/pupdial_softmodem-20180911.pet/file

I also tried with a USB wifi that uses driver zd1211rw.  Same problem.  
So, maybe the issue is not just with a ath9k device, but something else.

woodenshoe-wi:
> Months ago i created a forum and tried to lure people to there, but it was a failed attempt... aarf is lonely and talking to himself in the alternative puppy forum,

I'm not saying that the main puppy forum is perfect, but I don't think that trying the same thing (another forum) all over again will result in different results.  Some people might not want to sign up again at another site.

Even the conversation on GitHub has given one forum member the impression that there is a rather negative attitude here.  I hope none of my comments have given that impression, when I say that it is important for people to look after their own well-being first.

A mailing list has also been tried, http://woof-ce.26403.n7.nabble.com/ but it doesn't look like anyone has cleaned out the spam recently...

PR #1255, opened 2018-09-28
yes|ifplugd|ifplugd|exe.dev.doc,nls

peabee:


woodenshoe-wi:
`exe.dev.doc,nls` Shouldn't they all be commas?  Or is this a new syntax I don't know about?

peabee:
Oooops - just copied and pasted - now corrected....

PR #1260, opened 2018-10-01
Network connect: Correct some details; increase SNS network capacity

rserwin1:
The SNS change filters iwconfig output to avoid processing unused lines, on the assumption that the data otherwise exceeds some internal limit on variable contents, reported here: 
"SNS and places with over 80-100 wifi signals"
http://www.murga-linux.com/puppy/viewtopic.php?p=996815#996815

There are no complaints, so I assume it helps.


wdlkmpx:
You probably noticed that sometimes i edit sns and network wizard, when this happens a package can be considered doomed... where compatibility with anything other than the last revision of woofce is not taken into account, because woofce itself has very old distros and it's easy to build any of them anytime.

I guess you also noticed that aside from the 'networkmodules' file,   i haven't made any change that truly breaks backwards compatibility, hmmm..

rserwin1:
Backward compatibility is an issue only for packages that might be 
installed in older pups.  I offer the networking stuff in the forum for 
general use.  Let that be my task, to keep up with your progress.


wdlkmpx wrote:

PR #1268, opened 2018-10-14
slacko 14.0: xvkbd needed for X-Virtual Keyboard

ninaholic:
The one from 14.1 sort of works, though gives warning about font

screenshot 2
screenshot 1


ninaholic:
edit: Nevermind, it still doesn't seem to do anything when I hit letters on it :scream_cat:

PR #1279, opened 2018-11-01
pup_event_backend_modprobe: modprobe only substitute modules

rserwin1:
For kernels 4 and later, loads only modules to be substituted and relies on udev (?) to load the others.  Earlier puppies do not load modules processed by backend_modprobe.  If the automatic loading started with a version of something else, the test for this action should be changed accordingly.

Removed obsolete udev rules file and logic for passed module names because they were implemented and used by only PCI analog modem drivers that are no longer supported.

Removed PCI_OVERRIDE logic and MODULESCONFIG entry for "(none)" which prevented loading of any module for a particular vendor:product ID, because it was added and used for only an analog modem not supported by its assigned driver and is now ineffective because it cannot prevent automatic loading.

Removed usb_storage substitution logic because it is superseded by usb_modeswitch.

Changed all exit codes to 0, since they seem to be ignored except for a logged warning from udev.

Removed the apparently obsolete FASTBOOTMODULESLIST check, because I see no other reference to it as a boot argument or anything else.

Please alert me if any of these conclusions are wrong.

So, for recent puppies, backend_modprobe now modprobes only modules that:
  - override a PCI-device module, based on hardware ID (no defaults)
  - are preferred when multiple modules support a device, if different from initial module

Second commit converts backend_modprobe to a package so that it can be omitted from a puppy if its functions are not needed.  Udev rule and bootmanager check for it.  Also, udevtrace rule cleanup drops writing to udevtrace-modem.log.


rserwin1:
Pupdial should not be affected.  The changes relate to only the analog 
modems that have not been supported since precise or earlier.  It is 
still needed for serial and USB modems, both wired and wireless.  It 
supports wired modems, which pgprs and frisbee-mobile (which use similar 
logic), do not.

PR #1282, opened 2018-11-06
Reduce maximum wait time for wifi availabiility in SNS rc.network & frisbee.sh

rserwin1:
This is in response to reports of slow startup of network interfaces.  The wait times were increased to support an early version of EasyOS, which caused long loading times.  The issue was resolved, making the long wait maximum unnecessary.  Previously, the maximum was 8 seconds; this makes it (arbitrarily) 10 seconds.

In SNS, modules are checked sequentially until loaded or timed out.  So, if the first one checked were not available -- say, because the hardware has been removed but not the connection record -- a subsequent check would delayed by the timeout period.

A more efficient method would be to check all potential wifi modules concurrently, ending with the first one loaded.  I am working that possibility.

Re dhcpcd_dropwait_option: It is no longer used.  Dhcpcd_dropwait 6.7.1 waits only a fixed few seconds for a connection to recover, not a user-specified value conveyed by dhcpcd_dropwait_option.


wdlkmpx:
For wired ethernet, there's a quick fix... for earch wired interface launch a dhcpcd daemon that never exists, that way connection eventually happens, permanent connection.

I guess this should be the default behavior when no wireless interfaces are available.

For wireless ethernet there is wpa_cli or something.

I have a rewrite of network wizard, which includes a command line user interface (dialog - ncurses)

It's supposed to be as easy to set up as SNS but with advanced features (network wizard) showing splashes as actions happen (frisbee). It uses wpa_cli i think, it's also unfinished and outdated and based mostly on fatdog's net-setup

All sorts of anomalies happen in a puppy distro, so i guess the code base and functions should be from the current netwiz/SNS or something

PR #1285, opened 2018-11-15
SNS: improve automatic connection speed when many wifi networks present

rserwin1:
Module loading determination fixed to handle absent drivers or hardware.
Boot-up wifi network detection improved for when many networks are available.

### DIFF
--- a/woof-code/rootfs-packages/simple_network_setup/etc/udev/rules.d/51-simple_network_setup.rules
+++ b/woof-code/rootfs-packages/simple_network_setup/etc/udev/rules.d/51-simple_network_setup.rules
@@ -1,0 +1,4 @@
+# Accumulate list of modules to be loaded, for use by rc.network.
+
+ENV{MODALIAS}=="?*", ACTION=="add", SUBSYSTEM=="?*", \
+  RUN+="/usr/local/simple_network_setup/build_udevmodulelist"
--- a/woof-code/rootfs-packages/simple_network_setup/pet.specs
+++ b/woof-code/rootfs-packages/simple_network_setup/pet.specs
@@ -1,1 +1,1 @@
-simple_network_setup-2.0|simple_network_setup|2.0||Network|120K||simple_network_setup-2.0.pet|+gtkdialog|sns as package||||
+simple_network_setup-2.1|simple_network_setup|2.1||Network|136K||simple_network_setup-2.1.pet|+gtkdialog|sns as package||||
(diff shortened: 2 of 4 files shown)


woof-code/rootfs-packages/simple_network_setup/etc/udev/rules.d/51-simple_network_setup.rules
```diff
@@ -0,0 +1,4 @@
+# Accumulate list of modules to be loaded, for use by rc.network.
+
+ENV{MODALIAS}=="?*", ACTION=="add", SUBSYSTEM=="?*", \
+  RUN+="/usr/local/simple_network_setup/build_udevmodulelist"
```
wdlkmpx: I think this will run even when SNS is not the active network manager

rserwin1:
That is a good catch.  Although adding "PROGRAM="grep '=sns' 
/root/.connectwizardrc"," is tempting, the saved information would be 
needed if the user starts SNS during the session, because it invokes 
rc.network to make a connection.

But rc.network could test for the 'start' argument, because the 
network_default_connect script does not pass an argument.  So, with 
"start", it could ignore the "initiated" indications and just check for 
what is already loaded.

Please wait until I can commit a fix.

wdlkmpx:
I get the impression it will run even when no network manager is being used.

As you can see in https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-code/rootfs-skeleton/etc/udev/rules.d/50-udev-puppy-basic.rules#L25 ... it's assumed that frontend_change is so standard that is hardcoded and it will always run even when there is no desktop, or someone is using a partial alternative (pup-volume-monitor). That's the scope of that decision.

But frontend_change detects when X is not running and frontend_funcs (rox-specific) will refuse to perform tasks if ROX-FIler is not running, but the 'overhead' is there..

rserwin1:
The SNS changes do not rely on backend_modprobe, but use its output only 
if present.  SNS 2.1 uses similar processing of aliases but does not do 
any module substitution.  So, it is ready for the absence of 
backend_modprobe and the
ACTION=="add", SUBSYSTEM=="?*", ENV{MODALIAS}=="?*", 
RUN+="/sbin/pup_event_backend_modprobe"
rule.

I am changing the 51 rule and rc.network to avoid waiting for modules if 
started from the sns script.  OK?

wdlkmpx:
Ok, but I have a question, some weeks ago, in a pull request, you
posted a link to a forum thread where someone said that all network
managers failed except peasywifi.. does peasywifi is as slow as SNS

I'm saying this because you might want to inspect peasy wifi and "port
the logic" into sns, netwiz, frisbee...

Or maybe it's time for simple network wizard...

wdlkmpx:
Hmm these are the relevant messages, i see you already did what i suggested

http://murga-linux.com/puppy/viewtopic.php?p=997101#997101

I'm ok with the changes you plan to make then

rserwin1:
Changes committed, ready for merge consideration.

PR #1288, opened 2018-11-20
pdiag: Add current version as a package

rserwin1:


ninaholic:
@rsewin1 should all the pdiag pets be removed from DISTRO_SPECS then so it will use your new one?

example:
https://github.com/puppylinux-woof-CE/woof-CE/blob/262bb31def8b33f61929dc58b22d07805ddb2d3f/woof-distro/x86_64/ubuntu/xenial64/DISTRO_PKGS_SPECS-ubuntu-xenial#L546

wdlkmpx:
See commit https://github.com/puppylinux-woof-CE/woof-CE/commit/7447c5227198e67e7667a4c21aa772ba7fb0ffcf  

I aslo added many other changes, i did test them all, but who knows what may happen. I'm in a hurry

@ninaholic might want to build a slacko 14.0 iso+devx with this kernel
https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-distro/x86/slackware/14.0/_00build.conf#L40

for rerwin..

PR #1291, opened 2018-11-28
Packages-puppy-*-official: add pdiag pet_specs for xenial(64), upupbb

rserwin1:


wdlkmpx:
pdiag is a noarch package. the pet_specs is in Packages-puppy-noarch-official

For my hack to work as expected (rootfs-packages override pet pkgs), there must a pet pkg  and a rootfs-package with the same name and the pet_specs. The hack happens in 2createpackages

this
pdiag-20171205|pdiag|20171205||Utility|44K||pdiag-20171205.pet||Consolidated diagnostic data collector||||

should be in Packages-puppy-noarch-official, replacing the current entry

It's confusing, i guess that's why it's easier for me to put stuff in rootfs-skeleton and delete all references from everywhere else ..

rserwin1:
I thought I was doing as you say. Peebee loaded the pet into the noarch 
directory at ibiblio.  I committed the pet.specs to github woofCE.  What 
do you intend when you say "there must a pet pkg"?  Where should it be 
contributed, if peebee's uploading of it is insufficient?  Where should 
the separate pet.specs be contributed?

The copy of pdiag is already in root-fs and contains its pet.specs 
file.  So, I am confused as to the right way to get a package into a 
repo, with or without having it in root-fs.

Will you be taking care of whatever else needs to be done to make pdiag 
available outside of woofCE? Or do you prefer to keep it only in 
woofCE?  Either is fine, with me.  My main concern is that if distro 
developers include pdiag, they get the latest version.

wdlkmpx wrote:

rserwin1:
OK, I see that I put the pet.specs in architecture- and puppy-specific 
directories and that it should not be there.  Thanks for correcting that.


wdlkmpx wrote:

rserwin1:
Thank you for closing this PR.  I see that I confused some instructions 
from Peebee.  I should not have made the changes for pdiag.

The instructions are for my desire to add my new redshiftgui pet package 
to the repo, without submitting it for inclusion in woofCE.  Peebee has 
already uploaded it to pet_packages-xenial, -xenial64 and -upupbb.  He 
told me to add the pet.spec to Packages-puppy-xenial-official 
<http://distro.ibiblio.org/puppylinux/Packages-puppy-xenial-official> 
and the same for xenial64 and upupbb.  I just had pdiag on my mind and 
applied the instructions to it.

Should I now submit commits for redshiftgui similar to what I did for 
pdiag?  Or is there a better way to add the pet.specs?

wdlkmpx wrote:

PR #1293, opened 2018-12-02
redshiftgui: Add pet.specs to Package...official repo lists for xenial, xenial…

rserwin1:
…64, upupbb

Pet packages already uploaded to the repos.


wdlkmpx:
How can i explain this..

redshiftgui is a noarch package. the only file to edit is Packages-puppy-noarch-official.

Packages-puppy-noarch-official is used by all builds

I also think that noarch pet pkgs should be in a git repo, it gives a sense oif community..

To make a pet pkg a default pkg there's  something to add to the PKGS_SPECS file of all distro compat versions

`yes|redshiftgui|exe,dev,doc,nls `

Editing the PKGS_SPECS can be seen as in instrusion, i guess that's why micko and phil no longer edit their stuff? Or is it because micko is busy and phil doesn't know how to use git. This makes sense..

I already performed a massive cleanup and added entries and stuff to all woofce PKGS_SPECS, otherwise how am i supposed to fix bugs that have been there for years? and i'm kind of tired ..

peabee:
redshiftgui .pet when opened has an 800K executable inside - it therefore probably isn't noarch....perhaps "gui" in the name is misleading....

rserwin1:
Also, it may be limited to tahr, xenial and upupbb because it contains a 
back-level (compiled executable) libcurl.so file that may not work in 
later puppies.  So, I think it should be confined to particular pups.  
Or maybe it should be in "common", if we don't worry about other pups.

PB wrote:

PR #1299, opened 2018-12-26
RIGHTVER is wrong

peabee:
Replace with DISTRO_VERSION


ninaholic:
I see it too

screenshot

PR #1320, opened 2019-02-03
suggested clean up by wdlkmpx & xvkb help link fix for PKGS_HOMEPAGES

mrfricks:
there are a few problems with this clean up but i'll open a new issue


01micko:
> there are a few problems with this clean up but i'll open a new issue 

What are the problems?

Can you sort them or need help?

mrfricks:
> > there are a few problems with this clean up but i'll open a new issue
> 
> What are the problems?
> 
> Can you sort them or need help?

hi 01micko,
good to see you!
the issue is with the suggested cleanup to my _00build.conf which i think you have commented on before.
see https://github.com/puppylinux-woof-CE/woof-CE/issues/1321
i'm ok with it as i'm aware of the problem, but if it does get pulled before being fixed, others won't get working builds.

BTW good to see the blog back up! (only just seen)

01micko:
Well @wdlkmpx (#1321) is right that the symlinks should go but not even that would fix the issue. Naturally, you are too close to release for this to be practical anyway.

IMHO, these problems (using  a 'common' repo) open a can of worms with different directory trees. To have a 'common' repo then it must cater to the different trees, such as setting a different `libdir`, for argument's sake `/usr/lib-temp` and then in the `pinstall.sh` shifting the libraries (or whatever) to the appropriate `libdir` (be it `/usr/lib/x86_64-linux-gnu`, `/usr/lib64` or whatever).

Probably the best short term solution is to repackage the offending packages and place them in the puppy-bionic repo.

Anyone else with a good short term solution?

wdlkmpx:
I use a 32bit system with wine apps, but lately i was actually using a
64bit windows.. i guess that's why this hasn't been fixed yet

I'm not sure, i saw too many fixes in bionic, and some of them were already
applied "upstream", making it unreliable.

At least /lib64 should be in path.. but i think i haven't tested a
bionicdog or ubuntu bionic yet.

But checking the rootfs in bionicdog (2 months ago) i didn't see that, only
x86_64-linux-gnu in /etc/ld.so.conf.d

To be able fix this you must understand how ubuntu works, and that symlink
in /lib64 is the key, other than that, having libs in /lib64 and /usr/lib64
sholdn't be a problem.

I think i should create a 64bit dpup to fix this, but i'm not sure if time
will allow that, should i get out of bed and do something productive in my
life or should i stay, that's the question

wdlkmpx:
In https://github.com/puppylinux-woof-CE/woof-CE/pull/1326 i explained a possible fix for the /lib64 /usr/lib64 issue

But the problem with the upups is that pet packages must be edited, stuff removed, pet packages reuploaded with a different suffix -1 -2 -3.

So it's really not easy to fix.

I'll fix that issue, but maybe tomorrow, i'll also add some 20+ commits including the change that adds WOOF_VER to distro_specs and the other change that makes the init script refuse to upgrade if WOOF_VER differs, that way development won't so stressing for the people of the world

peabee:
> i'll also add some 20+ commits including the change that adds WOOF_VER to distro_specs and the other change that makes the init script refuse to upgrade if WOOF_VER differs, that way development won't so stressing for the people of the world

The above change as currently implemented is draconian and not in the spirit of Puppy Linux. Refusing to load somebody's save folder/file that they may have spent a long time configuring is not very user friendly. Init should warn then allow the user to ignore the warning if they so desire and should provide helpful advice on how to modify their pupsave after making a backup to allow the boot to proceed with both a new puppy version and their old pupsave.

wdlkmpx:
peebee you obviously don't understand the situation. you haven't added
many hacks to ensure previous savefiles survive the transition.. i
removed some of those hacks.. now upgrades are way more dangerous.

Adding changes is difficult enough, adding hack after hack thinking
about previous pupsaves is act of masochism. You must hate yourself to
do that.

WOOF_VERSION will only change when something big happens, it can be a
simple change like getting of rid of the multiarch symlink.

My approach is different, all the woofce "pups" are evolving at the
same rate with the same changes... you don't understand that simple
fact.

Now real changes start to happen, but there are people trying to
prevent those changes from happening (i.e. peebee). There is some
shady agenda that i don't understand.

I will not write text about helpful advice, but it's possible to offer
an option to proceed with the upgrade.

I wrote this while working on something unrelated to puppy, from
Windows (tm). You better think twice what you say when it comes to
development, i'm busy enough.

peabee:
>  but it's possible to offer an option to proceed with the upgrade. 

Yes please!

PR #1324, opened 2019-02-08
slacko 14.0 - remove some obsolete patches

ninaholic:
these ones don't seem to be needed any more

screenshot


01micko:
Seems simple enough and doesn't conflict with imminent upup release at all.

PR #1326, opened 2019-02-12
Packages-common64: Remove pets that contain libraries

woodenshoe-wi:
The apulse pet also contains libs, but they are not in a standard
path and should work whether there is a lib64 or not.

The original idea was that common64 should only have pets that
didn't have libraries, but these slipped in.


woodenshoe-wi:
Unless someone comes up with a proper way to handle common64 pets with libs they should not be allowed in the common64 package list.

wdlkmpx:
Currently /lib64 and /usr/lib64 are symlinks, but in a "normal" debian/ubuntu derivatives they are actual directories

Unlike the x86_64-linux-gnu symlinks (existing installations, pet/sfs packages with the symlink), fixing this issue seems easy.

/lib64 is basically empty, but it has a required symlink to /lib/x86_64-....../ld.so.xx, and that symlink is not present in /lib ... what does this mean? /lib64 should be in $LD_LIBRARY_PATH (somehow).

It might be a good a idea to add something like this

echo '
/lib/64
/usr/lib64
' > rootfs-complete/etc/ld.so.conf.d/lib64.conf

if building a x86_64 iso

woodenshoe-wi:
The petbuilds are not consistent.

e2fsprogs, gtk-server and usb-modeswitch use lib,
while f2fs-tools and sudo use lib64.

I wouldn't mind recompiling the five offending pets in both slacko 14.0 and tahr, and then adding them to the slacko6414 and trusty64 package lists, but I don't want to be using separate petbuilds.  The same petbuild would have to be able to make pets with libraries in lib/x86_64-linux-gnu when run from tahr, and put the libraries in lib64 when run from slacko.

If you think it would be easier to fix things in woof to allow pets with libraries in lib64 to work in debian/ubuntu based builds, I would still like the petbuilds to consistently put the libraries in lib64 instead of the current mix.

wdlkmpx:
Well the issue is fixed in bionicpup64, but who knows what may happen in xenialpup64 or tahrpup64 

Yes common pet pkgs should be fixed, recompiled and reuploaded.

Just a few apps are used, but the common* apps become more relevant for older pups

All generic apps should be compiled for when the need arises. Although some apps could be static, busybox, ntfs-3g

PR #1332, opened 2019-02-20
Network connect: SNS 2.2, Network Wizard 2.1 & PPM bug fixes and refinements

rserwin1:
Highlights:
 - Corrects SNS failure with SSIDs longer than 20 characters.
 - Moves link to sns from /usr/sbin to /usr/local/bin, consistent with Network Wizard and frisbee.
 - Clarifies SNS 'make default' message -- default no longer affects automatic connection. 
 - In NW, handling of multiple ethernet interfaces improved, respecting selected wired interface.  Wired connection prevents starting of wireless; wireless connection allows all connections to start (unchanged).
 - NW now respects predictable interface names.
 - Allows abort (X) of SNS & NW without changing the "current manager" (for reboot).
 - Adds or corrects test for (SNS, NW, PPM) tool already running.


woof-code/rootfs-skeleton/usr/local/petget/pkg_chooser.sh
```diff
@@ -5,14 +5,14 @@
...
-VERSION=2.1.1
+VERSION=2.1.2
 
 export TEXTDOMAIN=petget___pkg_chooser.sh
 export OUTPUT_CHARSET=UTF-8
 
 # Do not allow another instance
 wait
-if ps --no-headers -C pkg_chooser.sh,ppm | grep -qwvE "^$$|^ $$"; then #v2.1.1...
+if ps --no-headers -C pkg_chooser.sh,ppm | grep -qwv "$$"; then #v2.1.2...
```
wdlkmpx: Have you tested this in dpup and bionicpup64?

rserwin1:
Yes.  The problem was caused by the assumption that the process IDs 
would be only 4 or 5 digits right-justified in the first output column.  
That excludes smaller numbers.  At startup, some process IDs could be 
less than 1000. The fix includes all numbers above 99.


wdlkmpx wrote:

PR #1333, opened 2019-02-21
remove support for PUPMODES 6 7

### LINKED ISSUE #1331: sfs_load and PUPMODE 7
I was reading sfs_load on my phone and PUPMODE 7 is not supported
`PUPMODE7SUPPORT="false" # PUPMODE=7 is buggy`

I think i never used a savepartition, and with savefolders a savepartition becomes irrelevant, unless luks encryption support for savepartitions is added, something that yours truly will never implement

Does anybody use savepartitions? would it be ok to remove support for savepartitions?

wdlkmpx:
This completely removes support for savepartitions. However init can identify savepartitions
and will show a "devastating" info msg (see commit diff)

The fact that savepartitions are only supported in PDEV1 makes them even less appealing.,,

This also shows basically all the core scripts that deal with pupmodes, so gyrog might want to take a look

patch
[remove-support-for-pupmodes-6-7.patch.zip](https://github.com/puppylinux-woof-CE/woof-CE/files/2887690/remove-support-for-pupmodes-6-7.patch.zip)

ref https://github.com/puppylinux-woof-CE/woof-CE/issues/1331#issuecomment-465622873
closes #1331

does anybody object this change?

### DIFF
--- a/initrd-progs/0initrd/init
+++ b/initrd-progs/0initrd/init
@@ -8,8 +8,6 @@
 #Aufs layers setup by this script...
 #aufs layers:            RW (top)      RO1             RO2              PUPMODE
 #First boot (or pfix=ram):  tmpfs                         pup_xxx.sfs      5
-#pup_save is a partition:   PDEV1                         pup_xxx.sfs      6
-#ditto, but flash drive:    tmpfs         PDEV1           pup_xxx.sfs      7
 #Normal running puppy:      pup_save.3fs                  pup_xxx.sfs      12
 #ditto, but flash drive:    tmpfs         pup_save.3fs    pup_xxx.sfs      13
 #Multisession cd/dvd:       tmpfs         folders         pup_xxx.sfs      77
@@ -599,13 +597,6 @@ setup_save_file(){
  fi
 }
 
-setup_save_part() {
- rm -r -f $SAVE_LAYER
- ln -sv ${SAVE_MP} $SAVE_LAYER
- SAVE_FN="${SAVE_MP}" #for after switch symbolic link
- SAVE_NAME="$SAVEPART"
-}
-
 copy_folders() { #SAVE_LAYER=/pup_ro1
  echo -e "\n-------------------- copy_folders" #debug
  echo "SAVE_MP=$SAVE_MP"
@@ -1142,8 +1133,13 @@ if [ "${SAVE_MP}" != "" -a "$PRAMONLY" != "yes" ];then #have mounted save? parti
  if [ "$ONE_FS_IS_LINUX" = "yes" -a "$PSUBDIR" = "" -a -f "${SAVE_MP}/etc/rc.d/PUPSTATE" ];then
   # make sure it's not a full install
   if ! grep -q 'PUPMODE=2' ${SAVE_MP}/etc/rc.d/PUPSTATE ; then
-    PUPMODE=6
-    PUPSAVE="${SAVEPART},${SAVE_FS},/"
+    # PUPMODE=6 PUPMODE=7 - unsupported
+    dialog --backtitle "$DISTRODESC" --title "'Savepartition': unsupported - ${SAVE_MP}" \
+      --msgbox "A savepartition is not supported: it will be ignored.
+
+However you can save session again using a savefolder - which is basically the same thing, except that it does not 'pollute' the boot partition.
+
+And then copy all the relevant files from ${SAVE_MP} tp the new pupsave (${DISTRO_FILE_PREFIX}save)" 0 0 &>/dev/console
   fi
  fi
  #not sorted yet, may be pupmode=12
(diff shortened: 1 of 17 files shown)


jamesbond3142:
Just a comment - since you already have a working "savedirectory" option, the savepartition could be implemented as "savedirectory" where the path is "/" ... (that's how we do it in Fatdog).

wdlkmpx:
If it's implemented as a savefolder then it has be PUPMODE 12/13 with some quirks, that means some code has to be restored. 

So it would be basically a savefolder without PSUBDIR, i'm not sure how it will behave, but i'll experiment a bit.

Savepartitions in puppy were basically old code with no recent improvements, and i guess fatdog supports encryption of savepartitions and stuff, which does make it a bit better, but overall savepartitions with no encryption are a "bad" frugal install. Just speculating..

wdlkmpx:
I got something working with some extremely minor changes in init 144447e0c4ec4907050e9bdb697edb47500d1ff2

Well, it's pupmode 12, a savefolder, all apps recognize it as a savefolder, but savefolders and savefiles require a filename, which is empty in this case.

It's not a problem i guess, but pupsave-backup does require a subdirectory or a file.. that script is patched now.

I'm not an expert in aufs, but it seems to work fine... but savepartitions have to be created manually.. move the puppy files to the partition's root directory, `mkdir -p etc/rc.d` , `touch etc/rc.d/PUPSTATE` ..  and add something like this to grub4dos:

title boot 
  kernel /vmlinuz
  initrd /initrd.gz

Screenshot to show how a savepartition doesn't look that good:
screenshot

Let's say it's a hidden gift..

jamesbond3142:
It is a bonus, like you say.

To clarify, in Fatdog, we don't support encryption in savedirectory (in / or in /some/subdir). But we do support accessing encrypted partitions, so if you're really inclined you can use a unencrypted savedirectory in an encrypted partition, to more or less achieve the same effect.

wdlkmpx:
Mounting (encrypted partitions (in init) is certainly a good feature. Will be implemented in the future.

This change is relatively trivial, and i'll build a new iso today, so this is being merged...

These are the facts:

1) savepartitions are savefolders that use the whole partition instead of a subdirectory
   the new approach might need some fixes.. needs some extensive testing

2) "savepartitions" are supported as savefolders or savedirectories but shutdownconfig does not have an option  to use a whole partition as the savefolder

3) People already using a savepartition will be able to upgrade..

wdlkmpx:
Well, my income doesn't come from programming and i do have a million
things to do, even my room looks like a war zone, my free time can be
better spent certainly. But this donation stuff is so alien to me, and
being the foe number 1 that constantly breaks stuff that has worked
well for years, the thought of it even seems crazy.

fixmenus: i think peebee also uses 16 instead of 22, it seds a file or
something. I see both use the same style for splashes, which is not
the default.. those different settings will become the default ones..

Back to bionicpup, I read people say /usr/lib64 is a symlink, but in
my builds, which is the default build i see

puppy_bionicpup64_7.9.8.sfs/usr/lib64: directory

I fixed it days ago, and i  that's why i closed the issue about /usr/lib64

I don't know what to do or what to expect.

wdlkmpx:
Oops sorry, i posted a message to the wrong issue. Click on the wrong
gmail link.. will delete it

gyrog:
On savepartitions:
I would simply abandon them.

On encryption:
I get that some folk have some files that they want ot keep private, and hence encrypted.
But, I don't get why anyone would want to encrypt files that are public domain,
e.g. Puppy sfs files, software installed via PPM etc.
So why install Puppy on an encrypted device?
Why store a savedirectory on an encrypted device?
And hence no need for mounting encrypted partitions in init.
Better to provide access to an encrypted directory (Could be the mount point of a  encrypted partition), within the running system.
This would contain only files that the uer wants to remain secret.

Note: My current "overlay_init" does support mounting luks partitions in "init", but I've abandonded this idea in the "init" I'm currently working on.

wdlkmpx:
Well. savepartitions are not supported, they are savefolders that can't be created with the puppy tools. init recognizes savepartitions but loads them as savefolders instead of showing a message that they're not supported, this didn't require any special changes, only a few lines instead of the info msg.

Hmmmmm.... maybe your method to encrypt only a specific directory no matter the type of pupsave could be integrated and could eventually replace savefile encryption, but i guess this will also trigger a chain reaction.

Init becomes a bit paranoid when it finds a savefile and you don't provide the correct password, it doesn't let you continue in pupmode 5. If you have only one savefile, that indeed blocks bootup.

Some stranger might think the pc is locked, and the user itself might find himself locked out.. it once happen to me.. only one entry in bootloader and only savefile.

gyrog:
My idea for encryption is to create a "PupVault" file based on the current "luks savefile" code,
except it's not mounted in "init" but in "$HOME/Startup" using a gui to get the password.
Theoreticaly, such a file could be stored anywhere; in a savefile, in a savefolder, or outside the stack on any mounted partition.
Also it should not have shutdown problems since the required rw loop device is not a layer in the / stack, and should be able to be really umounted by rc.shutdown.
And of course there would be a utility to create and maintain the "PupVault".

wdlkmpx:
Well, the "savefile masacre" of rejected savefiles [when people don't want
to upgrade] happened because /pup_rw cannot be unmounted after it's added
to the aufs stack, but it can be unmounted before that.

Something that could be done her3 and now.... perhaps init should unmount
it anyway, and rc.sysinit should mount it again add the savefile to the
stack and remount it, that may fix the eternal savefile issue.

So you're saying you want something else on top of that or am 8 misreading?

PR #1340, opened 2019-02-27
release recipe for bionicpup64

mrfricks:


mrfricks:
ooops left in the lib64 symlink fix in _00build.conf.  i'll update that file once this goes through

wdlkmpx:
I see f2fs-tools, i guess this is to avoid breaking the /usr/lib64 symlink or something.

As i said, i have no idea what to expect. I haven't downloaded 8.0

But, for your information, /usr/lib64 is currently properly supported. ref: https://github.com/puppylinux-woof-CE/woof-CE/commit/107a3811969437b8d6fc09e2cacbb8cab3cd6c9f

PR #1368, opened 2019-04-22
Update cache files upon removal of builtin package

rizalmart:
Need to apply this routine in order to avoid troubles upon uninstallation of builtin packages

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/remove_builtin
+++ b/woof-code/rootfs-skeleton/usr/sbin/remove_builtin
@@ -56,6 +56,60 @@ remove_builtin_pkg() {
 		done
 	done
 	#-- 
+
+PKGFILES="$D/$PKG"
+
+if [ "`grep '/usr/share/glib-2.0/schemas' $PKGFILES`" != "" ];then
+ [ -e /usr/bin/glib-compile-schemas ] && glib-compile-schemas /usr/share/glib-2.0/schemas
+fi
+
+if [ "`grep '/usr/lib/gio/modules' $PKGFILES`" != "" ];then
+ [ -e /usr/bin/gio-querymodules ] && gio-querymodules /usr/lib/gio/modules
+fi
+
+if [ "`grep ''\.desktop$'' $PKGFILES`" != "" ];then
+ rm -f /usr/share/applications/mimeinfo.cache
+ [ -e /usr/bin/update-desktop-database ] && update-desktop-database /usr/share/applications
+fi
+
+if [ "`grep '/usr/share/mime' $PKGFILES`" != "" ];then
+ [ -e /usr/bin/update-mime-database ] && update-mime-database /usr/share/mime
+fi
+
+if [ "`grep '/usr/share/icons/hicolor' $PKGFILES`" != "" ];then
+ [ -e /usr/bin/gtk-update-icon-cache ] && gtk-update-icon-cache /usr/share/icons/hicolor
+fi
+
+if [ "`grep '/usr/lib/gdk-pixbuf' $PKGFILES`" != "" ];then
+ gdk-pixbuf-query-loaders --update-cache
+fi
+
+if [ "`grep '/usr/lib/gconv' $PKGFILES`" != "" ];then
+ iconvconfig
+fi
+
+if [ "`grep '/usr/lib/pango' $PKGFILES`" != "" ];then
+ pango-querymodules --update-cache
+fi
+
+for gtkver in '1.0' '2.0' '3.0'
+do
+ if [ "`grep '/usr/lib/gtk-$gtkver' $PKGFILES | grep "/immodules"`" != "" ];then
+  [ -e /usr/bin/gtk-query-immodules-$gtkver ] && gtk-query-immodules-$gtkver --update-cache
+ fi
+done
+
+if [ "`grep '/usr/share/fonts' $PKGFILES`" != "" ];then
+ fc-cache -f
+fi
+
+KERNVER="$(uname -r)"
+
+if [ "`grep "/lib/modules/$KERNVER" $PKGFILES`" != "" ];then
+ depmod -a
+fi
+	
+	
 	rm $D/$PKG
 	rm /tmp/remove_builtin_dirs$$
 	sed -i "\%|${PKG}|%d" /root/.packages/*-installed-packages


wdlkmpx:
Now we see why some people unified the ppm into a single big script, that way there's no need to place the same code in several scripts.

By the way, this script may also change.. in the experimental branch. 

I just noticed peebee released isos that include the changes in the 'experimental' branch -but not the latest changes- in the upupdd thread, maybe he should add lxpup upupdd so you can feel comfortable testing it, although rox has become quite usable.

To speed up processing, 2createpackages & 3builddistro should become 1.

And 2createpackages should behave like installpkg.sh, it should get the file list using `tar -taf`, `dpkg-deb --contents`, and that should be the filelist in builtin_files/package... i guess the current code does not handle those filelists properly.

Then it should extract the pkg contents probably directly to rootfs-complete, after all it's done, it should move stuff to _doc, _nls, devx, adrv, ydrv, etc..

Looks like a lot of work, and now that i use basically wine apps, i guess i only need a working puppy and then wine does the rest, i kind of lost the motivation to undertake such a task..

peabee:
> To speed up processing, 2createpackages & 3builddistro should become 1

PLEASE don't do this.....

wdlkmpx:
On Wednesday, April 24, 2019, PB <notifications@github.com> wrote:
> To speed up processing, 2createpackages & 3builddistro should become 1
>
> PLEASE don't do this.....
>
>

And the reason is...

So far i only see you want to avoid building a iso file from scratch, and
as a result some builds are broken . I build with a pentium 4 pc and i
don't see the problem, it gets really fast with faster hardware.

Reading the code i see the build scripts are quite disjointed and because
of this the overall process is slower, but it could be faster than already
is in the experimental branch.

I don't know how others build, but i don't add files, i don't post process
anything, it's a straightforward process.

peabee:
> And the reason is...

If woof-ce is the same version used in the last build
and
the only components that have changed are from the compat distro
then
there is no need to do a full 2createpackages (which even though you have made it faster is still quite a lengthy process)

so I don't want to be forced into doing a lengthy process that isn't needed.

Please leave things as they are!

I promise to do a full 2createpackages whenever it is actually needed in future....

wdlkmpx:
@peabee it could be easier if you just merge your stuff into the standard build for x86 distros. without gvfs of course. So there's no need to add custom stuff, a default build would have everything.

I also think that ubuntu bionic should be the last 32bit ubuntu pup. Hell i think bionic should be the last ubuntu pup. You should try with debian instead.

I noticed debiandog comes with 2 DE's, one to appeal to classic puppy users and another for everyone else. (default) Dpup stretch partially does that, except that it doesn't include openbox, it uses rox or pcmanfm with jwm, but there's no obvious way to change the desktop app. But 2 clearly distinct DE's are needed. . And your lxpups provide that, but without gvfs, without gvfs is the key.

it's really difficult when you isolate yourself and produce builds the way you do. Specially ubuntu pup builds. You should be releasing official puppies, since you really update your isos when you have to. The project needs more peebees, but releasing official stuff, and updating that official stuff when the time is right. 

At least all the slackos and all the dpups could feature that new approach.

It's a pity how everything is wrong.

Well, what you're trying to prevent is actually the only way to support any compat distro, and it certainly would be at least as fast as the current 3builddistro, but including 2createpackages in the process. But of course i don't  think i have enough time to do that. But your thoughts, your thoughts.. 

If you disable the option to run the `strip` command, it's fast. I build dpup isos really fast, it gets slow when it processes the devx stuff, but with the new approach that part will be faster.

I build isos with a slow machine, in dpup, which is configured to build as fast as possible, things slow down only when processing the devx stuff and compressing sfs's. everything else is fast.

I wish i could explain many things.

I read your mind, with no intentions
Of being unkind, I wish I could explain
It all takes time, a whole lot of patience
If it's a crime, how come I feel no pain.

jamesbond3142:
> I read your mind, with no intentions
> Of being unkind, I wish I could explain
> It all takes time, a whole lot of patience
> If it's a crime, how come I feel no pain.

You know, some of your exotic poetry should into Woof-CE. Some sort of easter eggs. You have too many of them, it is a pity to see them wasted in this place and never to be seen again. 

Firefox has the elusive book of Mozilla, perhaps we can have the revelation from "K9 from outer space" ...  :smile:

wdlkmpx:
It's not my poetry, i listened to many songs in english, when i was a kid and a teenager, before i could properly understand the words. Many of the things i say are borrowed from someone or something. Like a parrot, although I never speak in english. I just write.

"I wish i could explain" is from a song called "Carrie"
https://www.youtube.com/watch?v=bZoSGL96CzQ

Somehow the intro and the overall feel also reminds of "Wasurenaide (Don't Forget)"
https://www.animeigo.com/liner/anime/bubblegum-crisis-songs-1-3

Which is a sweet pop ballad, with a rocking background music, and heavy metal guitar solos. 

"When the time is right".. it's from a song called "When the time is right"
https://www.youtube.com/watch?v=k2C_WKbpp18

I listened to many good songs with poetry in several languages. Hmm i have a recording on my phone of me trying to sing "Fly me to the moon", i rarely sing and i can't reach high notes
[rec_2017-12-18_20-34-52.m4a.zip](https://github.com/puppylinux-woof-CE/woof-CE/files/3135296/rec_2017-12-18_20-34-52.m4a.zip)

In other wars.

Butterfly:
> > I read your mind, with no intentions
> > Of being unkind, I wish I could explain
> > It all takes time, a whole lot of patience
> > If it's a crime, how come I feel no pain.
> 
> You know, some of your exotic poetry could go into Woof-CE. Some sort of easter eggs. You have too many of them, it is a pity to see them wasted in this place and never to be seen again.
> 
> Firefox has the elusive book of Mozilla, perhaps we can have the revelation from "K9 from outer space" ...

*waving* (am still here :smile: ) That's a cool idea that could come about a couple different ways

One is something I've heard chatted up for Vinux (Linux for people with visual impairments, development is on unexpected "hiatus" right now). I've never installed Vinux, but they talked about having something that provided rotating user tips.

Was most likely a distribution-wide feature similar to how some singular software packages provide tips when they first open. Those tips always seem to have the *nice* option of being able to toggle them off via Preferences' entries.

Within Debian, `gparted` does something I really like. While `gparted` is running, it displays this in the terminal tab that's tied up with that activity:

```
======================
libparted : 3.2
======================

```

Number One, that helps me easily remember what program is tying up that terminal tab. Yay! That then remains in the terminal tab after `gparted` closes. 

A different program that I've unfortunately forgotten the name for also displays a quick k/t to its current developers (when run via terminal). I think that's also cool. I don't mind seeing the names of those who work hard to provide the packages I love.. :heart: 

And now more off-topic because it just occurred to me while running `gparted` to grab that snippet.. Wary5.5. is my very dependable (hsfmodem) dialup modem providing hero right now. It naturally runs as root so I've never encountered a terminal password prompt.

Does Puppy have a Puppy'esque cue at that point... or does it "just" ask for password? Am thinking to ask that because Debian-User listserv once had a tip that inspired me to change my terminal password prompt to read...

`♥*•♫♪ #GeekGirls Rock! ♪♫•*♥ 
`

That one's altered in `/etc/sudoers`. The line *I learned to add* and **DISCLAIMER** > that may or may not be totally copacetic is:

`Defaults passprompt_override,passprompt="♥*•♫♪ #GeekGirls Rock! ♪♫•*♥ "`

Mine's only useful to me unless someone else with Linux experience knows to expect a password prompt at that point. New users wouldn't grasp the fun and yet still purpose in having something like that right there at that point. They'd surely still need to see the word "password" in there somehow if that's changed. My apologies in advance if that's something that's already Puppy'ized. :dog: :dog2: 

PS The Woof-CE subject line that contained the word "LICKS" the other day? I haven't had a chance to open and comprehend it yet, but I"m STILL laughing! That's a good one!

peabee:
@wdlkmpx - Very brave to share your little recording with the world..... ;-)
It's not clear to me what part of my "stuff" needs to merged into something else??
I certainly don't want to get rid of gvfs....or to stop doing 32-bit builds.....the only 64-bit builds I will do will be Slackware based....I leave it to others to grapple with debian/ubuntu 64-bit architecture....
Not aware that I am "isolating myself...."??
Compressing sfs is actually the quickest part of the process on my 8-cpu desktop as it uses all 8 cores whereas all the rest only uses 1 core leaving 7 twiddling their thumbs :-(

PR #1373, opened 2019-04-29
Add replace package support for PPM

rizalmart:
If there is an already installed package entry in the system database, just update the package files list and its database entry.

### DIFF
--- a/woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
+++ b/woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
@@ -759,7 +759,30 @@ if [ "$DESKTOPFILE" != "" ];then
  fi
 fi
 
-echo "$DB_ENTRY" >> /root/.packages/user-installed-packages
+#If there is an already installed package, just update the package files list and its database entry
+xpkgname="$(echo "$DB_ENTRY" | cut -f 2 -d '|')"
+installed_pkg="$(cat /root/.packages/user-installed-packages | grep "|$xpkgname|")"
+
+if [ "$installed_pkg" != "" ]; then
+  installed_files="$(echo "$installed_pkg" | cut -f 1 -d '|')"
+  if [ "$installed_files" != "" ]; then
+    if [ -e /root/.packages/${installed_files}.files ]; then
+	while IFS= read -r xline
+	do
+  	 if [ "$(cat /root/.packages/${DLPKG_NAME}.files | grep "$xline")" == "" ]; then
+	  echo "$xline" >> /root/.packages/${DLPKG_NAME}.files
+	 fi
+	done < /root/.packages/${installed_files}.files
+	rm -f /root/.packages/${installed_files}.files
+    else
+     echo "$DB_ENTRY" >> /root/.packages/user-installed-packages
+    fi
+  else
+    echo "$DB_ENTRY" >> /root/.packages/user-installed-packages
+  fi
+else
+ echo "$DB_ENTRY" >> /root/.packages/user-installed-packages
+fi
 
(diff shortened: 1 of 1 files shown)


wdlkmpx:
To replace a package you must first uninstall the previous package. Delete all the listed files, run uninstall script, delete filelist, delete uninstall script, delete references in user-installed-zzz..

By uninstalling and installing again the process is reliable and faster, you don't run grep & cat a lot of times.

I never did this, because, well, i don't use the ppm. But in the experimental branch i have optimized some routines in the ppm, and deleted a lot of lines of course, as part of structural changes in the build system. I enjoy deleting lines.

I optimized the 'trim the fat' code, it would take minutes with packages with many files, but now it takes seconds. You might want to take a look.
https://github.com/puppylinux-woof-CE/woof-CE/blob/experimental/woof-code/rootfs-skeleton/usr/local/petget/installmodes.sh

But still it can be way faster by replacing the `grep`s with case `$ONEFILE in` see example in 
https://github.com/puppylinux-woof-CE/woof-CE/blob/experimental/woof-code/2createpackages#L426

My advice is that you look in the experimental branch to see the current state of the scripts
https://github.com/puppylinux-woof-CE/woof-CE/blob/experimental/woof-code/rootfs-skeleton/usr/local/petget

I'm not sure if it's ever going to be merged into testing though

ninaholic:
I have played with the Puppy Package Manager v2.5 and I like it better so far. It is simpler and fast and feels less buggy than earlier versions (no more two packages appearing on right side when you only want one for instance). For downloads I already know I usually switch it to Automatic mode, but Step-by-step is probably better default so you can know and control what's happening. I too like to just find the files on ibiblio or puppy forum and click on the pet of sfs manually instead usually. I run ppm sometimes when I'm lazy and in live mode so the changes don't matter. :laughing:

wdlkmpx:
Yes the PPM is less buggy and simpler in the experimental branch, although sometimes improvements have be reverted due to popular demand.

As far as i know installpkg.sh does not let you to replace or re-install a package, you have to delete some lines, i don't see that in this pull request.

It's possible to detect if a pkg is already installed, in that case installpkg.sh should uninstall the previous pkg first, before the new pkg is extracted or copied to /, then install it the usual way. 

So the code to add should be around lines 120-210

I explained how to do it in my first comment... i'm not exactly expecting rizalmart to do that, it would be the proper way to reinstall or replace a pkg though..

I haven't  tested this yet, ninaholic might to expend a couple minutes testing if this actually works... the file is here  https://raw.githubusercontent.com/rizalmart/woof-CE/patch-4/woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh

rizalmart:
I noticed that if a user try to install "package-1.0.pet" then suddenly install "package-1.0b.pet" it creates a duplicate entry in PPM registration it treated "package-1.0b" as a separate package although it was a same package. So this commit will just update the files list of an installed package and PPM database because for this example, files of package-1.0 was overwritten by package-1.0b much like replace package of android

wdlkmpx:
I think it's still  not right.

Sometimes you want to remove something from a package, plugins or something the breaks the app, the plugin or something will still be there after you install the revised .pet, you should uninstall the previous package.

What i see here is that you merge the contents of 2 different packages, well, theoretically different.

But i think i can merge this anyway, and i might fix the logic some day. It's not right.

I tested this after merging locally to testing and then to experimental, it works as adverstised, probably, i can't test using the file in the testing branch, installpkg.sh is not compatible.

wdlkmpx:
I created this page, just in case:
https://github.com/puppylinux-woof-CE/woof-CE/wiki/Building-using-the-'experimental'-branch

PR #1376, opened 2019-04-29
Update cache files during load and unload SFS modules

rizalmart:
Need cache file to be updated for more flawless integrations


woof-code/rootfs-skeleton/usr/sbin/sfs_load
```diff
@@ -424,8 +451,35 @@ afterwork() {
 		fi
 		fc-cache -f
 	fi
+	
+	if [ "$HAS_DESKTOPS" ]; then
+	 update-desktop-database /usr/share/applications
+	 update-desktop-database /usr/local/share/applications
```
wdlkmpx: should be [ -x /usr/bin/update-desktop-database ] && update..

PR #1377, opened 2019-04-30
woof-distro/arm/raspbian README.md

woodenshoe-wi:
How is this?

If you think I need to add some additional info, just give me a list of the topics and I’ll gladly write something up.


wdlkmpx:
I think it's ok for now

PR #1380, opened 2019-05-02
Improve underdog: hide some underdog init scripts and autostarts

rizalmart:
Need to hide some init scripts due to puppy's rc.services nature that executes scripts in init.d with executable permissions. Also some autostarts which was not compatible with puppy

### DIFF
--- a/initrd-progs/0initrd/init
+++ b/initrd-progs/0initrd/init
@@ -1066,9 +1066,32 @@ if [ "$UNDERDOG" ];then #boot parameter
  ensure_mounted "$UNDERDOG" "/pup_ud"
  if [ "$ONE_MP" ];then
   mount -o remount,append:${ONE_MP}=rr /pup_new
-  #if [ $? -eq 0 ];then
-   #do any necessary underdog fiddles
-  #fi
+  if [ $? -eq 0 ];then
+       #Hide /etc/xdg/autostart files let puppy run on top
+    if [ -e /pup_ud/etc/xdg/autostart ]; then
+      [ ! -e /pup_rw/etc/xdg/autostart ] && mkdir -p /pup_rw/etc/xdg/autostart
+	  for dfile in $(ls -1 /pup_ud/etc/xdg/autostart 2>/dev/null)
+	  do
+	   [ ! -e /pup_rw/etc/xdg/autostart/$dfile ] && touch /pup_rw/etc/xdg/autostart/.wh.$dfile
+	  done
+    fi
+    
+    #Hide /etc/init.d files let puppy run on top
+    if [ -e /pup_ud/etc/init.d ]; then
+      [ ! -e /pup_rw/etc/init.d ] && mkdir -p /pup_rw/etc/init.d
+      
+      touch /pup_rw/etc/init.d/.wh.rcS 2>/dev/null
+      touch /pup_rw/etc/init.d/.wh.rCS 2>/dev/null
+      touch /pup_rw/etc/init.d/.wh.rc 2>/dev/null
+      
+	  for dfile in $(ls -1 /pup_ud/etc/init.d 2>/dev/null)
+	  do
+	   if [ "$(echo "$dfile" | grep -E "httpd|lighttpd|nginx|mysql|postgresql|ziproxy|tinc|tinyproxy|tlp|transmission|VBoxService|wicd|vrrpd|privoxy|postfix|postgrey|vrrpd|smartd|squid|atd|apache|hiawatha|rsync|tazpanel|ftpd|timidity|lm-sensor|sane|thermald|cherokee|clamd|freeradius|nagios|x11vnc|vnstatd|unfsd|pptpd|pcscd|p910nd|vmtoolsd|ofono|nscd|ndo2db|nrpe|motion|miniupnpd|dnsmasq|dhid|cyrus|couchdb|chilli|connman|gpm|hostapd|hylafax|compcache|netdata|ntop|openldap|openssh|perdition|xrdp|ypserv|daemon")" == "" ]; then
+	    [ ! -e /pup_rw/etc/init.d/$dfile ] && touch /pup_rw/etc/init.d/.wh.$dfile
+	   fi
+	  done
+    fi
+  fi
  fi
 fi
 


wdlkmpx:
I'm afraid i'll never test this. When a puppy is "fairly" complete and/or when there are many choices to build from underdog becomes less interesting. But i'll merge this and i'll move that code to /sbin/setup_underdog where you can add more stuff if you wish

I'll also move the multisession cd stuff to /sbin as a helper script and also other code blocks, a more modular init where features are easily removed.

wdlkmpx:
Hmmm i don't recall ever testing or using underdog with a real distro, maybe using a puppy full install that i quickly deleted.. funny. I wonder what is the purpose exactly, to test another package manager, to experiment in a quirky way? Is it valid anymore?

In puppy there are many possibilities.. many kernels available and that can be compiled, etc. The only missing feature is the ability to use the default compat distro kernel in a frugal install.

I guess many puppy users, including those using wary, racy, puppy4, etc.. have come to the conclusion that they don't need anything else, the WINE pkg provided  by version2013 allows them to run many free and paid apps that are more than any distro can offer, and all of that using a tiny distro.

rizalmart:
I already tested that underdog script modifications using full install tazpuppy and frugal installed x-slacko slim. it works fine.

wdlkmpx:
Ok. It's my understanding that you have different stuff in your pups,
in some cases there might be incompatibilities

Specially when it comes to the ppm, what you see in the testing branch
is not the latest version

In the experimental branch another massive code deletion and
"rationalization" (in micko's words) is going on. After that
zwoof-next will be added as a proof of concept.

wdlkmpx:
Looking at the changes in init and rc.shutdown, so the underdog can be loaded and unloaded.

It seems to me that the underdog stuff is some sort of primitive chroot/container stuff, a very quirky and dangerous way to achieve what others have been doing with chroot/containers for ages.

Underdog is just too invasive and can easily lead to a broken system, it requires more improvements and stuff.

If you want to use the package manager or apps of some other installed distro, you normally do it through a chroot, chroot jail, container, something like that, don't know much about it.

Basically you mount stuff, mount --bind or something, maybe backup some files in the target system, copy /etc/resolv.conf  and maybe a few other files, and that's it you're ready to run apps. Then unmount stuff and restore a few files after the chroot ends..

Let's say there is already 1 underdog: the compat distro. With the underdog, there are 2 underdogs.

I hope you remove underdog from your buildsys and create something based on chroot or something. 

Having said, i think i'll delete the underdog stuff in the experimental branch, the current stuff in the testing branch will continue for a while, so you see it's there, but will be deleted somewhere else..

s243a:
Does this discussion includes mistfires modifications to underdog?

Anyway, if the underdog is the binary compatible version of linux (or a closely related version) then I expect there shouldn't be too many incompatibilities. This would allow the puppy stuff to run in ram and pull from the full version of linux when needed. 

disclaimer: I don't know how underdog works.

s243a:
mistfire wrote:
"I successfully make a breakthrough in underdog. It is now more usable and less troublesome than before. It is done by hiding the /etc/xdg/autostart files and some /etc/init.d scripts of underdog layer on top layer of aufs then update all the cache files. This method is now implemented on the woof-ce"
http://murga-linux.com/puppy/viewtopic.php?p=1027216&search_id=2125841891#1027216

wdlkmpx:
Underdog is some sort of chroot/container that is loaded in init as
the lower layer, in which you boot some other distro using a puppy,
and puppy takes over.

mistfire only tests in his puppies with small distros and with his
modifications, underdog is something huge, its implications are huge.
The tools that deal with remastering puppy will find trouble with a
loaded underdog and so on.

Answer to underdog: containers, chroot, etc. Problems: almost none.
Problems with underdog: many.

You can use any distro as the underdog and that distro may be highly
incompatible with the puppy you're using, i mean extremely
incompatible. Binary clashes, etc.

And aufs being just one of the possible unionfs, i guess underdog is
not properly supported by overlay or something, there are many things
that aufs can do that are not possible in overlay, with this a step to
unification is achieved..

wdlkmpx:
The underdog is the product of old concepts and limitations of puppy.

With woof2 many of the limitations were overcome.. by using another
distro as the puppy itself.

With Pkg another limitation is being overcome, now it's possible to
use cli pet pkg manager.

With wine it's possible to use many applications, that if you look
carefully, it will satisfy basically all your needs, so there's no
need for something as tricky as this.

And the asnwer to everything: portable packages.

Packages compiled in a fairly old distro with some static stuff and
with very basic dependencies that will run basically on most newer
distros.

I think someone produced something like that: a fairly static mpv, it
was compiled in precise, i read people uses is the slackos and all
newer ubuntu puppies, even in wary, racy (with glibc upgrade).

You see, there's no need for this..

wdlkmpx:
The key is to provide a way to run another distro without compromising
the pup system, and i think mistfire will surely implement something
like that very soon.

It's also worth mentioning again the importance of generic packages,
just like firefox, a very complex and modern app that runs in old
distros (as old as slackware 14.0, ubuntu precise).

Compiling generic packages is an art that is not hard to master. There
are few essential packages, a media player, a web browser, some other
essential apps, wine to fill the gaps. Isn't it funny, with wine you
run the latest version of some complex cross-platform qt5 app that you
can't compile in your old distro.

With underdog we're also dealing with legacy code.. reminds of a song
called "Prisoner of your lines", which has a sick and long nice guitar
solo

"Code is blind and
And code deceives you
You came along and captured me
Now i'm a prisoner of your lines

Trapped in time
I cannot leave this
I'm just a prisoner of these lines"

Which is the story of puppy. True story.

Now i'm on lighter note, have you ever noticed how cool the PS4 games look

I was watching some clips of God of War 4, it's almost a cinematic experience
https://youtu.be/8_CKqOuQvBk

At some point Baldur says: "Pain, it's so wonderful!". Indeed

"Scaterring Faye's ashes" is also worth watching, cinematic
https://youtu.be/VyzGBghdMDI

Mortal Kombat 11 also looks quite good and cinematic.

Now imagine spending your time having fun or having to work to earn a
living or being forced to raise your kids and then coming back to find
you don't understand the code you wrote long ago. That's why it's
important to delete code.

PR #1385, opened 2019-05-11
fix slackware64 puppy-specific repos

01micko:
Just fixing a minor omission


01micko:
James I'm merging this but is a solution to nothing as *lightofdawn.org* should not be the repo of a puppy build system. It is a stop gap until we get a repo organised.

jamesbond3142:
No problem Mick. When you have a better place i can send you the tarball
containing the stuff that needs to be uploaded. It's just a bunch of files
and folders; and there are only two packages there: xdialog and gtkdialog.

01micko:
Hi James,

There is a new repo created at ibiblio (and mirrors to follow)

http://distro.ibiblio.org/puppylinux/next/puppy-extra/

I also have slackware{64}-14.2 building and booting on VM and bare metal. I'll issue a PR when ready.

Send the tarball when you are ready.

PR #1393, opened 2019-05-29
Create repack_builtin

rizalmart:
This script repacks the builtin package back into pet package without deleting the builtin package


sc0ttj:
Just FYI, Pkg as a `pkg repack <package-name>` command that already does this, plus it can also repack any user installed packages as well.

s243a:
> Just FYI, Pkg as a `pkg repack <package-name>` command that already does this, plus it can also repack any user installed packages as well.

Thankyou for the tip scottj :)

PR #1395, opened 2019-05-29
Create repack-builtin.desktop

rizalmart:
Add menu entry for repacking builtin packages


peabee:
This seems like a very specialised function to have its own menu item cluttering up the menu.....puppy is often criticised in reviews for having too many items in its top level menus.....perhaps a different route to this function would be preferable (not sure what tho....part of the existing remaster?)

PR #1398, opened 2019-05-29
filemnt: Move image file mountpoint from /mnt to /media

rizalmart:
Moving mounted image from /mnt to /media will shows the mounted disk image on file managers like Thunar and PCManFM for easy unmounting. Also pup-volume-monitor supports detection of mountpoints in /media.

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/filemnt
+++ b/woof-code/rootfs-skeleton/usr/sbin/filemnt
@@ -137,7 +137,6 @@ if [ "$MNTEDLOOP" = "" ];then #not mounted on $MOUNTPOINT
          ;;
       img|hfv|vfd|dsk)
         mount -o loop "$imgFile" "$MntPt"
-        Err=$?
         ;;   
       *) 
          mount -t $Type -o loop,rw "$imgFile" "$MOUNTPOINT"
--- a/woof-code/rootfs-skeleton/usr/sbin/filemnt
+++ b/woof-code/rootfs-skeleton/usr/sbin/filemnt
@@ -35,9 +35,13 @@ if [ "$MNTEDLOOP" = "" ];then #not mounted on $MOUNTPOINT
     2fs) Type='ext2'     ; check_luks=yes ;;
     3fs) Type='ext3'     ; check_luks=yes ;;
     4fs) Type='ext4'     ; check_luks=yes ;;
-    sfs) Type='squashfs' ;;
+    sfs|lzm|xzm|sb|squashfs) Type='squashfs' ;;
     iso) Type='iso'      ;;
     wim) Type='wim'      ;;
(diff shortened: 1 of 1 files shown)


wdlkmpx:
I use pcmanfm+pup-volume-monitor.

I don't see the expected results...

You might want to share a 32bit binary of your pup-volume-monitor to see if it works..

rizalmart:
I think they remove that functionality on the source code. Here is the pup-volume-monitor that was used in X-Slacko Slim (Just remove .txt extension). Also detection of mounted images can also detected by  udisk2-volume-monitor

[pup-volume-monitor.tar.xz.txt](https://github.com/puppylinux-woof-CE/woof-CE/files/3234836/pup-volume-monitor.tar.xz.txt)

wdlkmpx:
It was simplified a bit, but no features were removed, only duplicate
code although there is a report that there is an issue in a recent 64
bit distro... because something is probably missing, an element or a
class that was defined before but it's no longer defined, if i trust
what my 7th sense tells me.

The binary you provided behaves exactly the same, what should i mount?
I mounted a savefile, a sfs file.

We're probably using completely different setups, i do use pcmanfm but
with jwm instead of lxpanel + openbox. I do use lxapps, but not the
whole environment.

Debian and slackware are also completely different.

rizalmart:
As long it was an image file format such as *.iso, *.sfs, *.img etc

rizalmart:
Here is the sample screenshot of Thunar running pup-volume-monitor that I sent to you
sample-screenshot

rizalmart:
Here is the sample screenshot of PCManFM running udisk2-volume-monitor (TazPuppy)
Screenshot

wdlkmpx:
Well the same binary doesn't add img files to mounted drives list in
the default build of dpup stretch. But i'll merge it anyway.

The other thing i noticed is pcmanfm's superiority in look and feel,
the shortcuts and functionality. When i tried thunar in lxslacko and
debiandog jessie i found i needed a mouse, and quickly dismissed both
isos, with pcmanfm i don't need a mouse.

And the good thing is, the latest version of pcmanfm and the whole
lxde compile in really old distros. But of course not all apps are
good, that's why you have gnome-mplayer, deadbeef, engrampa, gmrun,
geany.

peabee:
Problems:
http://murga-linux.com/puppy/viewtopic.php?p=1029738#1029738
http://murga-linux.com/puppy/viewtopic.php?p=1029739#1029739

rizalmart:
Peebee figure out the problem. There are some changes occured at libfm commited by lxde developers beyond version 1.2.5 . I think pup-volume-monitor must be updated

http://murga-linux.com/puppy/viewtopic.php?t=96875&start=345

PR #1401, opened 2019-05-29
Rewritten pupcamera

rizalmart:
Added multiple camera support (more than two cameras) even remote camera


wdlkmpx:
I tested this and it works, however the width was a bit small. 

Now the default width is 710

if the screen width is >= 720, dialog width = 600
if the screen width is >= 640, dialog width = 500

screenshot
Screenshot

wdlkmpx:
Oops i meant "screen width <= ...". You get the idea.

I added a few more changes to make it more informative
Screenshot(1)

PR #1402, opened 2019-05-30
filemnt.desktop: Add mimetype assosciation

rizalmart:
Add xdg compliant file association


s243a:
As a side note based on the mime info in the .desktop files we could auto-populate the rox file associations defined in:
/root/.config/rox.sourceforge.net/OpenWith

PR #1408, opened 2019-05-30
Create wine-wrapper

rizalmart:
Add a wine-wrapper script for handling windows programs on wine

PR #1409, opened 2019-05-30
Create wine-handler.desktop

rizalmart:
xdg desktop file association for wine


peabee:
#1408 #1409 #1410 
Should be issued as a .pet for proper testing....
should then be put into ibiblio repo so that wine support is OPTIONAL - included in build if required or installed from ppm by user if required.

PR #1411, opened 2019-05-30
pup_event_frontend_d: perform post-wine task

rizalmart:
If wine is installed it will run xdg-wine every ten seconds however if no wine program was running

### DIFF
--- a/woof-code/rootfs-skeleton/usr/local/pup_event/pup_event_frontend_d
+++ b/woof-code/rootfs-skeleton/usr/local/pup_event/pup_event_frontend_d
@@ -34,6 +34,14 @@ do
 	if [ -f /tmp/frontend_startup_lock ] ; then
 		continue
 	fi
+	
+	if [ $MINUTE -eq 10 ] ; then
+	  if [ "$(which wine)" != "" ] && [ "$(pidof wineserver)" == "" ]; then
+	    if [ "$(pidof xdg-wine)" == "" ]; then
+	      exec xdg-wine
+	    fi
+	   fi
+	 fi
 
 	MINUTE=$(( $MINUTE + 10 ));
 	if [ $MINUTE -eq 60 ] ; then


wdlkmpx:
One thing i noticed in xslacko the first time i tested it, is that it's not that lightweight, i used puppies way more lightweight, the ones i produce. It must be because of this sort of stuff.

I try to remove stuff and optimize the system performance, the opposite of this

wdlkmpx:
This bash script called pup_event_frontend_d that calls `sleep` -you can always see a `sleep` process in the slackos- will be replaced by a compiled binary again.. which behaves exactly the same. But it interacts directly with the kernel.

The 2nd or 3rd or 4th rewrite
https://github.com/puppylinux-woof-CE/initrd_progs/blob/master/pkg/w_apps_static/w_apps/pup_event_frontend_d.c

However i believe that app is redundant.. saves in pupmode 13 should only be triggered manually, no timer.

But one issue that was discovered is that the pup_event_frontend is almost really needed if you have a cd/dvd/bluray drive, well at least in woofce puppies udev fails to detect if a cd was inserted.. pup-volume-monitor also fails because it depends on udev, but pup_event_frontend_d triggers the update by trying to probe the disk state.

Otherwise pup_event_frontend_d can be considered optional or unneeded if you are radical.

I see pup_event_frontend_d might be ok for a new optional feature: update the system date every hour or 5 hours or 10 hours or 24 hours using ntpd (busybox) or nptdate.. for devices that don't have a hardware clock.

Although i'm a wine user, let's not forget that puppy also works in non-x86 hardware where  stuff that people use in x86 desktop pcs are rarely or never used.

A new processor architecture will be added to woof-distro soon (exp branch), i'll buy something.

ninaholic:
I try not to use wine, but when I need it usually I just run in ram (so it's cleared on exit) and type **winefile** in terminal to load things so a wrapper isn't needed.

jamesbond3142:
> But one issue that was discovered is that the pup_event_frontend is almost really needed if you have a cd/dvd/bluray drive, well at least in woofce puppies udev fails to detect if a cd was inserted.. pup-volume-monitor also fails because it depends on udev, but pup_event_frontend_d triggers the update by trying to probe the disk state.
> 

This has been unnecessary in the last six years, the kernel is able to do its own polling for devices. You need to configure the kernel to do this, but even if you don't, doing 

```
echo 5000 > /sys/module/block/parameters/events_dfl_poll_msecs
```

should do the job on modern enough kernel.

---

> Otherwise pup_event_frontend_d can be considered optional or unneeded if you are radical.

pup_event_frontend_d has a few jobs. Originally it was an udev replacement, but later on Barry decided to bring udev in (but still kept it because he would like to be able to run without udev - something that is very difficult nowadays; even if you don't udev you still need a replacement like mdev or mdevd or similar ...). So with udev in, its only purpose is to "unpack" packed firmware before it gets loaded. But this packing/unpacking job also lost its purpose when Mick decided to go with unpacked firmware (the files are unpacked but the entire firmware is packed as an SFS file, which overall makes things easier to manage). Now the only thing that is still pending in pup_event_frontend_d (the last time I looked at it):
1. Drive icons. If you use pup-volume-monitor then this is not needed.
2. Some other "background" stuff like background save, etc; which can be re-factored out into its own independent lightweight daemon.

Is it worth to do this? You decide. Fatdog64 lost pup_event_frontend_d it in 2012 and we never looked back. We still have drive icons (shell scripts using udev) and the periodic ram-save is handled by a 120-line shell script. Separating it into modular daemons makes it easier to manage; for example, we don't run the ram-save daemon if the (equivalent of) pupmode doesn't demands it.

wdlkmpx:
Well "kernel polling" does unlock a new level, that was good advice with an example. I tested it and it works, it's great to see it working.

I have this patch for the testing branch
[pup_event_kernel_polling__sleep-60.patch_fake.zip](https://github.com/puppylinux-woof-CE/woof-CE/files/3249919/pup_event_kernel_polling__sleep-60.patch_fake.zip)

It removes the code to "poll" every 4->6->10 seconds to see if a cd is inserted, it also changes the sleep from 10 to 60, every minute a lightweight operation happens.

The default ram save interval will be changed to 60 minutes, and it does happen, there was code to prevent the operation from happening if some specific apps were running, that's one of the reasons for the short delays, to keep trying until the save is performed.

So that's why i'm reluctant to add that change to testing, it's just one of the many cleanups and optimizations that must happen... or maybe i should restore the check "if some apps are running delay the save", but this time the delay is 1 minute. 

I don't know what to do.

wdlkmpx:
So, commit https://github.com/puppylinux-woof-CE/woof-CE/commit/c5fd425bc5a7cbaaeb6316435fb99f04c0ef8c25 basically adds the changes, but without introducing a different behavior

It's now a lightweight ash daemon, running `sleep 60` that you can always see running `ps`... how to avoid that, why is sleep  not a bash/ash builtin command?

I have a new pup_event_frontend_d.c that behaves exactly the same, should i just add static binaries to replace the script and the visible sleep 60?

The next step could be to get rid of it, currently it just sources /etc/eventmanager every 60 seconds and performs a few operations if (PM 13) RAMSAVEINTERVAL and/or POWERTIMEOUT are enabled, it's a bit dynamic and there's no need to restart it, it could be split into 2 daemons calling sleep 60 each one.. when the respective value is enabled. Hmm

wdlkmpx:
pup_event_frontend_d has changed..... so the change in this PR could happen every 60 seconds.

But pup_event_frontend_d will cease to be a "daemon" in the future.

So my advice is that you create a launcher script or daemon in /root/Startup or /etc/xdg/autostart .. and create a pet package for your wine-wrapper/xdg stuff, that you might want to share as a pet package.. 

Things change, and in the experimental there are many more structural changes and it's easy to build a woofce for puppy testing purposes... it's faster in the experimental branch, and completely automated using ./build.sh

wdlkmpx:
There is https://github.com/puppylinux-woof-CE/woof-CE-noarch

A place where anybody can open a pull request or an issue with an url pointing to a pet pkg to upload/update, the noarch pkg db in woof-ce will be updated afterwards, so builders con optionally include it and users can install it from the PPM.

Although i must that i messed it up when i moved the noarch repo from ibiblio to a github repo. I wasn't thinking straight, it's dead anyway..

peabee:
For example:
https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-distro/x86/slackware/14.2/DISTRO_PET_REPOS
won't ibiblio be used in preference to github??
```
PET_REPOS='
z|http://distro.ibiblio.org/puppylinux|Packages-puppy-*-official
...............
z|https://raw.githubusercontent.com/puppylinux-woof-CE/woof-CE-noarch/master|Packages-puppy-noarch-official
'
```

and what happens to ibiblio noarch? will it get out of date? what happens to all the pups in existence that still point ppm at ibiblio?

wdlkmpx:
Not if the package is unique to noarch or if it's found in the noarch
repo, 1download knows where all the package come from. I guess....

findpkgs was simplified a bit in the experimental branch, it will
strictly follow PACKAGELISTS_PET_ORDER, without strange logic
regarding grepping the distro compat name, etc.

They will get outdated, but i did  this months ago.. didn't you notice?

I was planning to move the rootfs-packages to noarch, and i was indeed
doing it. bold decisions must be taken sometimes. I later realized
that both build systems in the experimental branch could use the same
rootfs-packages.

It can be reverted if you wish, back to ibiblio.. but it's dead.. is
it hard to accept it's dead? It just doesn't work like that.

The quickest solution i found was a git repo, with pull requests and
issues so that a package is identified as compatible with woofce, not
all noarch packages are compatible with woofce... needs testing and
cleanups.

It's a changing structure/architecture, didn't i mention that?

The other quickest solution is a dedicated thread on the forum, but
only those who have access to ibiblio should open and monitor that
thread, otherwise it doesn't make sense.

This message couldn't be any clearer.

PR #1413, opened 2019-05-30
remasterpup2: some improvements

rizalmart:
1. Adjustable block size compression
2. Add support for initrd.xz


rizalmart:
Adjusting block size upon squashfs compression affects the sfs file size. The higher the block size set the smaller the sfs files created. however higher block size setting needs more cpu overhead was needed so it tooks longer time compress completely the filesystem into sfs file. So in my revision let the user set their prefer block size for creating the main puppy sfs file whether they want larger sfs file but faster to compress or smaller sfs file but slower to compress

PR #1415, opened 2019-05-30
overhauled sns

rizalmart:
With add automatic and manual ip settings
With refresh network interface
With refresh wifi scanning for wireless
Improved interface


wdlkmpx:
This should be reviewed by @rserwin1  as it might undo fixes that must be applied again

rizalmart:
I think rerwin1 must analyze the I code that I sent very well

wdlkmpx:
It's ok to keep pull requests open, other people might have opinions
to share. i'm not the only one

Testing something might take some time, i guess i'm the only one who
tests stuff in pull requests, but i'm also busy doing many other
things.

Bear in mind that basically it's all pretty dead here, so don't worry,
it's all dead..

PR #1420, opened 2019-05-31
shutdownconfig: Improved shutdownconfig dialog

rizalmart:
Instead of flashy countdown timer shown, just a progress bar that indicates countdown
based from shinobar's shutdown save session interface


wdlkmpx:
I haven't tested this, but it looks good to me. I noticed the flashy countdown timer doesn't show up in a raspi 1, for some reason, this may fix the issue. But what others think about this..

wdlkmpx:
The code to determine SAVECHOICE is broken.. 

This seems to work fine:
```shell

   eval "$(gtkdialog -p XY_DIALOG --center)"
   case "$EXIT" in
      abort) SAVECHOICE=255 ;;
      0)     SAVECHOICE=0   ;;
      *)     SAVECHOICE=zz  ;;
   esac
```

That dialog seems to be a candidate for a new box_dialog with --params, box_countdown. The final countdown.

PR #1423, opened 2019-06-02
Create pupmtp

rizalmart:
pupmtp is an application for mounting mtp devices via fuse by using simple-mtpfs by phatina 
Source code: https://github.com/phatina/simple-mtpfs
Requires fuse, libusb, and libmtp only

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/pupmtp
+++ b/woof-code/rootfs-skeleton/usr/sbin/pupmtp
@@ -1,0 +1,95 @@
+#!/bin/sh
+#pupmtp written by mistfire. Interface and behavior based from pmount for Puppy 4.0
+#manage mtp devices
+
+#using ${$} which is pid of script...
+MYPID=${$}
+
+. /etc/rc.d/PUPSTATE #v3.96
+
+[ ! -e /media ] && mkdir /media
+
+MINIFOLDERXPM='/usr/local/lib/X11/mini-icons/mini-filemgr.xpm'
+
+ALLPS="`ps`"
+OLDPIDS="`echo "$ALLPS" | grep 'gtkdialog' | grep 'PMTPGUI' | tr -s ' ' | sed -e 's%^ %%' | cut -f 1 -d ' ' | tr '\n' ' '`"
+kill $OLDPIDS 2>/dev/null
+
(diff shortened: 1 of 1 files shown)


wdlkmpx:
I compiled libmtp + simple-mtpfs, but i'm not sure how to test, i connected an android tablet and a phone, but apparently i don't know how to enable stuff haha. When i want something i just remove the micro sd card

So i guess i can merge this without testing it..

wdlkmpx:
It also requires libgcrypt for MTPZ devices  (e.g. Zune). Which can be disabled --disable-mtpz

I found that  libmtp + simple-mtpfs quality as generic packages with --disable-mtpz, it compiles in slacko 14.0, and runs in later versions of slackware, debian, ubuntu...

rizalmart:
I already tested that in my smartphone and it mounts the mtp and performs transfer files between across pc and mtp devices and vice versa

wdlkmpx:
Ok, i'm merging this i haven't been able to make it work, perhaps my device is not supported

jamesbond3142:
> Ok, i'm merging this i haven't been able to make it work, perhaps my device is not supported

Many reasons for that, yours is not the only one. We (=Fatdog) have simple-mtpfs-based MTP browser for a while now (which was based from what Mick wrote here: http://murga-linux.com/puppy/viewtopic.php?p=839972#839972 ). But your pupmtp seems to have a better looking GUI :)

First, libmtp has a list of known identifiers - if your device isn't in the list it won't work. But you can easily patch libmtp and re-compile to add your own device ID. Secondly, it has another issue, some devices are very picky; and/or libmtp has very serious timing issue. Branded devices (Samsung devices etc) are apparently more tolerant of libmtp and can happily work with it; but lesser known brands don't fare so well.

Well for me, it was so annoying because my phones are mostly not Samsung phones. I was forced to find another solution: adbfs. It worked for a while, but only up to Android 6. Adbfs failed completely on Android 7 onwards, oh joy :( 

I'm still looking a solution that works. Perhaps it is much simpler to install ftp server on the phone and use gftp for transferring files.

jamesbond3142:
PS: the above isn't a criticism of puptmp; rather criticism of general state of matter with MTP devices. It's good have this merged for people whose phones __do__ work with MTP properly.

PR #1446, opened 2019-06-03
kernel-kit/build.sh : strip-unneeded for kmodules

rizalmart:
This will search for all compile kernel modules and make the kernel modules smaller without being broken. This will activate if strip command exists

### DIFF
--- a/kernel-kit/build.sh
+++ b/kernel-kit/build.sh
@@ -1026,7 +1026,15 @@ else
 	KERNEL_MODULES_SFS_NAME="kernel-modules.sfs-${kernel_version}-${package_name_suffix}"
 fi
 
+if [ "$(which strip)" != "" ] && [ "$(strip --help | grep "\-\-strip\-unneeded")" != "" ]; then
+ for mods1 in "$(find "$(pwd)/output/${linux_kernel_dir}" -type -f -name "*.ko")"
+  do
+   [ $(file "$mods1" | grep "unstripped") != "" ] && strip --strip-unneeded "$mods1"
+  done
+fi
+
 mksquashfs output/${linux_kernel_dir} output/${KERNEL_MODULES_SFS_NAME} $COMP
+
 [ $? = 0 ] || exit 1
 cd output/
 if [ "$kit_kernel" = "yes" ]; then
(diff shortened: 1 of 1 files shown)


ninaholic:
Hi rizalmart,

Interesting idea. I think though, that the -b 1M is supposed to be defined here instead:
https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/kernel-kit/build.conf#L81

mavrothal:
```--strip-unneeded``` may break the modules as it may remove the names of the symbols that the module cals (insmod may not work).
```--strip-debug``` can be used instead but  puppy kernels are build with  ```CONFIG_DEBUG_INFO is not set```
Unless something has been changed in the strip utility lately  you may want to close this pull request

ghost:
I think @mrfricks used --strip-unneeded on some of his kernels, it would be interesting to know what his method of doing this was and if he still does this or came to the same conclusion as mavrothal.

wdlkmpx:
This can be an option that is disabled by default, then someone can test and assess how dangerous this is.

Sage might get mad when something does not work or seems to be broken.

The extra stripping in 3builddistro should not be a default either, years ago i wondered why an app was always crashing.. i found that one of the libs was segfaulting because it was stripped of its precious unneeded symbols.

rizalmart:
I already tested the stripped kernel modules on both real machine and virtual machine using that command parameter. From 80Mb+ total size of kernel modules it become 70Mb+ (9Mb+ size reduction). So far it works fine to me. I wonder to it will work fine to others

wdlkmpx:
I certainly won't test this, but i'll add it as an option that is disabled by default. You should learn to use build.conf.

That `-b 1M` is part of the COMP variable in build.conf, not something to add later.

I also wonder if the "Linux-libre (AKA deblobbed kernel)" code should be removed..

mavrothal:
> I already tested the stripped kernel modules on both real machine and virtual machine using that command parameter

Issues are may appear when connecting new hardware requires a new module that calls other modules NOT already  loaded.
It  might generate problems hard to detect down the road as will be hard to detect that the kernel is stripped and kernel messages might be missing specifying the issue at hand.
It might worth  adding some warning in the build.conf option.

wdlkmpx:
I added a warning.

Has anybody used LIBRE=1 recently? I wonder if that code is still working fine..

PR #1447, opened 2019-06-03
3builddistro-Z: Make output sfs files smaller

rizalmart:
This was achieve by stripping library files and executable files which was not stripped and set mksquashfs at 1M block size

### DIFF
--- a/woof-code/3builddistro-Z
+++ b/woof-code/3builddistro-Z
@@ -41,6 +41,32 @@ PKGLISTS_COMPAT="`echo "$PKG_DOCS_DISTRO_COMPAT" | tr ' ' '\n' | cut -f 3 -d '|'
 
 export WKGDIR="`pwd`"
 
+strip_binary_files(){
+
+sPATH="$1"
+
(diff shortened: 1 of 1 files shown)


wdlkmpx:
This will make the build way slower, it also doesn't take into account _00build.conf... SFSCOMP is defined there https://github.com/puppylinux-woof-CE/woof-CE/blob/8f4b5dbc52633e4fef518c27721c20e20d0b1014/woof-code/_00build.conf#L63 

Stripping of binaries may happen in 2createpackages:
https://github.com/puppylinux-woof-CE/woof-CE/blob/8f4b5dbc52633e4fef518c27721c20e20d0b1014/woof-code/2createpackages#L712

 and in 3builddistro..
https://github.com/puppylinux-woof-CE/woof-CE/blob/8f4b5dbc52633e4fef518c27721c20e20d0b1014/woof-code/3builddistro-Z#L1009

and it's optional in both cases.

This adds a new routine that is not optional, stripping of binaries may happen 3 times, and one of them is not optional

2createpackages already does this, optionally.

Does not qualify... will not be merged
And also a build is painfully slow when stripping is enabled..

rizalmart:
I remove now the file stripping. What do you think for this patch?

rizalmart:
Okay this pull request is now closed

wdlkmpx:
it's still not mergeable.

Wce is about build options, builders decide what compression methods or params are used and that is defined in _00build.conf

PR #1448, opened 2019-06-03
rc.shutdown: Remember installed hardware devices

rizalmart:


rizalmart:
Remembering installed hardware devices upon shutdown is the vital key for more portable puppy.

Here is how it works (concept only).
When puppy is booted from a certain machine to a new machine. It will compared the current installed hardware against the hardware which was detected on last. If it was different then it means there was a change in hardware so it will delete the xorg.conf file and reset it to default. Display setting of zarfy, grandr, arandr, and Desktop Environments will be deleted to reset it at default. Also networkmodules list will be repopulated.

jamesbond3142:
I suggest we don't follow the way of Windows. Puppy should be dynamically re-configuring itself at every boot - except for things that the user has specifically override (e.g GPU, Xorg driver).

rizalmart:
@jamesbond3142 you got what I mean. This was problem that still persists on puppy if the user always boot puppy on different machines and loads the same savefile.

jamesbond3142:
I've been away from Puppy too long so I don't know what's the state of the art right now. I'm still quite surprised to hear from you that it still create static configurations. Back in woof-next days I made Xorg run without xorg.conf; and I thought it had made it into the Woof-CE proper. I can't say anything else about saving soundcards etc.

But here is the point: You remember there was once a time where we all want Puppy to dynamically re-configure itself whenever it gets run in a new environment (even with a savefile); the main scenario being a USB flashdrive that you carry along with you. To achieve this, Puppy should __not__ save hardware configurations, except when absolutely necessary. The main issue at that time was Xorg. When running with Xvesa, it was easy - Xvesa can run on any machines without configuration (well you can set its resolution but if it doesn't show up because the monitor doesn't support the frequencies, you can always boot with pfix=nox and fix it). Xorg was more difficult, it needed a config file - but those days were long gone. Under most circumstances auto-configured Xorg will run very well on most hardware except the most exotic ones.

So I guess if you want to achieve this ability to re-configure Puppy at every boot (even when using a savefile), my suggestion is to first examine which config files gets created "automatically", because it shouldn't be. Config files should only be created by user requests (e.g to override a non-working Xorg, etc); not automatically. So you don't need to save hardware details etc - it will get confusing very fast (in my opinion). 

Now, once you're left with only user-created config files (which should not be that many), then you can detect whether machine has been changed. In my opinion, it's better to detect if the machine is different from the last time it was used, rather than trying to do an item-by-item hardware configuration. There are many ways to "id" a machine, "dmidecode" is one of them (I think dmidecode binary is in Puppy already by default). If you detect that the machine id is different from the last time it was run, then offer a popup dialog to clear all known user-created config files.

cheers!

rizalmart:
I have an idea on how to fix the xorg dilemna however it would require some radical changes.

1. always user defined configuration of xorg will reside at /etc/xorg/xorg.conf.d folder and let xorg.conf file dynamic.

2. Obtain xorg touchpad configuration by using X -confugure :1 command and extract it from xorg.conf.new and place it at /etc/xorg/xorg.conf.d

I will try this on my x-slacko slim before deploy it as pull request. I hope it works

wdlkmpx:
The understand this issue, we must understand the that a wce puppy works with and without xorg evdev.

The reason to keep both is that sometimes evdev may not work with a certain combination of  kernel version+xorg, specially with incompatible kernels you can select when building a wce puppy, or when you change kernel with an older kernel and you find your mouse dead.

To use xorg evdev stuff, the builder must not include the legacy mouse and keyboard xorg drivers, i guess this also prevents the synaptics stuff from being used.. this is the main problem. The xorg evdev stuff is also triggered when a touchpad is detected.

It's a mess..

woofce is different from x-slacko, the same logic may not always apply.

The problem i see here is 2 methods are supported, and some people are comfortable using the old method with their touchpads and so on, it's more "stable" in their machine with synaptics, and this indeed makes it less portable with that configuration.

That's why even a modern puppy with the latest compat distro is usable in really ancient machines..

People assume that the wce code is static and doesn't change.. it has changed quite a lot... so the code might be different from x-slacko, and so the same solutions may not apply.

wdlkmpx:
Edited my comment above just to clarify that a xorg evdev/udev stuff is also triggered when a touchscreen is detected, xorg.conf is deleted and a different xorg.conf is used, so that the puppy scripts to configure xorg stuff keep working but with the evdev driver (the new xkbconfigmanager is fixed to work evdev)

So, there is logic to delete xorg.conf but only to change between 2 xorg modes, only one of them makes the system fairly portable.

When the system is using the evdev driver or automatic adding of devices, rc.sysinit creates /var/local/xorg_udev

from rc.sysinit
```
### /var/local/xorg_udev is created when:
### - a touchscreen has been detected
### - missing xorg input-kbd and input-mouse drivers [ex: ubuntu xenial]
```

So if the xorg_udev flag file is missing, i guess rc.shutdown can just store the video chip id and that could be the only thing to compare to determine if puppy is being used in another machine, and xorg.conf is deleted. It's only a problem when changing from laptop -> pc or pc -> laptop i think.. but i haven't verified this, since i don't use the same pupsave in different machines.

jamesbond3142:
Do we still need that complications? Fatdog went xorg.conf-less in 2011, that was 8 years ago; and back then it worked on hardware as old as from 2007 (I don't have anything older anymore). The kernel was 2.6.39.4, I doubt that even a retro version of Puppy uses kernel older than 3.x nowadays.

wdlkmpx:
Well, i remember phil666 said that the xorg evdev stuff doesn't work with 3.xx and ubuntu puppy xxx.. tahrpup i think.

It means that for some reason, the evdev stuff is not compatible with certain combinations of kernel+xorg. or something.

I read people have been complaining how evdev misbehaves with some touchpads, the fixes arrived probably in 2016+, meanwhile they were using synaptics

So, for xx reasons, woofce supports many possible combinations and it's hard to know what to do without causing an uproar, since the project doesn't have dedicated testers, stuff is released directly without previous filters.

Back in the day we and micko had trouble with stuff and decided to xorg.conf while adding evdev support.

I guess the complication can be removed in the experimental branch, using only xorg evdev but still keeping xorg.conf, but that complication is what makes puppy highly stable for many people. Testng is needed to be aware of some gotchas..

wdlkmpx:
I also noticed a slightly different behavior with evdev vs kbd + mouse.

I also think i read issues with joysticks, gamepads, the essence of life.

In this machine, I noticed the mouse pointer moves, what is the correct word? "shaking" "trembling" without moving the usb mouse, when i unplug the mouse the ghostly activity ends. The keyboard seems to work fine. It doesn't happen with mouse_drv

But as jb notes, this complication makes puppy a bit unpredictable and "a mess", so this might get removed in the experimental branch (while still keeping xorg.conf).. @ninaholic might want to test this stuff in in slacko14.0 in netbooks by deleting :

/usr/lib/xorg/modules/input/kbd_drv.so
/usr/lib/xorg/modules/input/mouse_drv.so

rizalmart:
For the meantime, I will just avoid messing xorg.conf. I will just deal with display settings of desktop environment first.

jamesbond3142:
> Well, i remember phil666 said that the xorg evdev stuff doesn't work with 3.xx and ubuntu puppy xxx.. tahrpup i think.

Yes, I remember this, but I wasn't directly involved and I didn't investigate the cause so I have little to say about that.

However, as far as Xorg autoconfiguration is involved, the key is udev. Xorg depends on udev to find out keyboard, mouse, touchpad, etc etc. Example: in one of my adventure with woof-next with debian sid recently, I didn't pull enough of systemd dependency to make udev happy - so udev didn't work. And when the system went to Xorg - I've got a nice desktop, but no mouse or keyboard. So if your auto-configured keyboard/mouse has problem, the first culprit is udev. And it is true that certain version of udev depends on certain config feature to be enabled in the kernel; if you don't have it then udev will often fail to work properly (and many times silently), so yeah there is a bit of dependency there. 

But if we want to move on like what rizalmart suggests, these are the things that you need look at.

> In this machine, I noticed the mouse pointer moves, what is the correct word? "shaking" "trembling" without moving the usb mouse, when i unplug the mouse the ghostly activity ends. The keyboard seems to work fine. It doesn't happen with mouse_drv

I do note some differences between kbd_drv.so and evdev; since the keycodes are not the same (evdev keycodes is higher-numbered than kbd_drv.so); but in any other aspect I have not seen any difference. I'm not denying that what you said could not happen; I'm just saying it sounds weird. But we know many weird things do happen.

jamesbond3142:
PS: Just to clarify, I'm not advocating for autoconfiguration. If you aren't comfortable with that idea then by all means don't. I'm only saying this because rizalmart wants to have dynamically configured Puppy at every boot (which is good idea btw), and I just thought that one of the first thing to do is to get rid of the Xorg.conf stuff. If you think this is not worth it, then don't worry about it.

Back on topic to rizalmard - I still think that perhaps instead of detecting video chip id etc, just detect the "PC Serial Number" which you can get from dmidecode. Hash it (using md5) for privacy reason, and check if it is still the same as the one used previously. If not, then popup a dialog offering to delete all configuration stuff. Of course, this is assuming that the system works well enough to be able to display that popup and allow the user responds to it ... so you'd probably want to do this in CLI before going to terminal. But again, this is just an idea, suggestion. Take it if you like it, if not, then just drop it.

cheers!

wdlkmpx:
Well, this is how things work in woofce:

There are 2 xorg config "templates", and 2 different behaviors depending on the xorg.conf being used.

 xorg.conf is deleted and replaced with one of the templates when a sensible change is detected.. this is an unpredictable behavior, or a quirky behavior.

xorg.conf.udev is farily "neutral", despite being there it doesn't interfere with detection of stuff, it also makes it almost impossible to support serial mice (not talking about ps/2 mice).
Its purpose is to produce some magic for some reason and to let the traditional puppy scripts  edit configuration options.. it took a while to fix those scripts, instead of using some parts a desktop environment.

So, the current stuff is confusing, well at least making xorg/udev/evdev the default behavior will lead to a more consistent and portable system... by default.

Making xorg udev the only option will make everything less confusing and more predictable, it's all in the reflexes.

rizalmart:
jamesbond3142 dmidecode is good however it cannot detect changes if the same machine is used but video card and monitors was changed.

jamesbond3142:
> jamesbond3142 dmidecode is good however it cannot detect changes if the same machine is used but video card and monitors was changed.

Yes, you're right. But how often does it happen? The USB flashdrive being used in multiple computers is a common use case. But monitor replacement and video card are probably aren't that common; and my suggestion to handle that case is another pfix parameter, perhaps pfix=resetconfig or something like that.

rizalmart:
In case that a certain monitor or video card was broken and need to replace. This is just a proactive measure

wdlkmpx:
As i said a wce puppy can work in 2 modes. Very few pups use xorg udev by default.. xenialpup64, bionicpup64.

This may not be needed if xorg udev is enabled globally. It's good to remember that woofce code is not static, it keeps changing, and a ton of changes already happened, but are not merged into the default branch (testing)

PR #1449, opened 2019-06-03
_00build.conf: Set SFS comp block size to 1M

rizalmart:


jamesbond3142:
This does makes the SFS produced smaller, but the penalty is the run-time cost of loading programs later on will be much greater. We have had a similar discussion in Fatdog and we decided it is not worth it.

wdlkmpx:
The current default is:
SFSCOMP='-comp xz -Xbcj x86 -b 512k'

In dpup i just use -comp xz, i usually build with a p4 machine (trying to kill it), and it really takes a loong time with heavier compression methods, it already takes a long time with only 'xz'. I think i'll use gz with fewer packages.

I guess with 1M IT will also affect the running system with underpowered processors.. it will take much longer to load the system

It's not a good default. It can be a comment of course, just to let the builder know there are other options..

ninaholic:
It is certainly faster to finish building dpup than slacko for me too because the "compressing the sfs and devx" is the most time consuming part, especially on single core CPU or single+hyperthreading. Compressing the devx alone with -b 512K for me can take about 40 minutes. I use old machines to build too and they just keep on living :laughing:

jamesbond3142:
> In dpup i just use -comp xz, i usually build with a p4 machine (trying to kill it) ...

LOL. For the published builds of Fatdog we only use "-comp xz -Xbcj x86", with default blocksize. For test builds, I just use "-comp gzip -Xcompression-level 1" because I can't be bothered to wait that long.

> I guess with 1M IT will also affect the running system with underpowered processors.. it will take much longer to load the system

You will be surprised that the impact is visible even with high-powered processors when running it.

wdlkmpx:
So we now know why x-slacko slim seems to be unusually "heavy and slow" for a lightweight distro... if we achieve 150MB in size that's quite small by today standards, tiny... there's no need to push the limits and at the same time making something really slow, but "small".

I think i'll change the default
 SFSCOMP='-comp xz -Xbcj x86 -b 512k'

to
 SFSCOMP='-comp xz'

PR #1450, opened 2019-06-03
umount: remove empty mountpoint folders on /media

rizalmart:
The mountpoint folders created in /media will be removed if it is not used as mountpoint and it was an empty directory


wdlkmpx:
I'm testing this and i see the mountpoints are deleted by filemnt.

The media directory is empty, rc.sysinit also deletes empty directories in /media and /mnt

peebee must be cherry-picking changes again

peabee:
I was asked to test but only tested to the point that I saw an empty directory remained in the savefolder after shutdown.....I didn't reboot so if you say the directory is deleted at rc.sysinit then that should be fine..........

PR #1451, opened 2019-06-03
mscw2 and alsaequal.sh - taken from Fatdog64

jamesbond3142:
mscw2 is the fatdog-default-soundcard taken from Fatdog64 803. In addition to selecting the active soundcard, it also has options to enable/disable the sample rate converter, equaliser, software volume control, pre-amp, swapping stereo channel, downmixing, and enabling mixout (record what you hear).

A companion script is included as well: alsaequal.sh, which enables control of the equaliser (when equaliser is enabled on mscw2). While mscw2 is a GUI application, alsaequal.sh is a terminal app.

The look and feel isn't the best since I use Xdialog, but it is fully packed in terms of functionality. Screenshots attached.

xscreenshot-20190603T191831

xscreenshot-20190603T191813

xscreenshot-20190603T192019

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/alsaequal.sh
+++ b/woof-code/rootfs-skeleton/usr/sbin/alsaequal.sh
@@ -1,0 +1,23 @@
+#!/bin/ash
+# jamesbond 2011, 2014
+# updated Fatdog 700 - simplify, remove control, leave with "mixer" functionality only
+# 131130 L18L internationalisation
+
+# std localisation stanza
+export TEXTDOMAIN=fatdog
+. gettext.sh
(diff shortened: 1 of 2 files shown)


wdlkmpx:
This is a cli/gui app, so it also works in cli-only systems.

I briefly tested this, it seems to work as expected. I got confused because a few times it wrote to ~/.asoundrc but ASOUNDRC is not defined, it should always use /etc/asoundrc as established by rcns551.. there's some voodoo going on in dpup that just can't be explained with science

It's part of alsawizard (Menu->Setup->ALSA sound wizard)
Screenshot(2)

PR #1452, opened 2019-06-04
sfs_load: add porteus and slax modules support

rizalmart:


peabee:
This is dangerous if people install a 'foreign' sfs into a system with a different architecture - especially 64-bit. Puppy sfs are tested to be compatible with puppy and named appropriately if they are architecture specific.
Same result can be achieved by simply renaming the foreign modules to .sfs - but then the renamer would know what they were trying to achieve.

rizalmart:
Just add a warning message if non puppy sfs file is loaded by the user. Also renaming or changing file extensions was a hassle tedious task for newbies.

peabee:
Newbies should not be even considering installing 'foreign' sfs modules....

rizalmart:
Yes puppy has tested sfs modules. But producing an sfs modules is hard also the question is how far can the maintain those sfs modules. Maintaining a repository of a certain package format requires dedicated time for maintenance or else it will ended like antergos. Thats why I add those formats to increase extensibility but with a caution.

mavrothal:
@rizalmart 
This and others are interesting changes. However I think they are  directed at the well informed puppy user or (re)builder.
Puppy theoretically is an entry system for windows refugees rather that linux-hackers.
Maybe better idea to provide those as pets for the devx, assuming that  one loading/using devx nows what (s)he is doing.

s243a:
> @rizalmart
> This and others are interesting changes. However I think they are directed at the well informed puppy user or (re)builder.
> Puppy theoretically is an entry system for windows refugees rather that linux-hackers.
> Maybe better idea to provide those as pets for the devx, assuming that one loading/using devx nows what (s)he is doing.

I think puppy is a mix of both types of users but you can only protect the user so much. For instance if the user tries to install base libraries from the .deb files. Actually I tried this once upon a time following the debianization section of the puppy wiki! Needless to say it didn't go well!

That said I suppose this could be a pet but it is the type of thing that I would want installed so I'm happy either way.

wdlkmpx:
I'll test this soon soon with some module, to see if it's even possible to trigger filemnt by clicking on some alien sfs module.

I think it's not possible, squashfs files would normally be recognized by xdg mime stuff or something no matter the file extension, but  rizalmart added mimetypes that make those extensions  different file types (not a squashfs). 

It's not hard to maintain sfs files, you put a browser in it, replace it with the next version, repeat the last operation. You don't need to compile the browser.

You need a file hosting site to store that file.

I doubt the porteus people compile those big modules, those modules come from bigger organizations (firefox) or from the compat distro repo.. yes that module comes from slackware/debian/etc.. something you can install without their help.

The hard part is store anything in the puppy ibiblio repos, that's the only hard part, if you're an experienced puppy user, you know that there are people that produce sfs files that you can just download and use, but you're going a step further looking somewhere else.

For example, version2013 compiles the latest WINE version for puppy4+, wary5+. tahrpup+ .. that alone opens a whole world, a universe of apps you can run in your old puppy. This fact alone has universal implications and ramifications.

To be continued..

wdlkmpx:
Nothing happens when you click on porteus/slax modules unless you add new "file associations" for those file types, and that won't happen in woofce, but can happen outside.

I think it's ok to load or mount alien sfs modules by manually running `filemnt/sfs_load zzz` (that only can be done by someone who knows what they're doing), but it's not ok to search for them.

Also this patch is incomplete.. i think this hasn't been tested in a recent build.

I downloaded firefox.xzm from http://ftp.vim.org/ftp/os/Linux/distr/porteus/i486/current/modules/

tested the code, fixed the code, simplified the code and it's ready, but should i remove the code that searches for alien sfs modules?

The patch  adds 52 lines, my fixed patch adds 9 lines

It loads the .xzm file, but firefox doesn't work: `Couldn't load XPCOM.`

If you download firefox from the mozilla site, even the latest version work in old distros. And this old firefox doesnt work in dpup.

Hmmm i think.. these files should be renamed to .sfs by someone who knows what they're doing.

wdlkmpx:
I found that the firefox xzm i downloaded is... incomplete, it has symlinks to files that are included in another xzm module i guess... the base sfs probably

Should i install 001-core.xzm?

wdlkmpx:
Commit 0d837023e36b02da73f8e9d943061aff6c49af8a partially merges this.. adding only 4 lines.

you'll have to run `sfs_load slaxmodule`, it only searches for sfs files when you launch gui.

PR #1453, opened 2019-06-05
Create 89-puppy-mtp.rules

rizalmart:
Udev rule for auto detection of MTP devices

### DIFF
--- a/woof-code/rootfs-skeleton/etc/udev/rules.d/89-puppy-mtp.rules
+++ b/woof-code/rootfs-skeleton/etc/udev/rules.d/89-puppy-mtp.rules
@@ -1,0 +1,2051 @@
+# UDEV-style hotplug map for libmtp
+# Put this file in /etc/udev/rules.d
+
+ACTION!="add", GOTO="libmtp_rules_end"
+ENV{MAJOR}!="?*", GOTO="libmtp_rules_end"
+SUBSYSTEM=="usb", GOTO="libmtp_usb_rules"
+GOTO="libmtp_rules_end"
+
+LABEL="libmtp_usb_rules"
+
+# Some sensitive devices we surely don't wanna probe
+# Color instruments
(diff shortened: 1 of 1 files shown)


wdlkmpx:
This can't be merged. It's part of libmtp.

I compiled the latest version of libmtp and this file has 2540 lines... it's bigger.

A new .petbuild will be added, a generic or common i686/x86_64 package will be compiled, also a amrhf/aarch64 package, that's optional, builders may use other packages.

rizalmart:
Is there any way to autodetect mtp-devices using udev?

wdlkmpx:
Of course, libmtp uses udev rules, those udev rules trigger apps that
are part of libmtp. The file name is
/lib/udev/rules.d/69-libmtp.rules, part of libmtp, in this pull
request you changed the filename, and that will cause problems to
those who include libmtp in a build or installs it from the repos.

libmtp is found in slackware/debian/ubuntu repositories, but it's an
older version.. The latest version can be found in debian buster+ and
ubuntu disco+, slackware current.

But it also compiles and works in slacko 14.0, funny isn't it.

You seriously need to compile stuff, a good developer compiles and
knows stuff, a puppy user is forced to compile or hunt for packages in
the puppy forum. Or use wine.

Or in the worst possible scenario, uses packages from many distros,
including slax and porteus, arch linux, slackware current.

A puppy is a bit fat because it already comes with everything you need
to compile.

The following apps are part of the libmtp package, which also include
udev rules and a shared library:

mtp-albumart
mtp-albums
mtp-connect
mtp-delfile
mtp-detect
mtp-emptyfolders
mtp-files
mtp-filetree
mtp-folders
mtp-format
mtp-getfile
mtp-getplaylist
mtp-hotplug
mtp-newfolder
mtp-newplaylist
mtp-playlists
mtp-reset
mtp-sendfile
mtp-sendtr
mtp-thumb
mtp-tracks
mtp-trexist

PR #1458, opened 2019-06-06
88-puppy-autodetect.rules: Add MTP detection

rizalmart:
Based from udevadm info --query=all --path=[usb path from udevadm monitor] INTERFACE value of MTP lies at 255/255/*


wdlkmpx:
pupcamera & pupmtp should only be triggered if gphotofs & simple-mtpfs are installed and the devices are properly recognized.. something that is not happening in puppy.

libmtp includes this file:
- /lib/udev/hwdb.d/69-libmtp.hwdb

Which defines recognized ID_MTP_DEVICE s

This doesn't look good (your last commit):
` ENV{INTERFACE}=="255/255/*"`

This seems to be the correct approach:
`ENV{ID_MTP_DEVICE}=="1"`

This doesn't look good:
`ACTION=="add", SUBSYSTEM=="usb", ENV{INTERFACE}=="6/1/*", RUN+="/usr/sbin/pupautodetect camera"`

-- /lib/udev/hwdb.d/20-libgphoto2-6.hwdb

defines all recognized ID_GPHOTO2 devices

The current stuff in some cases might trigger pupcamera with the wrong devices.. fixed by b6828a3e53a73682b2abf8d7b2313bca70ee2028

PR #1461, opened 2019-06-10
removemodes.sh : Improve report dialog

rizalmart:


woof-code/rootfs-skeleton/usr/local/petget/removemodes.sh
```diff
@@ -39,8 +39,7 @@ report_window () {
  
  if [ -s /tmp/petget_proc/overall_petget-deps-maybe-rem ];then
   MAYBEREM="`cat /tmp/petget_proc/overall_petget-deps-maybe-rem | cut -f 1 -d ' ' | tr '\n' ' '`"
-  MAYBEREMMSG1="$(gettext 'The following package(s) are dependencies for the package(s) you just removed. You may want to remove them too or reinstall the package(s) you just removed'):
-<i><b>${MAYBEREM}</b></i>"
+  MAYBEREMMSG1="$(gettext 'The following package(s) are dependencies for the package(s) you just removed. You may want to remove them too or reinstall the package(s) you just removed'): ${MAYBEREM}"
```
mavrothal, line 42: I see that in this one (and the other 2 dialog related commits) you remove any text highlights.
This one in particular also provides the list of packages in the same line as the information text.
Of course aesthetics are  different and changing with time but would be a good idea to provide side by side images for an immediate visual suggestion.

Sorry for not commenting on the other pulls directly, but I also notice that you block gtkdialog window resizing in some cases, why not. 
Also code indentation  one case in #1462 does not exist. @wdlkmpx will rewrite it!  😈

s243a: Speaking of code indintation I prefer two spaces rather than tabs but I suppose tabs has one the battle and I'm in the minority with this preference. I have this preference becuse in most editors the spacing for tabs is too much by default and is very unkind to people with small screens or poor eyesight.

The whole "tab=4 spaces" assumes that people use a small font (i.e. good eyesight) or have a very large monitor!

Actually in bash/ash there is another advantage to using spaces over tabs. The advantage is that there is an option with heredocs to ignore leading tabs. So if you use spaces for you main indentation than you can indent your heardocs with tabs and have the leading tabs removed automatically in the final string but at the same time keep any space-type indentation in the final string.

wdlkmpx:
I don't see problem with tabs, less keystrokes. The code tends to look ok
in geany, but not in the console or leafpad. I use geany.

With 1 space it requires an extremely good eyesight. I think 3 spaces is
ok.. too much keystrok3s, so i use 1 tab

I use a 15" monitor, it's not that big. I try to protect my eyes, i'm
usually 1 meter away from the monitor.

PR #1463, opened 2019-06-10
installpkg.sh : Parse package information for non-pet

rizalmart:
parse Debian, Slackware, Arch, and Red Hat package information for PPM

### DIFF
--- a/woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
+++ b/woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
@@ -116,6 +116,163 @@ install_path_check() {
   exit 1
 }
 
+write_deb_spec(){
+
+if [ ! -f ${DIRECTSAVEPATH}/DEBIAN/control ]; then
+return
+fi
+
+#pkgname|nameonly|version|pkgrelease|category|size|path|fullfilename|dependencies|description|
+
+pkgname="${DLPKG_NAME}"
+nameonly=`cat ${DIRECTSAVEPATH}/DEBIAN/control | grep "Package:" | cut -f 2 -d ':' | sed -e "s#^ ##g"`
+version=`cat ${DIRECTSAVEPATH}/DEBIAN/control | grep "Version:" | cut -f 2 -d ':' | sed -e "s#^ ##g" | cut -f 1 -d '-'`
+pkgrelease=`cat ${DIRECTSAVEPATH}/DEBIAN/control | grep "Version:" | cut -f 2 -d ':' | sed -e "s#^ ##g" | cut -f 2 -d '-'`
+category=""
+size="${SIZEK}K"
+path=""	
+fullfilename=${DLPKG_BASE}
+description=`cat ${DIRECTSAVEPATH}/DEBIAN/control | grep "Description:" | cut -f 2 -d ':' | sed -e "s#^ ##g" | head -1`
+dependencies=""
+depends=`cat ${DIRECTSAVEPATH}/DEBIAN/control | grep "Depends:" | cut -f 2 -d ':' | sed -e "s#^ ##g" -e "s#, # #g" -e "s#(#|#g" -e "s# |#|#g" -e "s# [0-9]##g"` 
+
+for dep1 in $depends
+do
+depname=`echo $dep1 | cut -f 1 -d '|'`
+ if [ "$dependencies" == "" ]; then
+ dependencies="+$depname"
+ else
+ dependencies="$dependencies,+$depname"
+ fi
+done 
+
+echo "$pkgname|$nameonly|$version|$pkgrelease|$category|$size|$path|$fullfilename|$dependencies|$description|" > ${DIRECTSAVEPATH}/pet.specs	
+	
+}
+
+write_arch_spec(){
+
+if [ ! -f ${DIRECTSAVEPATH}/.PKGINFO ]; then
+return
+fi
+
+pkgname="${DLPKG_NAME}"
+nameonly=`cat ${DIRECTSAVEPATH}/.PKGINFO | grep "pkgname =" | cut -f 2 -d '=' | sed -e "s#^ ##g"`
+version=`cat ${DIRECTSAVEPATH}/.PKGINFO | grep "pkgver =" | cut -f 2 -d '=' | sed -e "s#^ ##g" | cut -f 1 -d '-'`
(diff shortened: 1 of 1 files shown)


s243a:
This sounds important to me. BTW did this merge request use to have meta-data in the title?

mavrothal:
Not sure if I understand correctly the goal of this, but looks to me that the idea is that installing packages from other than the binary compat distro will try to do its best to get require dependencies if missing.
If indeed the case,  is a double edged sword because if the dependencies are compatible that makes life easier, if not may break the installation.
In general, (even though this is almost never done in puppy...) is nice when a feature is suggested to also have a use-case description so will be clear what the goal of the suggested mod is.

rizalmart:
The idea of this code was to create a proper installed package entry to be registered at "/root/.packages/user-installed-packages". After the non-pet packages was extracted its package information will be written as pet.specs based from the receipts or information file given by a package and PPM will do the rest

mavrothal:
> After the non-pet packages was extracted its package information will be written as pet.specs based from the receipts or information file given by a package and PPM will do the rest

So if this package is not  a pet and is not a package from the binary-compatible distro (otherwise it would be in Packages-$COMPAT_DISTRO-main or so) how do you go about dependencies?
PPM may do more harm than good in these cases or just fail.
If this is about "wild" packages distributed as deb, rpm etc but intended for a binary compatible distro should be some way to ensure that would be installed only in compatible puppies.

s243a:
> > After the non-pet packages was extracted its package information will be written as pet.specs based from the receipts or information file given by a package and PPM will do the rest
> 
> So if this package is not a pet and is not a package from the binary-compatible distro (otherwise it would be in Packages-$COMPAT_DISTRO-main or so) how do you go about dependencies?
> PPM may do more harm than good in these cases or just fail.
> If this is about "wild" packages distributed as deb, rpm etc but intended for a binary compatible distro should be some way to ensure that would be installed only in compatible puppies.

This is already an issue in puppy because you can install .deb files by just double clicking on them but if you click on one that is too new then it will fail due to ones core libraries (e.g. libc) being too old. Granted at least if the package is from a comparable distro then the packages will probably be named the same. To allow for different package naming, sc0ttman's package manager (i.e. pkg) allows one to setup an alias list.

rizalmart:
Some non-pet package format describes its dependencies on its package information. Just look at the code that I made.

mavrothal:
I have looked at the code and have no problem with it. I think what is doing may be problematic.
In puppy people do (try) to install a mix bug of packages either from the repos or from the forum or from anywhere. And that's fine! But at least the repo packages are checked to work and for the wild packages is up to the user or the provider  to make them work.
If the package manager _formally_ offers for example, the facility to install an rpm package in a debian build with all its dependencies, should offer some assurances that will work properly. To generate a pet.spec file and then try to install dependencies of a different distro or report that dependencies are missing but is impossible to find (because in the other distro have different names) is not very helpful.
In general is not a very good idea to mix packages from different tool chains (distros or distro versions). Historically the effort of PPM to do all kind of different things  has been a major source of criticism. To further facilitate the "you can  install whatever" idea I believe will only make things worse for the users.
For the developer or the advanced user all these "little problems" can be fun, but for the person that want to just use the machine to accomplish some work, not so much.
Anyway, that's what I think, and people can pull as they see fit.

peabee:
I agree with @mavrothal - I see no need for this extra complication!

wdlkmpx:
Nutyx and rpm distros are not supported so their pkg info doesn't matter, arch is also not supported.

Stuff is being simplified so that other concepts/build systems/architectures are easier to introduce/integrate.

I'm not sure what code will later use that info, all i know is that the PPM is horrible in many ways imaginable, and it's been simplified a bit in the experimental branch. Should the simplified ppm be merged right now, so that it gets some testing and fixes?

I think it should be completely replaced by Pkg and Pkg should probably be forked to simplify it and fix it a bit.

Some puppy scripts are way behind the times, i wonder if it's acceptable to just delete the code and exec some speficic app or tell the builder/user to just install such specific app..

mavrothal:
The problem is that woof code is changing but there is no one building and testing the changes to any extended level (peebee's rolling release has probably the same 2-3 people testing). There are maybe some more private builds that suit the taste of the builder/tester, and that's it.
Is time for someone to prime a new "official" puppy and get some betas out for (hopefully) extended testing.

wdlkmpx:
I do test, a lot. Not as much i used to, and it's easier to understand
the changes and possible fixes when there is less code to deal with.

Some apps are clearly too flawed to even bother looking, the ppm qualifies.

Back in the way it was really hard to debug stuff.. it was really hard
3 years ago, but now it's not that hard, but time is certainly running
out. Money is important.

Only ninaholic understands how this works, building and providing feedback.

With 20 ninaholics this would be way easier.

Automation in 3builddistro will land in the testing branch after i
test a build i just finished.

The same can be done for 0setup, but peebee will probably not approve.

In the experimental branch a fully build automated build is achieved
by running ./build.sh, some of it will be merged to testing, but not
everything yet.

Unless people understand and agree that it will make it easier for
everyone, then the whole thing can be safely merged.

mavrothal:
As I said "There are maybe some more private builds that suit the taste of the builder/tester"
My point is that regardless of how easy or difficult is to build a puppy, what is important is the actual iso to be used by non-experts and tested in all different scenarios.

Looks like few year back one or a couple people were building and everybody else was "happy", now everybody is building and not many are happy.
It would appear that the puppy focus has shifted from usability to "buildability"!...

Bottom line, Changes should be tested in the wild. Extensively. 
By everyday users that do _not_ understand what is happening. 
ie have a new relatively stable  (a la beta 1, beta 2, RC, final) official puppy out ASAP

wdlkmpx:
Puppy has never been (that) "usable", it was always filled with fatal flaws. 

And the amount of changes that happen do require many builds, it's about testing what is being changed, unfortunately puppy has a terrible design, so everything has been changing.

Buildability, yes, at least it's better than trying to sabotage real changes.

Keep in mind that 98% of the development comes from those who don't release (officially). 

It's basically an undeserved service.

wdlkmpx:
It's not hard to remember when xenialpup was unviable. It wasn't the official builder who fixed the issue. It was the compulsion the dismantle the packages-templates and the rewrite of several apps that made it viable, but it took too much time, and it remained buggy, because some fixes arrived later. (xkbconfigmanager). It was easy to fix them by building and releasing another iso.

It won't be the builder who will fix the issue with the latest 64 bit upup.

I read even BK has given up on ubuntu since quirky beaver was exposed for what it is.. unusable.

Just so you know where true development comes from..

mavrothal:
I'm sorry if you got the idea that  I criticised your work in any way.
I did not
I just say that the proof is in the pudding. 
Ie the released and widely used OS.

Issue #1386, opened 2019-05-12
Default kernels


wdlkmpx:
x86.. To completely automate a build in all compat distro versions there must be a default kernel specified in build.conf, 3 in fact ,this can be edited anytime.

3 kernels for each compat distro version, the builder should comment out the kernel versions they don't want to use.

Default = *

slacko 14.0 - all i686 kernels = nopae
- 3.2 (i686 only)
- 3.14
- 4.1 *

slacko 14.1 - all i686 kernels = nopae
- 3.14
- 4.1 *
- 4.9

slacko 14.2
- 4.1
- 4.9 *
- 4.19

dpup stretch
- 4.1
- 4.9 *
- 4.19

etc.

The ibiblio repo should add a new directory where the kernel sfs and sources are stored

kernels
kernels/slacko/14.0/x86
kernels/slacko/14.0/x86_64
kernels/slacko/14.1/x86
kernels/slacko/14.1/x86_64
kernels/slacko/14.2/x86
kernels/slacko/14.2x86_64
kernels/dpup/stretch/x86
etc

wdlkmpx:
Using the latest kernel LTS version might not always be a good idea. aufs is not maintained and issues may arise. Advice is needed, i think @ninaholic is aware of issues in some kernel versions

Many of the above-mentioned kernel versions have reached EOL, which is good because there's no need to update them. There's no hype. There are security issues, but running as root is a security issue. So it's all good i think.

Kernel 3.2 is special, its only purpose to compile very old drivers but as ninaholic notes, the firmware tarball might be too new for that kernel. And overall the firmware stuff is a mess. A more up-to-date tarball for recent kernels is also needed, i think.

There's also kernel 3.18, but i think 4.1 might be a better option..

I'll compile all the kernels, unless someone wants to stress their machine a little bit..

PR #1467, opened 2019-06-19
update 3.18 kernel to final

ghost:
this was the last one on kernel.org, now they are only updating it on android site it seems:
https://lkml.org/lkml/2019/5/16/679


ninaholic:
I tried 3.16.69 with kernel-kit and the aufs offsets and warning are still exactly the same as the build.log for 3.16.68 it looks like. The dotconfig generated seems to be the same as 3.16.43, unlike in 3.18.105 to 140 where there was differences, but I'll know for sure when it finishes compiling in about 4 hours.

ninaholic:
Here is a huge-kernel + sources for 3.18.140 compiled with slacko 14.0 / 5.8.3 32-bit with default i686 nopae if you want to use or test:

https://transfernow.net/42k5z7l55nds

It says the link will be good for about 6 days. :cake:

ninaholic:
And here is a huge-kernel + sources for 3.18.140 compiled with slacko 14.1 / 6.3.2 32-bit with default i686 nopae if you want to use or test: https://transfernow.net/24knx2n8g20i

It is slightly larger because I tried without -Xbcj x86 -b 512K this time. It was made with the 6.3.2 iso posted here https://github.com/puppylinux-woof-CE/woof-CE/issues/1468#issuecomment-504787944

wdlkmpx:
Maybe it needs a patch to compile without throwing warnings. 

 The change that breaks the aufs patches was pr0bably backported from some 4.x kernel, and it has been dealt with, but we don't how know.

I'm not able to test, but you might want to test in pupmodes 12/13 and see /var/log/messages, also make changes to random files from different locations and see if persistence is ok.

Once you confirm everything's ok, i'll m3rge this

ghost:
Hi wdlkmpx,

I just tried making a savefolder on ext4 partition in both 14.0 and 14.1 with PUPMODE=12 (pmedia=atahd) and PUPMODE=13 (pmedia=ataflash) with some changes and customizations and all of them seemed to stick upon reboot so far. In /var/log/messages I was warned that my ATOM CPU recommends microcode update (good to know) and saw this message upon reboot but not sure what it means:

```Jun 25 15:56:55 puppypc22785 kern.warn kernel: aufs au_new_inode:452:ROX-Filer[6931]: Warning: Un-notified UDBA or repeatedly renamed dir, b1, ext4, Xorg.0.log, hi1595107, i4230.```

Everything seems to behave just as good as 3.18.105 for me so far though (which was a config made from 3.18.22 originally). You can feel free to test anything you want with it too though, and add it to your kernel project here if you think it's good: https://github.com/puppylinux-woof-CE/woof-CE/issues/1386 I might have a 3.14.79 kernel made with kernel-kit lying around somewhere too but not sure if it was made with 14.0, 14.1, or 14.2 yet, if that might be of use to you. :smile_cat:

Unrelated: Sometimes upon shutdown if I have been doing a lot of things, I notice the umount -a on shutdown can be slow, waiting up to around 30 seconds before things get unmounted, I think it might be because I have a swap partition and it's clearing it all too though. Also I notice that in 14.0, the first time I right-click on something in woodenshoe's Rox it usually doesn't register the action (in PUPMODE=5) but works the 2nd time and every time after (the old Rox did not have that issue).

woodenshoe-wi:
> Also I notice that in 14.0, the first time I right-click on something in woodenshoe's Rox it usually doesn't register the action (in PUPMODE=5) but works the 2nd time and every time after (the old Rox did not have that issue).

I have noticed that myself, but I don’t remember making any changes to the action handling code.  Only adding some menu entries and changing some shortcuts.  If I did, I have since forgot.

I did merge some updates that I found on other people’s forks of ROX-Filer that changed from old deprecated functions to the new equivalent, and at least in one case the new function didn’t behave exactly the same.

I will try to remember to take a look at it if nobody beats me to it, but I don’t have time to now.  Maybe later this summer or in January.

PR #1471, opened 2019-06-25
rc.shutdown: Improved cleanup upon shutdown

rizalmart:
Improve further cleanup to save more disk space. Cleanup will be deployed if the save file is used upon shutdown

### DIFF
--- a/woof-code/rootfs-skeleton/etc/rc.d/rc.shutdown
+++ b/woof-code/rootfs-skeleton/etc/rc.d/rc.shutdown
@@ -46,6 +46,162 @@ killzombies() { #zombie process prevents partition unmount.
  done
 }
 
+purge_user(){
+xhome="$1"	
+rm -rf $xhome/tmp 2> /dev/null #...note, not screening this out in any of above save modes.
+rm -rf $xhome/.thumbnails/* 2> /dev/null
+[ -d $xhome/.config/geany ] && rm -f $xhome/.config/geany/geany_socket* 2>/dev/null	
+}
+
+purge_user_cache(){
+xhome="$1"	
+for c1 in thumbnails midori google-chrome opera mozilla "moonchild productions" chromium slimjet iron slimjet brave maxthon waterfox vivaldi qupzilla falkon yandex qtweb gnome-web netsurf galeon xombrero kmeleon epiphany vlc wine rhythmbox media-art totem
+do
+ [ -d "$xhome/.cache/$c1" ] && rm -rf "$xhome/.cache/$c1"/* 2>/dev/null
+done
+
+[ -d $xhome/.config/qupzilla/tmp ] && rm -rf $xhome/.config/qupzilla/tmp/*
+
+[ -d $xhome/.gnome2/epiphany/favicon_cache ] && rm -rf $xhome/.gnome2/epiphany/favicon_cache/*
+[ -d $xhome/.gnome2/epiphany/mozilla/epiphany/Cache ] && rm -rf $xhome/.gnome2/epiphany/mozilla/epiphany/Cache/*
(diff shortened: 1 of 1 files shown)


wdlkmpx:
This doesn't look right.

Rc.shutdown only deletes lock files and clears global tmp dirs.

Currently it even backups /var/log/messages before deleting it.

This deletes too much stuff and will upset meany people.

What makes you think people would want to get rid of  amule's... unfinished downloads? That cleanup doesn't belong in rc.Shutdown but in some "cleanup script",  with a menu entry perhaps, "pup CCleaner" hahaha

rizalmart:
You may trim this code and choose what unnecessary files to delete

mavrothal:
Indeed too many things are deleted, however I can see the value in it and it could become user configurable. Should be called on the first shutdown after the savefile is implemented, and them preferably with a reminder on consecutive shutdown, ie "shutting down. 'So and so' will not be saved. Click here if you want to save them" or something.

s243a:
The one that annoys me is the ".trash folder", When I delete things it usually means that I wanted it deleted and it took me a while to realize that stuff wasn't being properly deleted. Worse yet, if you delete stuff in the .trash folder it will just be saved back to the .trash folder unless you change your file manager settings. 

I know that some puppy users like this .trash folder thing but in many puppys these files don't even show up when you click on the recycling bin icon.

Butterfly:
> The one that annoys me is the ".trash folder", When I delete things it usually means that I wanted it deleted and it took me a while to realize that stuff wasn't being properly deleted. Worse yet, if you delete stuff in the .trash folder it will just be saved back to the .trash folder unless you change your file manager settings.
> 
> I know that some puppy users like this .trash folder thing but in many puppys these files don't even show up when you click on the recycling bin icon.

Hi from a lurking, very happy fan! I fall into the "likes the trash bin" category. In fact, I like the trash can to hang with me, to stay full until I manually delete its contents. It's a cognitive thing where that's a last step self-protection perk in case I accidentally delete something I shouldn't.

So I'm just writing to say that it would be nice if there was a *CHOICE*, a toggle on and off option, to delete or not delete to help both @s243a and my personal needs. It sounds like @mavrothal is describing almost exactly what I'm imagining, right down to where it verifies that the User doesn't have any special needs for **each and every** given session.

**Tiny caveat...** I could see someone quickly becoming _quite agitated_ at being slowed down by that question **each and every** time if they never, ever change their mind so _then what..._? A secondary "ask me/don't ask every session" option, as well? Or a list of 3 *CHOICES* for yes, no, and "to be safe, you better ask me every time"? :thinking:  :smile:

s243a:
> , to stay full until I manually delete its contents. It's a cognitive thing where that's a last step self-protection perk in case I accidentally delete something I shouldn't.
> 
> So I'm just writing to say that it would be nice if there was a _C_

As a side note, I also find those, "are you sure you want to shut down prompts" annoying. The main reason is if the system is close to freezing it can make shutting down very time consuming but even when the system isn't close to freezing it is nice to be able to just tell the system to shut down and not have to worry about any prompts, especially if one needs to get to bed quickly. 

Any like what was said above, if these are all configuration settings then everyone can be happy...as long as they know about them.

bigpup1:
I am not sure what trash you are talking about, but it is not the Puppy trash that is on the desktop.
Puppy trash is a completely separate program from the file manager. 
You are probably talking about trash in some file manager that is not Rox file manager.
Rox does not have a trash.
All you can do in Rox is add a right click menu entry to put stuff in the Puppy trash.
In Rox, if you choose delete, it is deleted.

A lot of file managers have their own version of trash built into the program.

wdlkmpx:
This can be safely closed. 

A special mode for tinfoil hats should be supported where snapmergepuppy is disabled, maybe a new kernel param: `tinfoilhat` or `pfix=tinfoilhat`

PR #1472, opened 2019-06-25
savesession-dvd: cleanup before burning to disc

rizalmart:
This will save free disc space


s243a:
I think that these cleanup actions are most important in multi-session CD/DVD mode, since it is very easy to fill up a multi-session cd.

PR #1474, opened 2019-06-25
save2flash: Perform cleanup before saving to flash

rizalmart:


mavrothal:
Couldn't this as well as #1473, 2 and 1 combined into a script called by rc.shutdown as needed, and be user configurable?

PR #1476, opened 2019-06-27
petget: Fix TREE2 to retain full package version when no extension

rserwin1:


mavrothal:
Could you give an example of such a package ?
Is this for a package that will have a number after the dot instead of an extension or separates package version with a dot instead of a dash?
You may not want to install a package with no extension as petget may have issues with it, as it assumes is a pet.

wdlkmpx:
There is a problem with petget, petget itself, but there's another problem.

When i type `petget zzz-1.0` there is a confirmation dialog
```
Do you want to install this package: zzz-1.0.pet
```
I didn't specify `.pet` and the file doesn't exist, zzz-1.0 doesn't exist either.

When i type `petget -zzz-1.0` i see
```
Do you want to uninstall this package: zzz-1.0.pet
```

As you can see, you can specify phantom packages and petget will use some phantom logic

Logic that needs to be revised and probably rewritten

wdlkmpx:
Actually what i see is:

petget zzz-1.0 / petget -zzz-1.0
```
do you want to install/uninstall: zzz-1
```

After applying the patch
```
do you want to install/uninstall: zzz-1.0
```

It's possible that some apps may call petget without .pet, sfs_load was an example.. the offending lines were removed (sfs_load).

This does fix an issue, but as i said, petget needs a rewrite probably.

rserwin1:
From the beginning of script petget:
#Note: Puppy Package Manager is started from 
/usr/local/petget/pkg_chooser.sh for the GUI.
#'petget' is for commandline execution, especially from ROX-Filer.
#If passed param is a pkg name preceded by a + to install pkg, - to 
uninstall pkg.
# example: +xbubble-0.2.4
#For integration with Rox-Filer, commandline is a .pet|.deb|.tgz|.tar,gz 
with absolute path, install.
# ex: petget /root/xbubble-0.2.4.pet
#For integration with browser, ex: petget 
http://somewhere.on.internet/xbubble-0.2.4.pet
*#Note, petget also get called from /etc/rc.d/rc.update to remove 
packages. ex:**
**# petget -xbubble-0.2.4*

That is what I need to do in a script I am adapting (from Lucid Pup) to 
clean up user-installed packages and files that are replaced by or 
conflict with versions in a substituted distro or remaster.

rserwin1:
Thank you.  What do you think about making petget into its own package, 
in case someone eventually wants to replace it with pkg and its probable 
GUI?

wdlkmpx wrote:

rserwin1:
Per wdlkmpx:
As you can see, you can specify phantom packages and petget will use 
some phantom logic

I take it that you mean that petget should verify the intended package 
exists before issuing the prompts.  I could work that and probably other 
such fixes short of a rewrite, if that would be acceptable.

I think the original intent was to assume a default extension of .pet.

wdlkmpx:
It's ok to fix to petget, just see it as a casual user, what it shouldn't do, when it should/should not fail, etc. Always trying to detect obsolete/dubious code/comments to be deleted

extrasfsfind used to find .pet files and call petget as well, i heavily edited that script

PR #1479, opened 2019-07-02
pmount: changed mountpoint from /mnt to /media

rizalmart:


mavrothal:
Is there any reason for this change? 
Puppy mounts things in /mnt for ever and  many scripts hard-code /mnt/* .

And once more I would ask to provide some more info in the commit message or at least the pull requests.

rizalmart:
because due to nature of udisks2-volume-manager and pup-volume-manager which shows mounted drive on /media rather on /mnt folder unless there is an alternative solution

mavrothal:
Which system is that on? 
Any chances that sustemd-depended/originating udev is used?
You may also want to look at /etc/udev/rules.d/ and /usr/lib/udev/rules.d/ to tame the system.
Is #1478 also related to this?

PR #1480, opened 2019-07-05
rc.services: delete non-running pid files first

rizalmart:
Remove pid files which contains pid numbers that is not running at all. In order to start the services accurately. Those pid files does not remove if improper shutdown occured


wdlkmpx:
Commit https://github.com/puppylinux-woof-CE/woof-CE/commit/d919cedc03160b074c2b1d669253f4cee394ffab merges this PR

Issue #1481, opened 2019-07-05
Quicksetup in kms/modesetting


ghost:
On one of my netbooks, I type xorgwizard and select kms/modesetting video driver to see the desktop. After I do this recently, quicksetup first-run dialog does not pop-up any more and Puppy immediately barks at me instead :dog2: Running quicksetup from terminal gives an error about gtkdialog (attached below). This error started after experimental branch was made, maybe around commit 7000-7200 but I'm not sure exactly from which commit in that range yet. Reverting /usr/sbin/quicksetup itself did not fix anything so I am guessing it happens because of something in the underlying system, maybe recent busybox update or some update in part of gtkdialog itself? I don't think 1019 on quicksetup is the problematic line because if I remove that section it just gives me an error somewhere else, the line 1019 error message is slightly cryptic to me :laughing: It happens in both slacko 14.0 and 14.1 but I have not tried 14.2 or dpup yet, and don't know if it's a universal problem introduced or just on specific computers when using kms/modesetting.

I made a current slacko 14.0 iso here if you want to use it for testing:
https://transfernow.net/4978d1x06ca3
md5sum: 3d6047894dcd72c3751a9c0417252ebc

![quicksetup](https://user-images.githubusercontent.com/5408521/60736038-a3a9d980-9f23-11e9-9b1b-9912a75f7875.png)

Also I notice that when using Desktop Icon Switcher to change icon theme it does not refresh mounted or home drive icons (think 01micko might have mentioned something about this, might relate to the new cairo or theme changes).

![notrefreshed](https://user-images.githubusercontent.com/5408521/60736068-b91f0380-9f23-11e9-8bab-26497329a209.png)

Also noticed that the first dialog for slackware Gparted is missing the icon, which was easy to fix though by adding this to 00build.conf:

`ln -s /usr/share/icons/hicolor/32x32/apps/gparted.png /usr/share/pixmaps/gparted.png`

![gparted](https://user-images.githubusercontent.com/5408521/60736087-c50ac580-9f23-11e9-88b8-c2884264b01b.png)

With the new ram system it was a challenge to find where /mnt/sda4 was using PUPMODE=5 in Rox (the drive where puppy was loaded from) but I did end up finding my sda4 drive eventually in /initrd/mnt/dev_save instead :balloon: 

[...]

PR #1483, opened 2019-07-06
fix gparted icon on first screen #1481

ghost:


01micko:
Probably not the right way to go about this. Sure, fixes it for slacko based on 14.0 but what about your dpup spin mentioned in #1481 ?
I'll await further review / ideas before merging, but I think a dedicated icon in pixmaps/puppy might be a better option.

ghost:
Hi 01micko!

You're right, I checked in the dpup build and it also has the missing icon, and the ln -s line fixed it there too. Maybe a fallback in package-templates for "if no /usr/share/pixmaps/gparted.png exists but /usr/share/icons/hicolor/32x32/apps/gparted.png does, create a symlink" would work, and/or an edit to the petbuild, and/or a dedicated icon for it in rootfs-skeleton or the theme like you said.

Screenshot

edit: I guess the package-templates idea wouldn't work for 14.0 because in this one it actually seems to be a pet, in dpup stretch it comes from a deb from the dpup repository though.

ghost:
Here's the original icon for both though if anyone wants it :woman_artist: 

gparted

01micko:
I'm pretty sure we could make a nice icon that fits the pixmap/puppy design but that may not be the solution. Is the dpup version a pet or deb?  Maybe more options are needed in `gparted_shell`

relevant snippet, with a bug.. and maybe the problem?

```shell
[ -f /usr/share/pixmaps/gparted.xpm ] & gpicon=/usr/share/pixmaps/gparted.xpm
[ -f /usr/share/pixmaps/gparted.png ] & gpicon=/usr/share/pixmaps/gparted.png
[ -z $gpicon ] && gpicon=/usr/share/pixmaps/puppy/harddisk_scsi.svg
```

Unquoted variable in test is the bug. Result will be undefined.

ghost:
Interesting! I did not realize it was script (I thought icon was hardcoded in binary). :laughing:

I tried changing & to && and changing -z to ! and it seems to work now for both cases?

```
[ -f /usr/share/pixmaps/gparted.xpm ] && gpicon=/usr/share/pixmaps/gparted.xpm
[ -f /usr/share/pixmaps/gparted.png ] && gpicon=/usr/share/pixmaps/gparted.png
[ ! $gpicon ] && gpicon=/usr/share/pixmaps/puppy/harddisk_scsi.svg
```

ghost:
Basically, with this new code, it picks gparted.png if you have it, if not it falls back to gparted.xpm, and if both don't exist it falls back to harddisk_scsi.svg. I've tested all combinations I could think of and it worked as expected. With the old code, dpup had both gparted.xpm and harddisk_scsi.svg and it didn't pick any of them. I think this is good enough to merge, but feel free to review and test if you like ;)

01micko:
Looks ok to me :+1:

PR #1485, opened 2019-07-08
update wikka links to wikka.puppylinux.com

ghost:
there is also still some references to puppylinux.org in index.html, release-skeleton.bottom.htm and legal/puppy.htm but not sure if those are as noticeable and what should happen there, I just noticed these two when clicking them :smile: 

the blog post about the change: http://blog.puppylinux.com/new-address-for-the-wikka


01micko:
Feel free to fix _any_ other broken links.

PR #1493, opened 2019-07-14
busybox 1.31.0 and initrd_progs-20190714

wdlkmpx:
Before this PR can be merged

This file:
- http://www.01micko.com/wdlkmpx/woof-CE/busybox-1.31.0-aarch64_musl_static.pet
should be uploaded to
-

This file:
- http://www.01micko.com/wdlkmpx/woof-CE/busybox-1.31.0-arm_musl_static.pet
should be uploaded to
- http://distro.ibiblio.org/puppylinux/arm/pet_packages-armv6/

This file:
- http://www.01micko.com/wdlkmpx/woof-CE/busybox-1.31.0-x86_64_musl_static.pet
should be uploaded to
- http://distro.ibiblio.org/puppylinux/pet_packages-common64/

This file:
- http://www.01micko.com/wdlkmpx/woof-CE/busybox-1.31.0-x86_musl_static.pet
should be uploaded to
- http://distro.ibiblio.org/puppylinux/pet_packages-common32/

### DIFF
--- a/initrd-progs/0initrd/etc/rc.d/functions_x
+++ b/initrd-progs/0initrd/etc/rc.d/functions_x
@@ -23,10 +23,10 @@ fx_format_bytes() {
 	local BYTES=$1
 	local FORMATTEDSIZE="${BYTES}B" # bytes
 	if [ ${BYTES} -ge 1099511627776 ] ; then
-		FORMATTEDSIZE="`dc $BYTES 1099511627776 \/ p`"
+		FORMATTEDSIZE="`dc -e "$BYTES 1099511627776 / p"`"
 		FORMATTEDSIZE="`LANG=C printf "%.2f" $FORMATTEDSIZE`T" #terabytes
 	elif [ ${BYTES} -ge 1073741824 ] ; then
-		FORMATTEDSIZE="`dc $BYTES 1073741824 \/ p`"
+		FORMATTEDSIZE="`dc -e "$BYTES 1073741824 / p"`"
 		FORMATTEDSIZE="`LANG=C printf "%.1f" $FORMATTEDSIZE`G" # gigabytes
 	elif [ ${BYTES} -ge 1048576 ] ; then
 		FORMATTEDSIZE="$(( $BYTES / 1048576 ))M" # megabytes
--- a/initrd-progs/build.conf
+++ b/initrd-progs/build.conf
@@ -5,7 +5,6 @@
 INITRD_PROGS='
 busybox
 e2fsck
-grep
 #mp
 ntfs-3g
 resize2fs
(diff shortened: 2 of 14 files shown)


peabee:
3 files with specified destinations uploaded to ibiblio (not aarch64)

01micko:
@peabee Alright, no aarch64 repo yet.. We can make it one day. I suppose that will be the future. 

@wdlkmpx However.. since these busybox builds are a part of initrd_progs why can't we just merge them into the distro when initrd_progs is downloaded?

Anyway, merging.

01micko:
Bugger. Found the division bug.. need the stupid `2 k` with dc before '/' for 2 decimal places `3 k` for 3 and so on. Will fix and push shortly

wdlkmpx:
I think it worked fine in my brief test, that was a change i made 2 weeks ago, but didn't pay too much attention.

The busybox.pet is different from the initrd busybox, but both are compiled with my up-to-date musl cross-compilers.

My cross-compilers are updated every 6 or 12 months, so when i compile the same binaries, they may be faster/more reliable, as musl is being optimized/fixed. I can add more cross compilers for more architectures.

In fact i can compile many static C apps with initrd_progs.. for all the supported architectures.

This is what i now have in mind... woof-arch2 - a tarball, which in fact is superseded by woof-arch - the directory.

So woof-arch could allow more freedom, fixed compiled binaries by anyone... woof-arch2.tarball could be the base.

So with some logic, merge2out/setup.sh could download woof-arch2.tarball extract it, and overwrite binaries with the binaries from the git repo woof-arch.

Common32/common64 contain packages that run in any woofce x86 pup [starting from slacko 14.0], so if you don't feel like compiling stuff for zwn, you can just unpack the .pet packages and create a massive tarball, that's it. 

But i also see slacko 14.0 is too old for some packages (udev, etc), but it's the right base for everything else, gtk 2.24, the minimun version supported.

For arm, armv6 contain some generic packages, more generic packages con be found in raspup official pkg db.

wdlkmpx:
The new dc introduces an incompatibility, but if a busybox applet changes and behaves like the full app.. it's time to change the code.

Most busybox applets behave like the full apps, with limited functionality (missing features). There are a few exceptions, and dc was one of them. `busybox ps` is probably the biggest offender ever, a mistake.

For noarch scripts in the repo, i think very few use dc (and it probably doesn't break them (completely)), what can be done, identify them, fix them, increase version, reupload them. Gotta catch 'em all

I performed a search in my non woof-ce scripts (700kb), and i don't use dc in any of them, what a relief. I think people tend to use `bc` instead. It was reading puppy code how i became aware of `dc`.

I also think nothing is lost if some puppy gui scripts are just deleted and replaced with some advanced and lightweight compiled gtk apps [generic builds].

xrandrshell seems to be broken using k4.1 intel driver in an old machine, needs some delay or something, but in that case it's better to just delete it and use lxrandr (my future version) or zarfy. My lxrandr version has fixes and more features, need to push the changes. It will be better than zarfy.

lxrandr is shell script written in C, with enough time it can be rewritten using bash + gtkdialog implementing the same features, but it's too late for that.

PR #1498, opened 2019-07-16
sfs_load: some improvements

rizalmart:
Added GIO module cache update
Improving pango, gdk-pixbuf, and gtk-immodules update cache


wdlkmpx:
This looks ok to me. I see sfs_load needs some optimizations, less calls to external commands.

I think gtk1 is completely irrelevant, but i there are ancient plugins for epsxe (gtk2) for linux that use gtk1.. it's a pity really.

http://www.pbernert.com/html/gpu.htm

Other than that, if you use any other Play Station 1 emulator there are no gtk1 apps to support ... at all. I feel empty, i need to buy a new gamepad

PR #1502, opened 2019-07-21
use firefox dir instead of firefox* if found first

ghost:
Newer versions of firefox seem to look for local-settings.js in "firefox" instead of "firefox-68.0.1" (for example) yet contain both directories. I added a check for "firefox" first, and if it doesn't exist, fall back to "firefox-versionnumber" (so it doesn't store 2 directories in the variable and pick the wrong one). Maybe there is a better way to do this but it seems to work for me so far in Slacko64 (the puppy homepage still will not appear until 2nd time opening it because it shows Firefox welcome message first time though).

Screenshot to show examples of the code and what it does in Firefox 68.0.1:
https://ibb.co/6X9JhDW

Screenshot(1)


ghost:
I meant to add the spot.png one as a separate commit but seemed to have merged it into this one.

Before:
Screenshot

After:
Screenshot(1)

PR #1513, opened 2019-07-24
Update DISTRO_PKGS_SPECS-ubuntu-bionic

rizalmart:
Added desktop-file-utils

### DIFF
--- a/woof-distro/x86/ubuntu/upupbb/DISTRO_PKGS_SPECS-ubuntu-bionic
+++ b/woof-distro/x86/ubuntu/upupbb/DISTRO_PKGS_SPECS-ubuntu-bionic
@@ -109,6 +109,7 @@ yes|desk_icon_theme_jq8raised||exe
 yes|desk_icon_theme_moblin_stark||exe
 yes|desk_icon_theme_Moka||exe
 yes|desk_icon_theme_neon||exe
+yes|desktop-file-utils|desktop-file-utils|exe,dev,doc,nls
 yes|devmapper|libdevmapper1.02.1,libdevmapper-dev,libdevmapper-event1.02.1|exe,dev,doc,nls
 yes|dhcpcd||exe,dev,doc,nls|
 yes|dialog|dialog|exe,dev>null,doc,nls


peabee:
What are these utils and why are you adding them?? What problem are you trying to solve???

peabee:
Are they compatible with the Puppy menu structure???

rizalmart:
desktop-file-utils is a vital component to any xdg-compliant file manager like pcmanfm, thunar, etc for file associations. Without that, file associations for new added package or loaded sfs modules will just simply dont work on those xdg-compliant file management. One of the flaws in Puppy that needed to be fix. XenialPup has desktop-file-utils by default  but it seems not yet utilize in Puppy's Package Management in its peak release. It was lately implemented when I start to contribute here

mavrothal:
desktop-file-utils is included in Phil's pups. However, without a call from PPM or SFS_load is not going to be very useful.
You might want to do a proper pull and put all spec changes together and also add call to ```update-desktop-database```  in places where ```fixmenus``` is called. Preferably only when other than the ROX-filer file manager is also running.

rizalmart:
mavrothal my pull request about calling update-desktop-database on PPM and sfs_load is already merged in woof-ce (take a look on the commit history). One of the downside of the ROX-Filer is that rox is not xdg-compliant, instead it has own file association management. I wonder if woodenshoe-wi can modify ROX-Filer and make it xdg-desktop compliant

wdlkmpx:
I remember mimeinfo.cache was a problematic file while using pcmanfm and my list of file associations

update-desktop-database creates  $XDG_DATA_DIRS/applications/mimeinfo.cache 

I have this line in a script
rm -fv ${SYSROOT}/usr/share/applications/mimeinfo.cache #this file is always trouble

https://bbs.archlinux.org/viewtopic.php?id=97886

rizalmart:
mimeinfo.cache its just a raw database of file associations. But it can be easily override by placing custom default file association in mimeapps.list in $HOME/.config or in $HOME/.local/share/applications it will read it first before mimeinfo.cache for searching default file association

wdlkmpx:
I think i tried to create d4fault file associations for all users in /usr/share/applications/mimeapps.list, but with mimeinfo.cache present it ignored everything

rizalmart:
@wdlkmpx you can also try /usr/share/applications/defaults.list

wdlkmpx:
Mimeapps.list is the file that worked fine with pcmanfm,for default
applications and file associations.

This will certainly break some code and configurations, have to change
paths, i have an extensive list of file associations. this wilL probably
break pmime

rizalmart:
what paths are needed to change? and what files that are affected?

PR #1515, opened 2019-07-31
pmime: some changes

rizalmart:
1. Remove default.list and use mimeapps.list instead
2. Dont remove mimeinfo.cache
3. Regenerate mimeinfo.cache if it was missing


wdlkmpx:
Do you understand the code you're editing?

I see defaults.list is deleted, it's just a temp file. Its contents are added to my mimeapps.list before that it's deleted. You're also preventing /etc/mailcap from being populated with some wce defaults.

You're breaking stuff.

There's a problem here, you're trying to change stuff in a system you don;t use (apparently)

PR #1523, opened 2019-08-19
xwin: Improved support for uefi mode

rizalmart:
When puppy boots from UEFI. Use fbdev as fallback driver instead of vesa. VESA only works on bios only due to 10H interrupt which was not present on UEFI


wdlkmpx:
It's not exactly a fallback driver because that line is commented out if the specified driver is not found in the xorg drivers directory.

The script determines whether xorg.conf specifies a driver, if the driver does not exist.. the line is commented out.

This:
```
	Driver      "intelzz" #card0driver
```

Becomes:
```
#	Driver      "vesa" #card0driver
```

And it makes sense i guess.. if the driver specified in xorg.conf does not exist, xwin reverts Xorg to a more automatic behavior.. 

In the legacy branch that line becomes: `	Driver      "vesa" #card0driver` ( new driver in use)

The current behavior was implemented in Oct 7, 2016 by https://github.com/puppylinux-woof-CE/woof-CE/commit/3ee36fb9bd25bdac3871282870235bc675557857

Before that commit the fallback driver became the new driver, but since oct 7, 2016 that no longer is the case..

PR #1530, opened 2019-09-09
Update drive icon code so boot parts are distinguished from save part…

### LINKED ISSUE #1491: Maybe we need a new overlayed icon
On the pi, from commit f1ffba2db8897bc83bb26da57eb90166beacee4c it no longer boots into ram.. no problem, probably better that way, however on most pi's running puppy in RAM they are set up on 2 partitions so I get 2 'home' partition icons. No big deal except for aesthetics.

Maybe, if we could have a new overlayed icon with something other than the home icon or eject icon. I did knock up a quick overlay icon .. I called it boot_outline48.png

boot_outline48

Hmm.. that's a bit hard to see..

usb48
card48

01micko:
…s. closes #1491

See #1491 for explanation and screnshots.


wdlkmpx:
It seems to work fine:
znocopy48

I created a pupsave in another partition, but the psavemark stuff was broken in the initrd, this fixed it for me:
16bd87462e77ec1f403da91bc9143142ed55d481

Now I want to talk about something really important, today was one of those days when I needed to listen to something depressing to brighten up my day, with sound coming from the speakers.

I heard 6 barks, booting in pupmode 5  because of the psavemark bug in the initrd, until I fixed it.

So I was checking some Evanescence songs, not really convincing, but there is one song that stands out, the best by far, with a mermaid's calling luring men to their death, male and female vocals blending into some really dark, and some religious themes.

https://www.youtube.com/watch?v=ET3mhIfkCXA

There are people here whose ears are very sensitive to inhuman growls, screams, barks and cats fighting, so I muted only the screamo part, locally of course.

But 6 barks were heard and that's when I understood the lyrics were about puppy:

```
You will never be strong enough
You will never be good enough
You were never conceived in love
You will not rise above
```

All because of 2barks.au

PR #1535, opened 2019-09-15
xorgwizard: some improvements

rizalmart:
* Add all screen resolutions up to 8K for future proofing
* Use comboboxtext instead of combobox in dialogs to avoid accidental editing of predefined selections

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/xorgwizard
+++ b/woof-code/rootfs-skeleton/usr/sbin/xorgwizard
@@ -88,16 +88,40 @@ function xorg_screen_res_dlg_stuff() {
 		DEFAULT_RES=$(report-video res)
 	fi
 	#- screen resolutions (edit this)
-	RESOLUTIONS="1024x768
-800x600
+	RESOLUTIONS="800x600
+1024x600
+1024x768
+1152x864
+1154x768
+1280x720
+1280x768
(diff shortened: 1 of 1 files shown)


wdlkmpx:
I don't agree with using `comboboxtext`, (power) users can no longer choose to use some other value.

Otherwise all possible resolutions, bit depths and DPI's should be provided in the comboboxtext.

wdlkmpx:
However I think it's a good idea to add a few more screen resolutions,
maybe something like this:

1024x600
1280x720
1280x768
1440x960
1440x1080
1600x900	
1600x1200
2048x1080
3440x1440
4096x2160
5120x2880
8192x4320

(a commit will follow)

You need huge DPI's for 4k and 8k.. probably 180-400

I don't know how feasible is to reimplement the dccprobe stuff for x86
builds (uname -m = i686/x86_64), but ddcprobe sometimes hangs, that's
the problem.

If only xrandr reported all the supported screen resolutions, that
would be the only proper solution.

I see this info for my old 15'' flatron monitor (2004? -
https://www.newegg.com/lg-l1530s-15-xga/p/N82E16824005029 )

Screen 0: minimum 8 x 8, current 1024 x 768, maximum 32767 x 32767

32767 doesn't make sense

PR #1538, opened 2019-09-16
remasterpup2: some improvements

rizalmart:
Adding fail safe code for converting package .files into builtin files


wdlkmpx:
I haven't tested this but the change looks good to me.

PR #1540, opened 2019-09-16
Create update-desktop-database placeholder script

rizalmart:
To ensure the file associations on xdg-compliant file managers works upon installing packages or loading sfs modules even desktop-file-utils was not installed (due to package size concerns). A placeholder script will be used. It works exactly the same as the original update-desktop-database binary executable.

### DIFF
--- a/woof-code/rootfs-skeleton/usr/bin/update-desktop-database
+++ b/woof-code/rootfs-skeleton/usr/bin/update-desktop-database
@@ -1,0 +1,52 @@
+#!/bin/bash
+#This was a place holder in case the desktop-file-utils is not installed
+#It generates xdg file association database. It works the same as the binary update-desktop-database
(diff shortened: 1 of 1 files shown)


wdlkmpx:
After bafb90680acf24c2fb0dca226f3fe8151a2c9919 desktop-file-utils might not be needed. This is the latest version of the file:
https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-code/rootfs-skeleton/usr/bin/update-desktop-database

It works fine and is fast enough, although the compiled binary a few times faster

They say awk produces even faster results than pure shell code, but I'm yet to learn it, I found that properly optimized bash code is really fast, fast enough in some cases. But it's never too late to learn awk, providing that your brain is in good health unlike mine

PR #1543, opened 2019-09-17
poweroff: Adding non-acpi support

rizalmart:
Instead of just a frozen screen if acpi is disabled or not supported, just show a message that it was safe to turn off the computer

### DIFF
--- a/woof-code/rootfs-skeleton/sbin/poweroff
+++ b/woof-code/rootfs-skeleton/sbin/poweroff
@@ -11,7 +11,14 @@ done
 
 script=${0##*/}
 
+if [ -d /proc/acpi ] || [ -d /proc/apm ]; then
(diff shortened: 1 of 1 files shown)


wdlkmpx:
This works as advertised and there's a button to reboot the PC.

However, I think acpi is an x86 thing. My raspi died and I haven't bought another one but I guess `/proc/acpi` and `/proc/acpi` are not present in a raspi.

But here's the thing, the raspi doesn't really shutdown you have to unplug it.

So this behavior is ok even in a raspi, but message should be different: "You can now unplug the PC"

This is being merged and @01micko might want to test this in a raspi.

01micko:
No problem on pi zero. Just shows "safe to turn off computer" which is unambiguous.

PR #1544, opened 2019-09-18
Create sync-rox-icons

rizalmart:
This script syncs the rox mimetype icons with mimetypes on /usr/share/icons/hicolor
An attempt to make rox-filer xdg-desktop compliant


wdlkmpx:
This script is probably not needed..

This is what I see:
- you can change the icon theme in Option -> Types -> Icon theme
- rox already uses hicolor to display icons that are missing in the "current" or "selected" icon theme

wdlkmpx:
There are several rox versions out there. The old one, jun's, other versions.. the version used and enforced in woofce is rox-woodenshoe. 

And woofce keeps changing, if you're testing with an old iso,  it probably contains issues that are already fixed.

PR #1559, opened 2019-09-27
remove raspup stretch (superseded by raspup buster)

wdlkmpx:
raspup buster works in raspberry zer0, 1, 2, 3 & 4
and also has many improvements and packages

and raspup is still somewhat unknown, so getting of stretch

so far it's only about supporting the raspis
and raspup buster provides even better support


01micko:
For historical and educational purposes I'd leave it, or at least merge testing >> master first.

I learned a lot from the work of @woodenshoe-wi and it still may be a more viable alternative for pi 1 due to smaller size.

01micko:
Ok, I've forked woof at latest commit and here: https://github.com/01micko/woof-CE-raspi-stretch-archive so I'll merge this.

wdlkmpx:
The only that's missing is probably the kernel sources sfs

I don't use a mouse most of the time, and the raspi 1 was unusable with xorg mouse keys...... in very old desktop pc's it's reasonably fast.

I guess the raspi 1 and zer0 qualify as cli-only and/or server boards. pfix=nox

Now, slackware doesn't qualify for a full GUI puppy, although I see it's adding more and more packages in slackware current (i.e. speex, lame, etc)

But it's good if you want to recompile/replace many things (happening in slacko14.0/14.1).. and it's the most compatible and manegeable compat distro.

And musl should be the raspbian's libc.

PR #1602, opened 2019-10-03
menu_build_recent_docs improvement for long file+path

peabee:
[quote="B.K. Johnson"]
Note that this operates differently from the OOTB version in 2 ways.
1. Only the file names are displayed, not paths.
2. When a previously listed file is renamed, deleted or moved, it is excluded from subsequent lists.
[/quote]

### DIFF
--- a/woof-code/rootfs-packages/jwm_config/usr/local/jwm_config/menu_build_recent_docs
+++ b/woof-code/rootfs-packages/jwm_config/usr/local/jwm_config/menu_build_recent_docs
@@ -4,16 +4,17 @@
 export TEXTDOMAIN=jwm
 export OUTPUT_CHARSET=UTF-8
 
-XBEL_PATH=/root/.local/share/recently-used.xbel
-MAX_ITEMS=20
-
-tail -n $((MAX_ITEMS*13)) $XBEL_PATH > $HOME/.jwm/tmp/recent_doc_dump
-
+XBEL_PATH=$HOME/.local/share/recently-used.xbel
 if [[ -e $XBEL_PATH ]];then
-   ITEMS=$(sed -rn 's_.*file://([^"]*).*_<Program label="\1">rox -s "\1"</Program>_ p' "$HOME/.jwm/tmp/recent_doc_dump") 
+    MAX_ITEMS=20
+    IFS=$'\n'
+    ITEMS=$(awk -F'"' '/file:\/\//  { sub("file://","",$2); print $2}' $XBEL_PATH)
+    printf -v ITEMS "${ITEMS//%/\\x}"               #convert UTF-8
+    ITEMS=$(find $ITEMS   | tail -n $MAX_ITEMS)     #remove non-existent files, cut list off at MAX_ITEMS
+    ITEMS=$(echo "$ITEMS" | sed  -r 's_(.*/)(.*)_<Program label="\2" tooltip="\1\2">rox -s "\1\2"</Program>_')
 else
-   ITEMS="<Program label=\"${XBEL_PATH##*/} $(gettext 'not found')!\"></Program>"
+    ITEMS="<Program label=\"File ${XBEL_PATH##*/} $(gettext 'not found')!\"></Program>"
 fi
-echo '<JWM>'
-echo -e "${ITEMS//%/\\x}" | tail -n $MAX_ITEMS
-echo '</JWM>'
+echo  "<JWM>
+$ITEMS
+</JWM>"


01micko:
Going by what MochiMoppel said [here](http://www.murga-linux.com/puppy/viewtopic.php?p=1035839#1035839) is this the appropriate action?

Going forward, especially. It is really a bug in JWM and whatever else that can't handle the long filename.

Surely it is better to truncate the path/filename. This requires more thought IMHO, however I wont close this awaiting further comment.

peabee:
Guess it depends which problem you think is worse.........multiple files with same name but no path or long filenames with paths messing up display to be unusable.....
Certainly middle truncation to a max length e.g. first (or last?) 20 chars of path + 3*dots + last 20 chars of filename would be an answer.....

01micko:
It's doable ...

01micko:
... like so

```shell
#!/bin/bash
#this script is started from $HOME/.jwmrc

export TEXTDOMAIN=jwm
export OUTPUT_CHARSET=UTF-8

XBEL_PATH=/root/.local/share/recently-used.xbel
MAX_ITEMS=20
MAX_PATH=80

tail -n $((MAX_ITEMS*13)) $XBEL_PATH > $HOME/.jwm/tmp/recent_doc_dump

if [[ -e $XBEL_PATH ]];then
   ITEMS0=$(sed -rn 's_.*file://([^"]*).*_<Program label="\1">rox -s "\1"</Program>_ p' "$HOME/.jwm/tmp/recent_doc_dump") 
   ITEMS=$(echo "$ITEMS0" | while read path
   do
     ENT=`echo "$path"|grep -o '=.*>r' | sed -e 's%=%%' -e 's%>r%%' -e 's%\"%%g'`
     SIZE=`echo "$ENT"| wc -c`
     if [ $SIZE -gt $MAX_PATH ] ; then 
       ENT_FIRST="${ENT:0:$(($MAX_PATH - 5))}"
       ENT_SECOND="${ENT:$(($SIZE - 5)):4}"
       NEW_ENT="$ENT_FIRST~$ENT_SECOND"
       echo "$path"|sed "s%$ENT%$NEW_ENT%"
     else
       echo $path
     fi
   done)
else
   ITEMS="<Program label=\"${XBEL_PATH##*/} $(gettext 'not found')!\"></Program>"
fi
echo '<JWM>'
echo -e "${ITEMS//%/\\x}" | tail -n $MAX_ITEMS
echo '</JWM>'
```

You _could_ get `MAX_PATH` as a factor of screen width, or tune it. I only picked 80 because of console default width, it could be higher. Play around with it. IDK about CJK widechars though. One for Mochi.

EDIT: if you want to change the tilde to 3 dots, go for it.
Adjust `ENT_SECOND` by adding _equal_ amounts to the _numerals_ in the var if you want to show more of the ending, but also adjust `ENT_FIRST` by the same.

peabee:
Thanks Mick
I get
sed: -e expression #1, char 95: unknown option to `s'
from line 23
with this file:
/mnt/sdb1/Backups/daily.1/mnt/home/My_Documents/My_Money/Purchases/Kitchen 2019/Electra FF170WFF/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx

01micko:
hmmm ~ recreated that path (sda1 instaed sdb1) and no problem here

longpath

peabee:
It's line 23 thats the problem - the double quotes cause the error. (but even if replaced by single quotes the substitution doesn't occur.
The line can be replaced by:
echo "${path/$ENT/$NEW_ENT}"

peabee:
```
#!/bin/bash
#this script is started from $HOME/.jwmrc

export TEXTDOMAIN=jwm
export OUTPUT_CHARSET=UTF-8

XBEL_PATH=/root/.local/share/recently-used.xbel
MAX_ITEMS=20
MAX_PATH=76

tail -n $((MAX_ITEMS*13)) $XBEL_PATH > $HOME/.jwm/tmp/recent_doc_dump

if [[ -e $XBEL_PATH ]];then
   ITEMS0=$(sed -rn 's_.*file://([^"]*).*_<Program label="\1">rox -s "\1"</Program>_ p' "$HOME/.jwm/tmp/recent_doc_dump") 
   ITEMS=$(echo "$ITEMS0" | while read path
   do
     ENT=`echo "$path"|grep -o '=.*>r' | sed -e 's%=%%' -e 's%>r%%' -e 's%\"%%g'`
     SIZE=`echo "$ENT"| wc -c`
     if [ $SIZE -gt $MAX_PATH ] ; then 
       ENT_FIRST="${ENT:0:$(($MAX_PATH - 38))}"
       ENT_SECOND="${ENT:$(($SIZE - 37)):38}"
       NEW_ENT="$ENT_FIRST...$ENT_SECOND"
       echo "${path/$ENT/$NEW_ENT}"
     else
       echo $path
     fi
   done)
else
   ITEMS="<Program label=\"${XBEL_PATH##*/} $(gettext 'not found')!\"></Program>"
fi
echo '<JWM>'
echo -e "${ITEMS//%/\\x}" | tail -n $MAX_ITEMS
echo '</JWM>'

```

01micko:
Well, do you want to suggest your edit to BKJ and see if he wants to try it? :+1:

01micko:
I gave your version a test and it seems good, however consider a small eee pc and we'd probably run into the original problem.. even on a res of 1024x768. I still have no idea what would happen with CJK.

I will back pedal on my original stance, I do see a need for this especially if one downloads a web page which these days can have ridiculously long filenames. I guess we can thank google for that, and possibly ms, amazon,apple too.

01micko:
Ok, the edge of my menu is just over 800px (~830) so an eee pc with 800x600 res _should_ be ok. Since `wc -c` counts bytes widechars (CJK)  _should_ be ok as IIRC these are 4 bytes.

Do you want to issue a new PR?
For the record, this HP lappy uses 1366x768... the only real issue is if someone increases the global font size! But for sake of agility (this script needs to be fast) I think it's fine as is. I notice Mochi [here](http://murga-linux.com/puppy/viewtopic.php?p=1035362#1035362) modified his original to sort by date changed. That would require even more gymnastics to truncate. Too slow methinks.

peabee:
Have sent it to BKJ - will wait for his feedback before a new PR.....

PR #1606, opened 2019-10-08
Update box_passwd

rizalmart:
Some gui improvements

### DIFF
--- a/woof-code/rootfs-skeleton/usr/lib/gtkdialog/box_passwd
+++ b/woof-code/rootfs-skeleton/usr/lib/gtkdialog/box_passwd
@@ -100,7 +100,7 @@ fi
 #=================================================================
 
 S='
-<window title="'${TITLE}'" image-name="/usr/share/pixmaps/puppy/file_lock.svg">
+<window title="'${TITLE}'" image-name="/usr/share/pixmaps/puppy/file_lock.svg" width-request="320" resizable="false">
   <vbox>
    <frame '${TEXT}'>'
      [ "$3" ] && S=$S'<text xalign="0" use-markup="true"><label>"'$3'"</label></text>'


wdlkmpx:
setting size + making a dialog unresizable... is that a good idea?

I saw the same issue in xkbconfigmanager, I decided that it was a bad idea, so I changed the code, I made the dialog smaller, did not set a default size, did not include the weird vbox stuff with scrollbars and it's not resizable, the only thing I added from the PR.

Those who have big screens, will need big font sizes.. and this will make things look really bad. You have think about all possible situations in order to make something suitable for small and big screens.

rizalmart:
For password dialog, yes it is. But it depends, if the dialog was meant to resize then resizable should be enabled. However if the dialog does not meant to be resize, then resizable should be disabled. Otherwise the gui will look broken on end user's perspective

wdlkmpx:
xkbconfigmanager looked bad in dpup with the font size I'm using.

It's because it had a default size specified in the html code, without
the fixed size, gtkdialog respects the font size and produces bigger
or smaller dialogs.. depending on the font size.

640x480 should be minimun WxH, anything smaller doesn't make sense.

I'll merge this in a few hours.

rizalmart:
In that case, here are some options

1. The font size must be standardize
2. Make the objects in the dialog scalable with the use of hbox/vbox space-expand and space-fill combinations

Take a look on my commits on file_chooser. It scales properly upon resizing.

wdlkmpx:
Let's put it this way, if it looks bad in the very famous
Polished-Blue gtk theme, it will be merged, but some changes will be
discarded.

I already edited many dialogs, I made many dialogs smaller, because
the Polished-Blue theme somehow has a big font.

I'm using a 15" monitor, so I can tell when something is becoming too
small. I also have a netbook, and in my last test, most dialogs were ok there.

box_passwd is a script that receives arguments and the text can change.

So, let's think about how things could work fine with big and small
fonts, with different languages.

wdlkmpx:
I guess some of the changes will be reverted, someone who notices that some dialogs should be resizable should comment. 

Partview was meant to be resizable. It's hard work to revert stuff

PR #1615, opened 2019-10-12
move finn to /home and make loginmanager functional

wdlkmpx:
this requires:
- busybox su
- a minimally compiled sudo (the compat distro sudo might not work)

see common32/common64 pet repos
also petbuilds/generic/basic/sudo


wdlkmpx:
Just raising awareness of this.. it may remain open for a few weeks or months.

I have this working in dpup. I can easily switch between root and finn.. after restarting of course

finn:
- rox desktop/jwm: works but there's a problem with the drive icons / jwm: I suspect fix menus is broken
- pcmanfm/jwm: pcmanfm works flawlessly / jwm: I suspect fix menus is broken

It's just the starting point, it's the 3rd time I'm doing something like this and I kind of forgot some things. Apparently rox must be started by root for it to work properly, looking at the error msgs, that's probably what I was doing

Documentation:
Screenshot(4)

Screenshot(5)

01micko:
Everything has been hard coded to `/root` forever. I can take a look at jwm_xdg_menu and probably move stuff to `/etc/xdg/` although IIRC jwm stores it's system rc in `/etc/system.jwmrc`, but I suppose that can be made a symlink.

That way jwm configuration will be system wide and simply running `sudo fixmenus` should fix that.

I remember when I moved fido to `/home/fido` there were issues with the drive icons, however being in `/tmp` should be ok as long as `/tmp` has the sticky bit set.

wdlkmpx:
A year or two years ago I made the rox desktop work properly with
/home/fido or finn

rox must be started by root. udev calls the rox scripts and rox uses
su fido.... and stuff... I think I copied or moved /root/.xinitrc and
/etc/X11/...... and made it a special script that is always started by
root... in the sudoers file

I also moved /root/.XLOADED to /var/shared/.XLOADED or something..
you'll see code in init referencing it.

It's actually not that hard normalize stuff, but it takes time, and
more importantly this means breaking backwards compatibility for
everyone that doesn't use the default woofce stuff. I think I had a
huge commit that changed 100-200 files?

Once I cancelled the projects... I gradually merged many changes from
those lost projects to testing, but now I need to remember the past.

While I was talking about fixing fixmenus.. I made a change that
actually fixes it when it's run by finn:

https://github.com/puppylinux-woof-CE/woof-CE/commit/f5e30ab5a37ad1b269e2aa3e44a1c36e67b32fd8

So, this is almost safe to merge, fixing the hardcoded /root in
everything might take too long, and it's almost christmas, I can hear
the phone ringing with some news that will led me to see $$$ and I'll
become busy.

finn is experimental, it's not wise to use it unless you're a
developer wanting to fix everything to work with it. But let's just
say that if we install the right apps that don't need to be fixed,
finn becomes quite usable.

wdlkmpx:
Long story short, I made a ton of changes that make rox desktop work properly with /home/xxx .. but currently needs to be fixed/reimplemented.. but about 90% is already done.

The change from /root/.packages to /var/packages was actually part of those changes, /root/.packages became a symlink. The /var/packages change was merged when I was really angry. Most of the really interesting changes happen when I'm not in the mood.

I also moved /root/.pup_event to /tmp/pup_event_frontend.. I moved /root/spot to /home/spot

I made many changes paving the way to this PR. 

Just like when I made a ton of changes until the multiarch symlinks were safe to delete.  Anybody else would have started from scratch.

01micko:
Why don't you create a new branch?  'finn' or something. Barely anyone test forks including me. If the PR is small I usually merge locally and do a build to test then merge to upstream.

wdlkmpx:
Its' been a month, time flies.

This pr is mostly OK to merge, but the rox desktop needs fixing, and I know how to fix it, something I've done before. So I'll just merge this and fix the issue later.

The structural changes that are happening require an active community of builders that are willing to use what they are given and provide patches or something, but it's all about releases, it's all about.. the word.. hmmm.. the meaning of the word.

Punishment incoming.

PR #1622, opened 2019-10-17
remasterpup2: Improved package conversion

rizalmart:
Improving package integration from user installable to builtin upon remastering puppy


wdlkmpx:
remasterpup2 is a script I'll never use. So I'll never test this, but I'll merge it soon

Here is my attempt to update remasterx by shinobar
[remasterx.gz](https://github.com/puppylinux-woof-CE/woof-CE/files/3740695/remasterx.gz)

I think it needed more fixes, but as I no longer remaster sfs's, I always create them from scratch, I didn't feel the need to continue

When testing this stuff, lz4 compression (or no compression at all) makes sense

wdlkmpx:
I guess something like this could be faster:
`sed -i "s%${oldpkg}%${line}%" /tmp/root/.packages/woof-installed-packages`

PR #1625, opened 2019-10-24
removepreview.sh: More uninstallation improvements

rizalmart:
1. Added package file deletion for PUPMODE 13
2. Improved removal of empty directories. Ignores "/" on the package file list
3. Improved updating cache files

This commit was fully tested


wdlkmpx:
I really hope it was fully tested in a recent woofce puppy. it would be wrong to assume that some other non-woofce distro works exactly the same.  I'll test it later.

One day I'll remove all gtk1 stuff

PR #1629, opened 2019-10-25
installpkg.sh: Improve package upgrade/downgrade

rizalmart:
Ask the user to keep the package files which was not a part of upgrading/downgrading installed packages.

For example an installed package-1.0 contains
/usr/bin/app
/usr/bin/file-C

Then the user was installed package-1.1 contains
/usr/bin/app
/usr/bin/file-B

Therefore package-1.0 was replaced by package-1.1
Since /usr/bin/file-C was not a part of package-1.1, The user will give an option to keep /usr/bin/file-C or delete it. 

This commit was fully tested

### DIFF
--- a/woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
+++ b/woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
@@ -627,7 +627,15 @@ if [ "$installed_pkg" != "" ]; then
 	  else
 	   #Delete the file which is not a part of upgrade
 	   if [ -e "$xline" ] && [ ! -d "$xline" ]; then
-	    rm -f "$xline"
+	    
+	    if [ -d "/initrd/pup_rw$xline" ] && [ ! -L "/initrd/pup_rw$xline" ]; then
+	     rm -f "/initrd/pup_rw$xline"
+	    fi
+	    
+	    [ -e "/initrd${SAVE_LAYER}${xline}" ] && rm -f "/initrd${SAVE_LAYER}${xline}"
+	    
+	    [ -e "$xline" ] && rm -f "$xline"
+	   
 	   fi
 	  fi
 	 fi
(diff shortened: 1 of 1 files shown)


wdlkmpx:
I... I... I try to understand this, but despite the explanation somehow I can't...

What if the packages are incompatible and by keeping files from the old package you break the new one? What if zagreb installs and replaces many VLC's versions and keeps breaking it even more?

libs. if you  want to install a new version of the lib with a different .soname... if libpng-14.0 is the package installed, then maybe adding libpng16-16.0  makes sense. I learned that from debian.

rizalmart:
I explain to you little further. If you read the code. The user was ask to keep the old files or not when he try to upgrade/downgrade the package. I'll give you example:

If zagreb installs vlc-1.0 then he upgrade to vlc-2.0. PPM asks if he will keep the files of vlc-1.0 which was not overwritten by vlc-2.0.

here is the sample contents of vlc-1.0
/usr/bin/vlc
/usr/lib/libfoo1.so

here is the sample contents of vlc-2.0
/usr/bin/vlc
/usr/bin/libfoo2.so

You will see here that if he install vlc 2.0, the vlc executable files of vlc-1.0 will be overwritten by vlc-2.0 while libfoo1.so does not overwritten by vlc-2.0

If he choose yes then the files of vlc-1.0 which was not overwritten will simply add the unoverwritten files to package file list of vlc-2.0.

So it means the contents of vlc-2.0.files upon upgrading vlc-1.0 contains
/usr/bin/vlc
/usr/lib/libfoo1.so
/usr/lib/libfoo2.so

Those old files will be removed if vlc-2.0 was removed.

If he choose no then the files of vlc-1.0 which was unoverwritten by vlc-2.0 will be remove. A clean upgrade for short.

This is also applied on downgrading installed user packages

mavrothal:
Looked briefly at the patch and not sure is there. Associated files, say /usr/lib/libfoo1.so maybe used as dependencies of latter packages and not reinstalled by those. Removing them with the upgrade/downgrade will brake those packages. 
If I remember, PPM had (has?) code to check for those scenarios. You might want to go through that road (if still present)

PR #1634, opened 2019-11-04
Create psynclient.desktop

rizalmart:
psynclient is a gui frontend of synclient for configuring synaptic touchpads. It works the same as flsynclient but it has advanced options, extremely lightweight, and architecture independent.


peabee:

PR #1635, opened 2019-11-04
Create pupmixer

rizalmart:
pupmixer is a gui frontend for alsamixer written in bash and gtkdialog. It supports multiple soundcards.


peabee:
retrovol is not in rootfs-skeleton..........

PR #1636, opened 2019-11-04
Create pupmixer.desktop

rizalmart:
pupmixer is a gui frontend for alsamixer written in bash and gtkdialog. It supports multiple soundcards.


peabee:

PR #1637, opened 2019-11-04
installpkg.sh: Clean upgrade/downgrade package installation

rizalmart:
Just perform clean upgrade/downgrade if the same package name but different version is detected


wdlkmpx:
There is a problem here.

To perform an upgrade the pkg must be uninstalled first. And I really hope  it has been tested in a recent woofce iso.

Anyway Pkg should completely replace the PPM one day.

PR #1638, opened 2019-11-06
xorgwizard: Add startx option

rizalmart:
Add start X option in xorgwizard commandline menu

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/xorgwizard
+++ b/woof-code/rootfs-skeleton/usr/sbin/xorgwizard
@@ -383,7 +383,7 @@ rm -f /tmp/xorg.cli /tmp/xorg.cli.res
 
 case $XCHOICE in
 	"-") exit ;;
-	"1)") startx ;;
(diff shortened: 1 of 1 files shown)


wdlkmpx:
This breaks the shutdown sequence in dpup and probably in all wce pxps

rizalmart:
Can you give me some details why it breaks shutdown sequence?

wdlkmpx:
xwin or something.. will reply tomorrow.. cellphone running out of battery 2%

wdlkmpx:
When I select `Start X Window System` Xorg starts fine.

But when I want to reboot, it only drops to the console and xauth is missing, and I need to type reboot again.

I think it's because it's bypassing xwin or something.

rizalmart:
That's weird. startx should start Desktop Environment normally. I modified now the script. instead of using startx to start DE, it uses xwin command

wdlkmpx:
It works fine now. Well there some outstanding anomalies. It should be fixed by deleting most of the codebase.

PR #1679, opened 2019-11-27
rc.sysinit: Load ndiswrapper

rizalmart:
If a installed ndis driver is found. load the ndiswrapper module


wdlkmpx:
I've seen code in sns and netwiz that deal with ndiswrapper and at least 5 network managers are supported.. what does @rserwin1 think of this change?

PR #1691, opened 2019-12-07
Create network-privacy.desktop

rizalmart:


peabee:
I think this should be provided as a .pet in the ibiblio common32/64 repos rather than as part of Woof-CE.
1. it is a specialist app only likely to be used by a very few users
2. macchanger is quite a big dependency (185KB compressed)
3. macchanger is not provided by Slackware (Salix does have it) see pkgs.org
4. will complicate the menu structure further for ordinary users
5. are there any known requests on the forum for its addition?

@rizalmart - would be great if you could learn how to combine all items into a single pull request - there were 22 emails associated with this single change!!!

01micko:
@peabee and @rizalmart 

Please see #1715

01micko:
The whole lot should be one PR anyway. I'm not reviewing every singe PR here so I'm closing until it's consolidate.. a PR for a .desktop file? Linus would swear! :joy_cat:

PR #1693, opened 2019-12-07
Update network_default_connect

rizalmart:
Added randomizing mac address before connecting to network (requires macchanger)


01micko:
Is this **off** by default? 

Case in point: user has a server with static IP set by the router.

Issue #1716, opened 2019-12-11
can't create save file when using ssh (headless)


01micko:
I've solved the issue of booting headless and connecting via wifi in e8647f4 in PUPMODE=5 so I logged in to a pi over ssh, using ` mpscan -t 6 -p 22 10.1.1.1 - 10.1.1.254` to find the IP.

However I've tried a few methods to save session; 
- log in with X forwarding
- without X forwarding
- killing X
- `nohup wmreboot` (in X)
- `nohup reboot` (no X)
- a few other ways

I didn't try `shutdown-config` directly. Maybe it will work?

Wondering if @gyrog or @woodenshoe-wi have any ideas?

This is not a raspberry pi specific problem.

s243a:
Isn't there some flag that is set that triggers snapmergepuppy? I think said flag is created when you press the save button.

01micko:
@s243a you misunderstand, I can't create the save file from pupmode 5. Perhaps I should change the title.

gyrog:
shutdownconfig is interactive, but it is supposed to have a non-X mode, so maybe it will work with ssh.
The alternative would be to write what is needed diretly to /tmp/shutdownconfig_results.

01micko:
Thanks @gyrog 

I'm using X forwarding so running shutdownconfig directly _does_ work - with bugs.

1. There is constant loop of a 'broken pipe', varies in frequency but doesn't seem to affect the result.

2. No matter what reboot routine I try to use, the only one that works is `nohup busybox reboot`. The touble with that though is that the ssh VT is left hanging, a simple close of the window seems sufficient.

3. On the pi though (this is a pi specific anomaly) it reboots to pupmode 5. Luckily, I have a gtkdialog GUI to set the default boot option and set it to the raspupave. I can ssh in and run that GUI. Another reboot and we land in pupmode 13, however stuff isn't saved. I need to run `save2flash` then kill stuff and run `nohup busybox reboot`

It would be nice to fix all this but I can live without as long as I carefully document my steps. It probably would be better booting `pfix=nox` and not using X forwarding. I'll try that on the weekend.

01micko:
`shutdownconfig` actually works with X forwarding. A save folder is created where I want it.

The problem seems to be rebooting and shutting down, even prepending with `nohup `.

With out X forwarding may well be the go, but we don't have CLI utils for some things, such as syncing to a time server, need to fix that.

s243a:
[...]

PR #1723, opened 2019-12-14
slacko: use slackware version in pupsave

wdlkmpx:
It's possible to create pupsaves from Live CDs.

It's also possible to boot directly from ISOs [and create pupsaves] with special grub4dos/grub2 menu entries, so it's easy to test default stuff in real hardware.

(example menu entry can be found in this commit
or in initrd-progs/0initrd/sbin/isoboot)

The pupsaves are always located in HDD/SSD/Flash/etc drvs.

But there's a problem with slacko: pupsave collision.

New pupsaves:

- slacko 5.8.12  : slacko140_32 [save[.4fs]]
- slacko64 5.8.12: slacko140_64 [save[.4fs]]

- slacko 6.5  : slacko141_32 [save[.4fs]]
- slacko64 6.5: slacko141_64 [save[.4fs]]

- slacko 7.0rc  : slacko142_32 [save[.4fs]]
- slacko64 7.0rc: slacko142_64 [save[.4fs]]

### DIFF
--- a/initrd-progs/0initrd/sbin/isoboot
+++ b/initrd-progs/0initrd/sbin/isoboot
@@ -21,14 +21,15 @@
 # - a savefolder is used
 
 #===================
-#    grub4dos  
+#  grub4dos  - https://github.com/chenall/grub4dos/releases
 #===================
-#title Start pup.iso
-#  find --set-root --ignore-floppies /ISO/pup.iso
-#  map /ISO/pup.iso (0xff)
+#title Start /ISO/pup.iso
+#  set iso=/ISO/pup.iso
+#  find --set-root --ignore-floppies %iso%
+#  map %iso% (0xff) || map --mem %iso% (0xff)
 #  map --hook
 #  root (0xff)
-#  kernel (0xff)/vmlinuz find_iso=/ISO/pup.iso
+#  kernel (0xff)/vmlinuz find_iso=%iso%
 #  initrd (0xff)/initrd.gz
 
 #===================
--- a/woof-distro/x86/slackware/14.0/DISTRO_SPECS
+++ b/woof-distro/x86/slackware/14.0/DISTRO_SPECS
@@ -1,11 +1,11 @@
 #One or more words that identify this distribution:
 DISTRO_NAME='Slacko Puppy'
 #version number of this distribution:
-DISTRO_VERSION=5.8.10
+DISTRO_VERSION=5.8.12
 #The distro whose binary packages were used to build this distribution:
 DISTRO_BINARY_COMPAT='slackware'
 #Prefix for some filenames: exs: slackosave.2fs, slacko-5.7.1.sfs
-DISTRO_FILE_PREFIX='slacko'
+DISTRO_FILE_PREFIX='slacko140-32'
 #The version of the distro whose binary packages were used to build this distro:
 DISTRO_COMPAT_VERSION='14.0'
 #-
--- a/woof-distro/x86/slackware/14.1/DISTRO_SPECS
+++ b/woof-distro/x86/slackware/14.1/DISTRO_SPECS
@@ -1,11 +1,11 @@
 #One or more words that identify this distribution:
 DISTRO_NAME='Slacko Puppy'
 #version number of this distribution:
-DISTRO_VERSION=6.4
+DISTRO_VERSION=6.5
 #The distro whose binary packages were used to build this distribution:
 DISTRO_BINARY_COMPAT='slackware'
 #Prefix for some filenames: exs: slackosave.2fs, slacko-6.3.2.sfs
-DISTRO_FILE_PREFIX='slacko'
+DISTRO_FILE_PREFIX='slacko141-32'
 #The version of the distro whose binary packages were used to build this distro:
 DISTRO_COMPAT_VERSION='14.1'
 #the kernel pet package used:
--- a/woof-distro/x86/slackware/14.2/DISTRO_SPECS
+++ b/woof-distro/x86/slackware/14.2/DISTRO_SPECS
@@ -1,11 +1,11 @@
 #One or more words that identify this distribution:
 DISTRO_NAME='Slacko Puppy'
 #version number of this distribution:
-DISTRO_VERSION=6.9.9.9
+DISTRO_VERSION=7.0rc
 #The distro whose binary packages were used to build this distribution:
 DISTRO_BINARY_COMPAT='slackware'
 #Prefix for some filenames: exs: slackosave.2fs, slacko-6.9.9.9.sfs
-DISTRO_FILE_PREFIX='slacko'
+DISTRO_FILE_PREFIX='slacko142-32'
 #The version of the distro whose binary packages were used to build this distro:
 DISTRO_COMPAT_VERSION='14.2'
 #the kernel pet package used:
--- a/woof-distro/x86_64/slackware64/14.0/DISTRO_SPECS
+++ b/woof-distro/x86_64/slackware64/14.0/DISTRO_SPECS
@@ -1,11 +1,11 @@
 #One or more words that identify this distribution:
 DISTRO_NAME='Slacko64 Puppy'
 #version number of this distribution:
-DISTRO_VERSION=5.8.10
+DISTRO_VERSION=5.8.12
 #The distro whose binary packages were used to build this distribution:
 DISTRO_BINARY_COMPAT='slackware64'
 #Prefix for some filenames: exs: slacko64save.2fs, slacko64-5.8.8.sfs
-DISTRO_FILE_PREFIX='slacko64'
+DISTRO_FILE_PREFIX='slacko140-64'
 #The version of the distro whose binary packages were used to build this distro:
 DISTRO_COMPAT_VERSION='14.0'
 #the kernel pet package used:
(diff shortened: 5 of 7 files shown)


01micko:
The idea has merit but I don't agree with the execution. It will mess up many scripts that rely on `$DISTRO_FILE_PREFIX`.

Also `raspup` and `dpup` will have the same problem.

Why not use `$DISTRO_DB_SUBNAME`?

```
# . /etc/DISTRO_SPECS 
# echo $DISTRO_DB_SUBNAME 
slacko6414.2
```

The debians then we can append the version to `DISTRO_DB_SUBNAME`
eg:
` DISTRO_DB_SUBNAME=raspup10 # buster`

wdlkmpx:
Ah.. I didn't think much about how many scripts will be affected.

I see a lot of scripts use DISTRO_FILE_PREFIX, and this would make everything clearer... literally.

Looking at the code, so far the only breakage I notice happens in 3builddistro.. the slacko logo will not be renamed to puppylogoXX.png. And this does not bother me...

I'll build a couple of ISOs to see the results.

$DISTRO_DB_SUBNAME is a good alternative but it will affect all builds and init must be modified to look for ${DISTRO_DB_SUBNAME}save*, then ${DISTRO_FILE_PREFIX}save. 

shutdownconfig must also be modified to use $DISTRO_DB_SUBNAME  if available then fall back to $DISTRO_FILE_PREFIX

I think $DISTRO_DB_SUBNAME is not available in zwn.

Some builders will be upset by the change, since DISTRO_FILE_PREFIX indeed defines the pupsave.

Perhaps there needs to be a variable only for pupsaves. Anyway it's always easy to break backwards compatibility when fixing the unfixable.

peabee:
I wouldn't want pupsaves named by anything other than DISTRO_FILE_PREFIX.......they have to be consistent with .sfs naming

peabee:
If you do change the DISTRO_FILE_PREFIX for slackos then you also need to change the examples line above......

01micko:
```
.  /etc/DISTRO_SPECS
echo ${DISTRO_DB_SUBNAME/$DISTRO_COMPAT_VERSION/}
```

Ok, I get what you are getting at @wdlkmpx and that is trying to automate pupsave creation. Why not just append a random number?

gyrog:
I don't see a possibility for a "save" collission.
If each frugal install is in a unique directory, then it doesn't matter if all "save"'s have the same filename, heck, all Puppies could just save to "save" and there would still be no collision.

wdlkmpx:
Save collision: this happens when booting from live cds or isos.

If you create a pupsave with slacko 14.2 (cd) and then use a slacko 14.1 cd, it will find and use the slacko 14.2 save..

To understand this you must be testing stuff, I built like 30 isos in the past few days and decided that it's just easier to just copy and paste a iso file to a directory and boot from it.

Of  course it's just easier to test with qemu without rebooting. But this may help multisession cd users find another way to do things..

wdlkmpx:
Talking about qemu, I don't have a way to test a puppy in one of the obscure machines that only support efi32 (bootia32.efi), but using qemu to boot a .iso with a edk2 bios (efi32/efi64/arm64) I can test uefi-booting even in a machine that doesn't support uefi, to test efi32 you don't even need a 64-bit capable processor. VMs make everything easier.

wdlkmpx:
Writing this from slacko141-32 6.5. I created slacko141-32save.... I changed the slacko DISTRO_FILE_PREFIX to slackoXXX-XX.

Perhaps I should change 32 -> i686 or x86 / 64 -> x86_64

devx_slacko141-32_6.5.sfs
slacko141-32-6.5.iso
- initrd.gz
- puppy_slacko141-32_6.5.sfs
- vmlinuz
- zdrv_slacko141-32_6.5.sfs

From what I can see only the common64 eudev is a failure or something, need to improve the recipe cos it's a bit hackish for 64 bit builds.

Live CD, this is the menu that is generated by the build system and the default installers (bootflash, puppyinstaller).

Screenshot(1)

Screenshot(2)

The menus are not affected, although they should be.

No breakage occurs. This is a [sensitive] change that may mess with branding andss.. but it doesn't get any clearer than that.

01micko:
I've had a long hard think about this PR and my conclusions follow..

### script breakage
While in woof itself nothing breaks there are surely a dozen or more third party scripts (contained in PET packages or SFS packages, either included in the `PKS_SPECS_TABLE` or in the repo) that will _not_ conform to the changes proposed.

### use case
What is the use case?
It seems to be simply booting off the iso, either hard copy or via `grub4dos` or `grub2` soft copy and creating a save is the primary use case. That will be rare case where a 'collision' may occur.

### upgrading
This makes upgrading slacko from say, 6.2 to 7.0 far more difficult for the average Joe. Sure, upgrading is certainly not perfect but I'm sure more users will try that than those who try to use an automatic save file.
If a user decides _not_ to install any extra packages, which I'm sure plenty do, upgrading can go very well. Or, if a user _remasters_ and then upgrades that can go well too. Obviously remastered changes will be lost but I'm sure we're not dealing with complete idiots _all_ of the time.

### intention
What is the intention? Put in code that auto-creates the savefile? Surely user input is needed (_yes: I want to save_, _no: no save for me thanks_) and there must be a better way; as I said earlier, append a random number or something.

### woof is editable!
If a builder fears a 'collision' is probable, then they can easily edit `DISTRO_SPECS` accordingly.

wdlkmpx:
How can you upgrade an incompatible pupsave, nonsense.

01micko:
> How can you upgrade an incompatible pupsave, nonsense.

I see you've never tried it.

wdlkmpx:
I have performed several upgrades and added many hacks to make them
easier - I removed basically all of the hacks I added.

The only side effect is a weird behavior and many bugs that have no
explanation. Only when you make no changes to the codebase an upgrade
is relatively safe, as long it's a small pupsave with very few
changes.

01micko:
> Only when you make no changes to the codebase an upgrade
is relatively safe, as long it's a small pupsave with very few
changes.

Basically what I [said](https://github.com/puppylinux-woof-CE/woof-CE/pull/1723#issuecomment-568256379)

Probably more importantly, a pupsave should _never_ be created without user permission. If permission is required, then appending _any_  suffix is a no brainer. People generally don't want stuff written to disk if permission is not given.

I'll refer you to the [pupdesk.flg fiasco](http://murga-linux.com/puppy/viewtopic.php?p=703111#703111) .. and there is plenty more, that link is the tip of the iceberg.. started a flame war in fact.

wdlkmpx:
Of course a pupsave of slacko141 is created without user permission,
you only need to create a slacko142 pupsave from CD and slacko 141
will use it and will perform an upgrade.

wdlkmpx:
Notice how I never talk about slacko 5, 6 or 7. Because I never talk
about tahrpup 5, 6, or 7. It's just tahrpup and tahrsave. But I think
it should be tahrXXsave to make it more consistent.

There seems to be a misunderstanding here. The save collision happens
if you create a pupsave from a physical live CD. If you test or use
several pxppx versions from cd, then it will happen. Booting from iso
or qemu [vhd] just creates the same scenario.

wdlkmpx:
And the main reason to remove multi-session cd support is... you can't
create a multi-session cd from a iso file or from a usb flash drv with
dd-ed iso or from qemu. There's no way to test it.

I dismantled all of my cd drives, my dvd drives are starting to fail
[need maintenance], and my bluray burner drive.... burning video files
to a 25GB optical disc might be a good idea.... and all I can think of
is that I need to transfer many video and audio files from dvds to a
hard drive before it's too late.

But it's easy to have many pupsaves and manage them from CD, currently
shutdownconfig is called only when no pupsave has been found [pupmode
5]... but the basename is all you need to know. By duplicating the
pupsave or by manually creating a new pupsave and appending chars to
the base pupsave name, you'll have a menu with pupsaves to choose
from... at every boot. That's what I'd call a multisession cd.

01micko:
OK, on more thought..

`slacko` is more or less my branding. I don't pretend to own it, anyone can use it in the context of a slackware based puppy. It started out as `spup` and my first ever release was `spup-100` based on slackware-13.0. My first 'official' release was _slacko-5.3_ based on slackware-13.37 released from Barry's woof.

Appending the 'o' is an _Australianism_ ...  micko, nicko,  davo, johnno,  (shazza, dazza, bazza - yep [we're a weird mob](https://www.youtube.com/watch?v=GCSKD_pn2Rk))

Why not revert back to `spup` in woof-CE, do your thing but don't delimit with underscore as all the system sfs are delimited with underscore.

No need to distinguish the 32 bit version as it is just plain old `spup`, 64 `spup64` then just append `130, 133, 140, 141, 142, 150` or whatever version resembles upstream.
So `spup64141` or `spup142` .. you get my drift.

To address the save creation, why not make it a boot param `pfix=forcesave` or something and then, in `init` mount where we have booted from (based on UUID or something) and create the save folder, so long as the partition is of type `linux`. Of course booting on bare metal with a hard CD/DVD with multiple partitions of type `linux` creation should be aborted as we have no idea where the user wants to put the save.

Actually, we have enough tools in init to create a _savefile_ (ext[2,3,4], f2fs) so it could go onto a windows partition.

01micko:
Actually, the `forcesave` thing would fix my issue #1716

wdlkmpx:
The reason I used a delimiter is because 64141 looks like a big
number. I'm not going to reopen (I can't) or make other changes to the
pr.

busybox mke2fs cannot create ext3/ext4 fs. It's quite easy to add the
real static mke2fs, it's compiled but discarded.

pxppx already supports using an empty pupsave, init detects it and
writes a flag file that is read by rc.update.

A few months ago I made rc.update a bit more "intelligent" to make it
work better even without the flag file created by init.

So it only takes creating a directory or a 2fs/3fs/4fs file and set
some variables to make it work.

PR #1730, opened 2020-01-03
init: Add switch init option

rizalmart:
Add option to switch init system by putting init=[systemd|runit|minit|openrc|sysvinit] on boot parameter otherwise it use busybox init by default

### DIFF
--- a/woof-code/rootfs-skeleton/sbin/init
+++ b/woof-code/rootfs-skeleton/sbin/init
@@ -38,7 +38,6 @@ openrc)
  [ -f /usr/sbin/openrc-init ] && INITEXE='/usr/sbin/openrc-init'
 ;;
 systemd)
- exec 1>/dev/console 2>&1
  [ -f /lib/systemd/systemd ] && INITEXE='/lib/systemd/systemd'
  [ -f /bin/systemd ] && INITEXE='/bin/systemd'
  [ -f /sbin/systemd ] && INITEXE='/sbin/systemd'
--- a/woof-code/rootfs-skeleton/sbin/init
+++ b/woof-code/rootfs-skeleton/sbin/init
@@ -14,7 +14,39 @@ if [ "$DISPLAY" ] || [ "$XDG_CONFIG_HOME" ] ; then
 	exit
 fi
 
+INITTYPE="$(cat /proc/cmdline | tr ' ' '\n' | grep -E "^init=" | cut -f 2 -d '=')"
+
 INITEXE='/bin/busybox init'
+
+case $INITTYPE in
+minit)
+ [ -f /bin/minit ] && INITEXE='/bin/minit'
+ [ -f /sbin/minit ] && INITEXE='/sbin/minit'
+ [ -f /usr/bin/minit ] && INITEXE='/usr/bin/minit'
+ [ -f /usr/sbin/minit ] && INITEXE='/usr/sbin/minit'
+;;
+runit)
+ [ -f /bin/runit ] && INITEXE='/bin/runit'
+ [ -f /sbin/runit ] && INITEXE='/sbin/runit'
+ [ -f /usr/bin/runit ] && INITEXE='/usr/bin/runit'
+ [ -f /usr/sbin/runit ] && INITEXE='/usr/sbin/runit'
+;;
+openrc)
+ [ -f /bin/openrc-init ] && INITEXE='/bin/openrc-init'
+ [ -f /sbin/openrc-init ] && INITEXE='/sbin/openrc-init'
+ [ -f /usr/bin/openrc-init ] && INITEXE='/usr/bin/openrc-init'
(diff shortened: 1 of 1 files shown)


s243a (commented):
I think this code just covers the full install case. Correct? If so then there would need to be a corresponding change in Initrd for the frugal install case.

I'm on my phone now so I can't look at the full code.

rizalmart:
Nope it covers frugal too. It doesn't need to change the initrd init script. I tested this and it works.

s243a:
Cool.I looked and you are correct. If the woof-CE custodians decide to implement this then they might want to consider whether or not there should be an option to explicitly specify "busybox init". I agree that "busybox init" should be the default, but if someone forks woof-CE then they might want to change the defaults.

01micko:
I have an opinion on this but for now I'll keep it to myself.

I would like the opinion of @peabee @mavrothal @wdlkmpx @gyrog @mrfricks @jamesbond3142 and others before a decision is made.

Just the word `systemd` in puppy base code may scare many off. It caused [furore](http://www.murga-linux.com/puppy/viewtopic.php?p=1044667#1044667) when it was simply mentioned in `PKGS_SPECS_TABLE`, because it is dependency of some debian based applications.

If it ends up being rejected, then there is nothing to stop a noarch `pet` being installed to overwrite `/sbin/init`, after all, woof-CE is meant to be extensible.

s243a:
> I would like the opinion of @peabee @mavrothal @wdlkmpx @gyrog @mrfricks @jamesbond3142 and others before a decision is made.
Agreed.  
> Just the word `systemd` in puppy base code may scare many off. It caused [furore](http://www.murga-linux.com/puppy/viewtopic.php?p=1044667#1044667) when it was simply mentioned in `PKGS_SPECS_TABLE`, because it is dependency of some debian based applications.

In fairness, many people didn't understand why. After some discussion I think what came out is that systemd-udev is more about branding than an actuall systemd dependency. Therefore, at the moment it seems like a harmless change and it's good that @wdlkmpx has provided this temporary fix but in my opinion in the long term it would be much better to use eudev instead.

> If it ends up being rejected, then there is nothing to stop a noarch `pet` being installed to overwrite `/sbin/init`, after all, woof-CE is meant to be extensible.

Perhaps this is a politically better approach to the suggested change. In other words put this option in rootfs-packages, which the puppy builder can consciously choose to include in their build. The only downside of this approach is code maintence. An important change might get put into the rootfs-skeleton version of init that doesn't make it into the pet. 

Anyway, avoiding systemd is high on my priority list but I favour giving puppy builders choice over universally protesting systemd. Many recent applications only provide systemd startup scripts. That doesn't mean that they can't be started other ways but a user might not know how to do it. So far I have been able to avoid systemd but this might not always be the case. 

BTW, we can make it so a pet can't overwrite /sbin/init if we want. What you do is put a pristine copy in initrd and copy the pristine copy over the savefile version at boot. There could be a warning at boot to say that init was changed and the user could choose to boot with either the changed version or a pristine version.

peabee:
I have not heard a compelling case for this PR. As far as I know the current Puppy init works entirely adequately and successfully in the Puppy environment. If apps are moving to somehow be dependent on systemd for their installation then it may be that we will need to accommodate that if we want to have those apps available. I am not aware of any mainstream apps that are currently so encumbered. The backlash against systemd seems to be against its pervasive nature, its tendency to suck in ever more functions into a monolithic structure, and the attitude of its developers and protagonists. These attributes seem to be at odds with the Puppy philosophy of embracing the best modular components from wherever they are available. I would therefore be against any PR that suggested Puppy was embracing or approving of systemd. I think that Mick's suggestion that there are already plenty of mechanisms in Woof-CE available to system builders to configure their systems in whatever way they want is entirely correct and we shouldn't confuse developers or users by providing multiple choice solutions to problems that don't exist. I therefore vote against this PR.

jamesbond3142:
The PR is already closed, but here is my opinion anyway.

1. /sbin/init should not be a script. It should be a symlink to the actual init binary.

2. Initialisations that now are being done in /sbin/init script, should be re-factored into:
a) initrd /sbin/init, or
b) /etc/rc.sysinit

3. The binary /sbin/init (point 1 above) should have a hook to run /etc/rc.sysinit for its initialisation (using busybox init, this means adding the proper entry in /etc/inittab; in other init systems, similar hook should be provided).

4. When 1-3 has been done, changing init is just a matter of changing the symlink and adding the appropriate hooks. This can be done using a pet package; something that Woof-CE builders would use to customise their build.

5. I don't see the benefit of "run-time configurable init system" (e.g. passing boot parameter to specify init system to use) for regular users. I can see it helping developers for easy testing, but I don't see it useful for anyone else.
Any regular users who really want to use a different init system can always install the pet (see #4 above) and be done with it. They can always "recover" to the default init system by booting with pfix=ram.

Issue #1715, opened 2019-12-07
Coding style and PR etiquette.


01micko:
## coding style

There is plenty of messy shell code in woof. @wdlkmpx and myself (and others) over the years have attempted to clean it up somewhat but if someone commits code with a mix of tabs and spaces then we end up back at square :one: 

### basic guidelines

If a script has spaces ( 1 or 2 ) as the indent then follow that.
If a script has tabs as the indent then follow that.
Please don't mix them up.

### code quality

EG:
```shell
if [ "`which program`" != "" ] ; then 
  exec program
fi
```
Let's analyze this
1. we perform a conditional test
2. we call an _external_ program - `which` _in a subshell_
3. we compare it to `""` (essentially _nothing_) 
4. the result is evaluated and we proceed with the block or skip the block depending on result.

So what could go wrong?

User or builder installed **GNU** `which` giving vastly different output than `busybox which`

Let's try it another way
```shell
if type program >/dev/null 2>&1 ; then
  exec program
fi
```
So what happens here?
1. call the bash/ash/dash internal `type` to determine if `program` is installed and suppress the output
2. if `type` is successful the condition is satisfied and we proceed or skip as before

Can we see that this is faster? In a large script that has to test perhaps for several programs to determine which one to use then this will have a significant speed boost.

Imagine this on a pi Zero

```shell
browser=''
# choose first available browser
for B in firefox seamonkey chromium google-chrome opera vivaldi links lynx midori 
do
  # [ "`which $B`" != "" ] && browser=$B && break # too slow
  type $B >/dev/null 2>&1 && browser=$B && break
done
```

Basically, `bash` and `ash` are very powerful and can speed up your code if used diligently. Of course sometimes we can't get away from using external calls, especially `sed`, `grep` and `awk`. I myself am no expert in shell internals and I often refer to ash or bash man pages; there are many to be found online. Redirections can also be very powerful as well as the operators. Take time to familiarise yourselves more with bash and ash and you will be pleased with the results. Happy coding :smile: 

## pull requests

If you have a pull request that requires that new programs need to be installed in all `PKGS_SPECS_TABLE` s variables (`DISTRO_PKS_SPECS-$DISTRO_BINARY_COMPAT-$DISTRO_COMPAT_VERSION` file), then please make that _one_ pr rather than a separate pr for each version.

[...]

PR #1733, opened 2020-01-06
rc.shutdown: Improve unload save file/folder

rizalmart:


01micko:
Please provide a better description re #1715 

Merging but I won't do it again!

PR #1753, opened 2020-02-17
installpkg.sh: Avoid deleting critical files

rizalmart:
Add code to avoid deleting critical files and files which was part of other packages installed upon upgrading/downgrading packages

### DIFF
--- a/woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
+++ b/woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
@@ -588,6 +588,29 @@ if [ "$DESKTOPFILE" != "" ];then
  fi
 fi
 
+CRITICALFILES="/bin/busybox
+/lib/libBrokenLocale.so.1
+/lib/libanl.so.1
+/lib/libc.so.6
+/lib/ld-linux.so.2
+/lib/libcrypt.so.1
+/lib/libdl.so.2
+/lib/libm.so.6
+/lib/libnsl.so.1
+/lib/libnss_compat.so.2
+/lib/libnss_dns.so.2
+/lib/libnss_files.so.2
+/lib/libnss_hesiod.so.2
+/lib/libnss_nis.so.2
+/lib/libnss_nisplus.so.2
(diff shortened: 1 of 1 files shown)


mavrothal:
Just stumbled upon this so here is my 2c. 
These are all files usually in the main SFS. Why any process would try to delete them (ie add whiteouts)? 
If any process try this, that process should  be fixed instead.

01micko:
I concur with @mavrothal 

This may be superfluous anyway if a roll-back of PPM is needed, which is quite probable.

s243a:
> I concur with @mavrothal
> 
> This may be superfluous anyway if a roll-back of PPM is needed, which is quite probable.

If the ppm is rolled back, could there be a branch to try and troubleshoot the current package manager? Perhaps the issues are easier to fix than one might think. I don't know what improvements might be lost by a rollback vs what features were lost in the current ppm due to  the possibly obsession with trying to simplify it. 

Regarding this commit though, these files are mostly protected by the variable:

PKG_NAME_IGNORE in /var/packages/PKGS_MANAGMENT

for instance one can protect some of these files partially by adding libc6 to this var. 

It's possible that someone might want to upgrade these packages, and in theory it could be done in pupmode 13. I bet one could even switch from pupmode 12 to 13 in a live system by appending tmpfs to the top layer and then re-evaluating all layers. Actually, I bet one could could even go from a full install to a pseudo-pupmode 13 using the same kind of trick. 

The tricky thing with a package like glibc is certain libs have to match the class loader and if they don't you'll get a segmentation fault. This means that if you upgrade glibc this way you have to make sure that you delete all incompatible libs. 

Perhaps the rules to protect these packages, should only be in place when the related PKG is part of the PKG_NAME_IGNORE variable. Another approach to make this kind of thing possible is to first swap in static versions of any utility that the package manager needs so that if the ppm messes up during process, things will still run. Maybe an sfs with the static libs could be appended to the top layer and then removed after the upgrade. This sfs would only be downloaded if one wants to update glibc.

01micko:
This wouldn't work in slacko64 anyway so is incomplete.

PR #1760, opened 2020-02-27
allow decimal in package size for ppm (puppy package manager)

s243a:
In the ppm (Puppy Package Manager) any error in the size calculation breaks the package manager and makes it refuse to install or download anything. A decimal in the pkg size shouldn't break the ppm. 

This was an issue with arch32pup but I think that peebee probably modified 0setup to remove decimals. I don't think that one should have to do this both because the size calculation will be more accurate if we use decimals and also the code is more maintainable if things like this don't break the package manager.

### DIFF
--- a/woof-code/rootfs-skeleton/usr/local/petget/installpreview.sh
+++ b/woof-code/rootfs-skeleton/usr/local/petget/installpreview.sh
@@ -260,18 +260,19 @@ $(gettext "Please cancel installation, close the Puppy Package Manager, then cli
    fi
    ADDSIZEK=0
    [ "$DEP_SIZE" != "" ] && ADDSIZEK=${DEP_SIZE%[A-Z]} #remove suffix: K M B .. etc
-   INSTALLEDSIZEK=$(( $INSTALLEDSIZEK + $ADDSIZEK ))
+   INSTALLEDSIZEK=$(echo "$INSTALLEDSIZEK + $ADDSIZEK" | bc )
    echo "$INSTALLEDSIZEK" > /tmp/petget_proc/petget_installedsizek
   done < $ONEDEPSLIST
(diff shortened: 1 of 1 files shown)


mavrothal:
So we want to call het another external function for 1kB accuracy?
This is an unnecessary I think.

s243a:
> So we want to call het another external function for 1kB accuracy?
> This is an unnecessary I think.

If this fix was in place it would have saved some time with the development of arch32pup. The fact that the ppm is so easy to break makes developers not want to maintain it and might also discourages distro builders who are trying to add a new external repo format. The fact that the ppm is so easy to break may be why arch support was dropped from woof-CE. 

I'm sure all versions of puppy have the "bc" utility, but for sanity purposes I'll check "precise light" since it is one of the lighter weight versions of puppy and maybe also a slackware variant. 

I don't think that the ppm is the best place to try to optimize puppy in terms of minimizing dependencies. Pkg is much better suited to that endevour since it already has fewer dependency requirements since it doesn't require a gui. 

In that spirit Scott created this issue for pkg:
"[Require Busybox only: remove coreutils (etc) as dependencies](https://gitlab.com/sc0ttj/Pkg/issues/64)"

So I suppose in a way this is "unnecessary" but it does improve maintainability and since there is some idea of dropping the ppm in the future in favor of pkg, the more maintainable the ppm the better for anyone who wants to add a "retro-puppy" look to their build.

s243a:
Since "bc" is a busybox applet it should be widely available:

https://boxmatrix.info/wiki/Property:bc

01micko:
Instead of making this commit woof-wide why don't you make it arch specific? PPM is slow enough as is, re @mavrothal 's post by calling externals.

Issue #1763, opened 2020-02-28
ppm doesn't update pkg installed status in GUI until GUI is restarted


s243a:
I'm just trying to trouble shoot the issue mentioned by @mavrothal in [PPM is broken? #1756](https://github.com/puppylinux-woof-CE/woof-CE/issues/1756), about the ppm (puppy package manager), not updating in the gui display with new installed package statuses. 

The most relevant place to look in the code is the search part of the gui:

```
        <entry width-request="250" activates-default="true" is-focus="true" primary-icon-stock="gtk-clear" secondary-icon-stock="gtk-find">
          <variable>ENTRY1</variable>
          <action signal="activate">/usr/local/petget/findnames.sh all</action>
          <action signal="activate">refresh:TREE1</action>
          <action signal="activate">/usr/local/petget/show_installed_version_diffs.sh & </action>
          <action signal="secondary-icon-release">/usr/local/petget/findnames.sh all</action>
          <action signal="secondary-icon-release">refresh:TREE1</action>
          <action signal="secondary-icon-release">/usr/local/petget/show_installed_version_diffs.sh & </action>
          <action signal="primary-icon-release">clear:ENTRY1</action>
        </entry>
```
[/rootfs-skeleton/usr/local/petget/pkg_chooser.sh#L467](https://github.com/puppylinux-woof-CE/woof-CE/blob/c6a99e8135bbc7209162a677700e84f694250378/woof-code/rootfs-skeleton/usr/local/petget/pkg_chooser.sh#L467)

This search control calls the function findnames.sh, which has the following code:

```
 #120827 search may find pkgs that are already installed...
 if [ -f /tmp/petget_proc/petget_installed_patterns_all ];then #precaution.
  grep -f /tmp/petget_proc/petget_installed_patterns_all -v /tmp/petget_proc/petget/filterpkgs.results > /tmp/petget_proc/petget/filterpkgs.results.notinstalled
  grep -f /tmp/petget_proc/petget_installed_patterns_all /tmp/petget_proc/petget/filterpkgs.results > /tmp/petget_proc/petget/filterpkgs.results.installed
  cp -f /tmp/petget_proc/petget/filterpkgs.results.notinstalled /tmp/petget_proc/petget/filterpkgs.results
  if [ -s /tmp/petget_proc/petget/filterpkgs.results.installed ];then
   #change category field to "complete" (display /usr/share/icons/hicolor/scalable/status/complete.svg)...
[...]

Issue #1756, opened 2020-02-22
PPM is broken?


mavrothal:
Tried Peebee's [Arch32pup](http://www.murga-linux.com/puppy/viewtopic.php?p=1049512#1049512) that is using the latest PPM in woof, with the exception of 0setup, and it appears badly broken.
In "Auto" mode gives no indication of what is going to be installed and how the installation progresses. Also You need to restart PPM to update the list of installed or removed packages. In step by step mode is even worse!. It fails when the package to be installed has dependencies met with a gtkdialog error! Also when you pick download only it goes ahead and installs the packages!
I'm assuming a lot of it is the result of b61c8df and related changes but the bottom line is that as Pkg GUI is also in bad shape, Puppy has not any acceptable  GUI for the provided package managers.

BTW 0setup can support Arch packages with the attached patch that is based on a9ccd299 with minor modifications by peebee and me.

[0setup.diff.gz](https://github.com/puppylinux-woof-CE/woof-CE/files/4239392/0setup.diff.gz)

mavrothal:
Hmm I guess I have trouble attaching the file.
Here it is then
```
--- a/0setup	2020-02-22 11:51:42.000000000 +0200
+++ b/0setup	2020-02-22 13:30:59.920151623 +0200
@@ -266,6 +266,82 @@
    #get the category...
    ${FIND_CAT} /tmp/${ONE_PKGLISTS_COMPAT}temp > $ONE_PKGLISTS_COMPAT #130126
   ;;
+  arch)
+      #130126 arch code rewritten by noryb009: call find_cat afterward...
+      #130306 BK: sub() funtion to fix version relation operators...
+      #130306 BK: filtered out bad chars from description (note tortuous way to escape ' char)...
+      LANG=${LANGORG} echo "Processing ${ONE_PKGLISTS_COMPAT} please wait..."
+      xDLFILE=${ONE_PKGLISTS_COMPAT}pre #actually it's a tarball: core.db.tar.gz, extra.db.tar.gz, community.db.tar.gz
+      rm -rf sandbox0 2>/dev/null
+      mkdir sandbox0
+      cd sandbox0
+      tar xf "../$xDLFILE"
+      RETSTAT=$?
+      cd ..
+      if [ $RETSTAT -eq 0 ];then
+       rm -f "$xDLFILE"  
+       xxDLFILE=${ONE_PKGLISTS_COMPAT##P*-}  #this will be core, extra, community  130306
[...]

PR #1764, opened 2020-02-28
Fix ppm, post install installation status in Gui. See: https://github…

s243a:
….com/puppylinux-woof-CE/woof-CE/issues/1763

Closes:
[ppm doesn't update pkg installed status in GUI until GUI is restarted #1763](https://github.com/puppylinux-woof-CE/woof-CE/issues/1763)

Partially Addresses:
[PPM is broken? #1756](https://github.com/puppylinux-woof-CE/woof-CE/issues/1756) 

I tested this in arch32pup, using autoinstall mode. I installed supertuxkart and then which I searched for tux in the search bar, the correct installation status was shown.


01micko:
In your commit message you can just write "fixes bla bla. closes #1234" and that will automatically close the referenced issue or PR (#1234 is actually about pupdial hehe)

PR #1757, opened 2020-02-24
xorgwizard: Use xorg-autoconf if available

rizalmart:
xorg-autoconf is a script that automatically generates xorg.conf with high accuracy and precision which able to configure the video card with proper xorg driver and its parameters

http://www.murga-linux.com/puppy/viewtopic.php?p=1051102

PR #1768, opened 2020-03-02
Create xorg-autoconf

rizalmart:
xorg-autoconf is a utility to generate a full xorg configuration with high accuracy and precision. It properly configures the graphics cards with proper drivers and settings depending on the video card installed on the computer. It gives also finer control for the user on some graphics cards (if a user has an experience on bash scripting). Also the generated xorg.conf file was compatible with puppy xorg utilities


01micko:
I'm happy to merge this bearing in mind my comments in #1757

PR #1770, opened 2020-03-03
Create glib-memory-alloc

rizalmart:
Some gtk+ apps look for G_SLICE environment variable values. Otherwise the app depend on these variable will not run


dimkr:
@rizalmart This script makes memory allocation and deallocation slower, and can increase memory consumption. Do you have an example of an application that needs this?

PR #1785, opened 2020-03-12
WIP: in ld.so.conf add: include /etc/ld.so.conf.d/*.conf

s243a:
This line is typically added to ld.so.conf in debian distributions. WIthout this line Arch32Pup cannot find the firefox libs because there is no post install script in the firefox package to symbolically link the somewhere in the library search path (i.e. LD_LIBRARY_PATH). 

See:
http://murga-linux.com/puppy/viewtopic.php?p=1052454#1052454

### DIFF
--- a/woof-code/rootfs-skeleton/etc/ld.so.conf
+++ b/woof-code/rootfs-skeleton/etc/ld.so.conf
@@ -6,3 +6,4 @@
 /opt/mozilla.org/lib
 /opt/samba/lib
 /root/my-applications/lib
+include /etc/ld.so.conf.d/*.conf


s243a:
I'm also discussing this issue here:
https://github.com/techrockedge/archpup32/issues/2

In the above link I note that I should also edit [/woof-code/3builddistro#L560](https://github.com/puppylinux-woof-CE/woof-CE/blob/a3e9dcecd46cc3095bea0c48ae06ceba7730798b/woof-code/3builddistro#L560), so until I make that change I will prefix this title with "WIP" for Work in Progress. The actual ld.so.conf in rootfs-skeleton, probably doesn't do anything but it is worth having it match as close as possible to the final output incase people reuse the rootfs-skeleton for other purposes (e.g. woof-next...now zwn branch)

01micko:
When this PR is no longer _WIP_ please change the title and I should get a notification.

rizalmart:
> This line is typically added to ld.so.conf in debian distributions. WIthout this line Arch32Pup cannot find the firefox libs because there is no post install script in the firefox package to symbolically link the somewhere in the library search path (i.e. LD_LIBRARY_PATH).
> 
> See:
> http://murga-linux.com/puppy/viewtopic.php?p=1052454#1052454

About your config suggestion
I made a pull requests about *.conf files in /etc/ld.so.conf.d on petget and sfs_load
in order to recognize conf files in /etc/ld.so.conf.d folder ldconfig must be run to update the ld.so.cache file

PR #1795, opened 2020-03-30
petget: Improve package entry

rizalmart:
if PKGNAME and DB_nameonly was blank use the package filename (without file extension)


sc0ttj:
What is contained in `PKGMAIN`? And where does that come from?

Can you update the PR description a bit please :) ..trying to follow along..

rizalmart:
> What is contained in `PKGMAIN`? And where does that come from?
> 
> Can you update the PR description a bit please :) ..trying to follow along..

PKGMAIN is a variable for the filename of a package without file extension

PR #1796, opened 2020-03-30
installpkg.sh: Move package files to a folder

rizalmart:
Moving package files and uninstall script to folders to organize PPM folder structure

### DIFF
--- a/woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
+++ b/woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
@@ -46,6 +46,8 @@
 #130314 install arch linux pkgs. run arch linux pkg post-install script.
 #131122 support xz compressed pets (see dir2pet, pet2tgz), changed file test
 
+[ ! -e /root/.packages/package-files ] && mkdir -p /root/.packages/package-files
+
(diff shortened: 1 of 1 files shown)


s243a:
I haven't looked at the code yet but be warned, this will have a large impact on pkg (scottman's package manager). If you want to do this then you should get the related changes approved by Scott. It should also be done in a back compatible way.

P.S. I do like the idea though. We just need to be sure that we are willing to update pkg as well and at the same time.

s243a:
One way that this could be done so that pkg doesn't need to be changed right away is to have some config variable that says where the package files dir is located.

sc0ttj:
Yeah I don't see much benefit in this change, on it's own... 

I _do_ see it would be nice to move the `.files` files into their own dir, but not really worth it IMHO...

rizalmart:
> Yeah I don't see much benefit in this change, on it's own...
> 
> I _do_ see it would be nice to move the `.files` files into their own dir, but not really worth it IMHO...

For me its very worth it. If the ppm data folder structure was organized, it was very easy to perform audit in certain situation also it prevents accidental deletion of a target file

01micko:
While Pkg is now a possibility in a woof build this PR brings an incompatibility.

PR #1798, opened 2020-03-30
remasterpup2: Improved remastering

rizalmart:
* Added support on some applications
* Remove screen resolution settings

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/remasterpup2
+++ b/woof-code/rootfs-skeleton/usr/sbin/remasterpup2
@@ -465,19 +465,24 @@ $(gettext 'Please wait...')"
 	cp -af /root/.jwmrc-tray /tmp/root/ #v411
 	cp -af /root/.fvwm95rc /tmp/root/
 	cp -af /root/.icewm/menu /tmp/root/.icewm/
+	
 	#rox desktop settings...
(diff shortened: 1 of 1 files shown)


woof-code/rootfs-skeleton/usr/sbin/remasterpup2
```diff
@@ -525,11 +588,12 @@ $(gettext 'Please wait...')"
 	#modify /root/.packages/ files, as all installed packages are now part of the new cd...
 	cp -a -f /root/.packages /tmp/root/
 	touch /root/.packages/user-installed-packages
+	
 	# 141008: move *.files to ~/.packages/builtin_files/
 	cat /root/.packages/user-installed-packages | \
 	while read -r ONEPKG
 	do
-		ONEFILE="/tmp/root/.packages/`echo "$ONEPKG" | cut -f1 -d '|'`.files"
+		ONEFILE="/tmp/root/.packages/package-files/`echo "$ONEPKG" | cut -f1 -d '|'`.files"
```
01micko: I think this is easier to read if we use the 
```shell
ONEFILE="/tmp/root/.packages/package-files/$(echo "$ONEPKG" | cut -f1 -d '|').files"
```
syntax rather than
```shell
ONEFILE="/tmp/root/.packages/package-files/`echo "$ONEPKG" | cut -f1 -d '|'`.files"
```
syntax.

Of course this is just an opinion. :smile_cat:

PR #1802, opened 2020-04-04
xwin: startup improvements

rizalmart:
Improve support for major desktop environments


01micko:
I'm happy to merge this but I think that we really need to separate starting of wm or DE from `xwin` and `/root/.xinitrc`

Issue #1750, opened 2020-02-10
f2fs could be treated as PUPMODE=12 on flash media.


01micko:
A long time ago I wrote a USB installer that modified the `init` script (in `initrd.gz`) to treat f2fs as _fast medium_ . It was a pure hack that simply replaced all instances (not many of them) of '13' with '12'. This made Puppy think that it was installed on a _hdd_ or _ssd_ so no periodic saving would occur. Saving occurred directly as it does on a normal frugal install to hdd/ssd. 

Now that _f2fs_ has matured somewhat, and the tools too, I believe it is time to treat an `f2fs` formatted partition as a normal partition like on a fast media (hdd/ssd).

bigpup1:
The whole idea of running in pupmode 13 was to limit the number of writes to a USB flash drive.
This is suppose to help make it last longer.
Reading from a flash drive has no limit.
But, writes are suppose to have a specific number of times limit, before the drive starts to show errors.
f2fs does have a lot of good features, but it still writes data, to the drive, when a write command is executed.

bigpup1:
You always can run a Puppy in pupmode 12, by simply using the pmedia=atahd command, in the boot menu entry, that is booting Puppy.
Use pmedia=usbflash or ataflash to make Puppy run in pupmode 13
I personally, would like to make the choice.
Most of the time I run Puppy, installed to any type drive (HDD,SSD,USB), in pupmode 13, by using this pmedia=usbflash command.
Save icon on desktop.
Set to no auto timed save.
Ask to save or not at shutdown.

01micko:
The operative word in the title is **could**.

It _could_ also be a boot parameter.

It _could_ also be **sdcard** and **micro-sdcard** specific and _only_ for `f2fs`.

I'm not trying to _force_ my ideas @bigpup1 but open a discussion which ~~might~~ will improve Puppy Linux.

gyrog:
When pupmode=13 was implemented, usb sticks were small, expensive and precious.
I would argue that these days usb storage devices are none of these.
For my 2 cents, pupmode=13 is an obsolete Puppy feature.
So, for me the question is;
Is it worth the effort of removing pupmode=13?

Note: The only "kindness" I do for my usb sticks and SD cards, is to format them with f2fs rather than ext4.

mavrothal:
PUPMODE 13 is handy when you want to experiment without risking your savefile integrity.
It would be nice if it remains as an option

[...]

PR #1810, opened 2020-04-29
init: improve PUPMODE refinement

rizalmart:
If the savefile drive filesystem was already a flash filesystem. Keep PUPMODE 12. Otherwise redefine it as PUPMODE 13


01micko:
I _really_ want to merge this as it addresses #1750 ... _but_ @bigpup1  does have a valid concern with the taking away of user choice.

Alternatives:
1. add a pmedia=usbflashf2fs param to override pupmode=12 and allow flash friendly filesystems to run in pupmode=13. This would be very easy to do I think.
2. define the pmedia=usbhd for flash friendly filesystems in `/usr/sbin/shutdownconfig` - a little more difficult

I think that alternative 1 would be more robust but something would have to be added to the readme for the init to publicise this measure.

Thoughts?

PR #1851, opened 2020-08-13
/etc/group: added netdev group

rizalmart:
netdev group is required for network manager daemon


01micko:
Can't you add a FIXUPHACK? I do not intend to include network manager any time soon.

rizalmart:
@01micko
Moved to fixuphack

PR #1863, opened 2020-09-03
Upgrade frisbee & pgprs: restore mobile log display, internationalize, refine code

rserwin1:
The GPRS display of the progress of a connection attempt has failed to complete in recent Puppy releases due to a regression in the Xdialog "logbox" and "textbox" options.  Once the initial line(s) were shown, subsequent lines failed to appear, implying that the connection had stalled.  The GPRS functions now use the only remaining technique that works: piping the output of the "tail --follow" command.  For easyOS, which uses the busybox version of "tail" which does not support --follow, the display is delayed to allow the full log to display.

In addition, improvements are made to the messaging related to potential conflicts between pgprs and frisbee mobile if they are used at the same time.

The other changes, which maintain the existing wifi/ethernet functionality, are to the coding format, replacement of the deprecated ifconfig and iwconfig commands, replacement of the uses of the busybox "ps" command by pgrep and ps with options, and resolution of warnings produced by the "shellcheck" tool.  The code is now fully internationalized as domains "frisbee" and "pgprs".

With the major changes in code indentation (whitespace), the git difference listings are daunting; for better detail, the kdiff3 tool produces side-by-side listings showing the changes in each line.

The "connectwizard+" commit changes the point at which pgprs makes itself the "current manager", so that its setup/cofiguration window can be viewed without impacting current networking until an attempt is made to connect.

### DIFF
--- a/woof-code/rootfs-packages/frisbee/etc/gprs.conf
+++ b/woof-code/rootfs-packages/frisbee/etc/gprs.conf
@@ -1,9 +1,5 @@
-# /etc/ppp/gprs.conf - constants for pgprs and frisbee (gprs)
+# /etc/gprs.conf - constants for pgprs and frisbee (gprs)
 # State data is retained in /root/.config/gprs.conf.
-
-# Lock file descriptor numbers controlling access by pgprs and frosbee:
-GPRS_SETUP_LOCK_FD=111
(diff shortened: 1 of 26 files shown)


01micko:
Thanks Richard

PR #1895, opened 2020-11-12
removepreview.sh: Added layer evaluation

rizalmart:
Upon removal of packages. Evaluate the layered root filesystem for non PUPMODE 2 setup

### DIFF
--- a/woof-code/rootfs-skeleton/usr/local/petget/removepreview.sh
+++ b/woof-code/rootfs-skeleton/usr/local/petget/removepreview.sh
@@ -54,6 +54,10 @@ elif [ ! "$DISPLAY" ]; then
  [ $? -ne 0 ] && exit 0
 fi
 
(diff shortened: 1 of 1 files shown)


sc0ttj:
Why? What does this fix? What was the problem before this change?

(Should I include it in Pkg?)

rizalmart:
I was playing with lxpupsc. I tried to remove a whiteout file of a builtin file on /initrd/pup_rw the builtin file doesn't show on top layer immediately. Unless issue with that command. I noticed also that udba=reval was called upon adding a package where the package contents go to save file instead of /initrd/pup_rw. So its better to call udba=reval after package removal to refresh the top layer in case it is not yet updated

01micko:
@sc0ttj does that answer your question?

sc0ttj:
I think so... so this fixes the list of packages to be removed?

If so, sounds like we need it in PPM I guess (and prob in Pkg too).

PR #1906, opened 2020-11-25
alsawizard: Added PulseAudio detection

rizalmart:
Notify user if PulseAudio was running


01micko:
Thanks for simplifying the dialog

PR #1914, opened 2020-11-25
Create asktosave_session

rizalmart:
asktosave_session is a dialog for saving session in PUPMODE 13 if autosave is not enabled

It works the same asktosave_func in /etc/rc.d/rc.shutdown but it can also create an answer file saved in /tmp/session_ramdisk_result in order to be used by /etc/rc.d/rc.shutdown script upon reboot/shutdown


01micko:
@gyrog is working on shutown changes in the `init-experiment` branch. I'll wait and see if there are conflicts when the is merged back to `testing`. Thanks

PR #1918, opened 2020-11-27
ipinfo: Improved GUI layout

rizalmart:


01micko:
Glad you didn't return to the the icanhazip code... it caused a [flame war](http://oldforum.puppylinux.com/puppy/viewtopic.php?p=748910&search_id=1019167890&sid=2a3795307bd264b9a0410550907e0149#748910l) over several threads for a long time.

PR #1921, opened 2020-11-27
Update DISTRO_PKGS_SPECS-slackware-14.2

rizalmart:
Added alsa plugins


01micko:
I am reluctant to pull these for a few reasons
- most of the libasound* libs in the package have hard dependency on libpulse - understandable but then full pulseaudio and deps must be included.
- retrovol would need to be removed - pavaudio (? volume control?) or similar would need to be included
- pequaliser probably wouldn't work as it is
- maybe pmusic the same
- older versions should be kept light, however I'm willing to consider pulse for 14.2 when bluetooth 5 stuff becomes more viable and deal with the fallout later

PR #1937, opened 2020-12-01
installpkg.sh: Some improvements

rizalmart:
* Add second pass to scan for files in package's installed file and perform some actions
* If /etc/rc.d/PUPSTATE was missing, use PUPMODE=2. This allows to use PPM in pristine root filesystem under chroot process


01micko:
since #1937 depends on #1935 I'd prefer if they were one PR

Issue #1847, opened 2020-08-08
Asus C201 Support


dimkr:
These small and light ARMv7 Chromebooks no longer receive updates from Google, and they're perfectly capable of running Puppy.

I got an old 2GB model to boot Puppy, using https://github.com/dimkr/woof-CE-c201.

Wi-Fi with a AR9271 USB dongle works fine, audio doesn't (don't know why) and persistent installation shouldn't be a problem, because a 2GB image can easily fit a 16GB image that can be written to the eMMC using dd (see https://github.com/dimkr/devsus#usage).

For some these laptops are junk, but they're also a resource. They can be donated to disadvantaged children that lost access to education due to the pandemic, unlike students that have a computer and take advantage of remote learning.

Duplicate of #605.

01micko:
Hi @dimkr .. long time no see.

Now that I have some experience developing [raspup](http://raspup.eezy.xyz) and [rasvuan](http://01micko.com/13-Rasvuan-Puppy-Alpha-Release) and built many packages the infrastructure of woof is more likely to be more accommodating for other `arm` processor projects.  Unfortunately, raspberry pi boards are the opposite of 'free' thanks to the choice of the board devs choosing broadcom SOC.

It would be nice to get my hands on a c201.. we'll see.

dimkr:
Actually, I have another Chromebook, the one I was talking about in https://medium.com/@dima_8716/go-development-on-a-99-arm-laptop-26e32d27670b. As far as I remember, it should receive updates until 2025 and it shouldn't be very hard to add support for it (plus all other models that use the same SoC) too, because most of the work is getting the kernel to build (and multiple projects already rebuild the Chrome OS kernel - Arch Linux ARM for example) and the creation of a bootable image.

01micko:
Ok, so there are a number of arm chromebooks http://www.linuxmadesimple.info/2019/08/all-chromebooks-with-arm-processors-in.html. That's just a 2020 list.

This hisense one looks interesting .. https://archlinuxarm.org/platforms/armv7/rockchip/hisense-chromebook-c11

So.. armhf or aarch64? That is the question

If I got hold of one I'd have to source it overseas, too expensive here.

dimkr:
[...]

PR #1948, opened 2020-12-09
Asus C201 support (#1847)

dimkr:
These thin, light and inexpensive Chromebooks stopped receiving Chrome OS updates on June 2020. But that doesn't mean they're trash, because 1) they still can provide reasonable web browsing experience 2) Linux, Mesa, etc' still work great on them and 3) there are distros that work well on these laptops.

I decided to port Puppy Linux to these laptops because I want something with a good "out of the box" experience with an application for every common task, something that doesn't take much space on the SSD, and something lightweight. Puppy used to excel at these things in the days where computers with 1 GB of RAM or less were common, but with the stagnation of Moore's law and the large amount of RAM in computers from the last decade (the Lenovo X220 from 2011 had 4 GB!), some of that magic got lost, and this is something I'll need to compensate for.

The Raspberry Pi work by @01micko this port much easier, because general ARM support is already in place. However, the coreboot+deptcharge combo makes things difficult, and so does the Chrome OS partition layout.

Therefore, Puppy Linux support for the C201 (and ARMv7 Chromebooks, in general) consists of:
- [frugalify](https://github.com/dimkr/frugalify), a solution that allows something similar to a "frugal installation" with a partition instead of a savefile, *although there's no initramfs involved in the boot process*
- PUPMODE=6, similar to 5 ("full installation") in many ways, but supports SFS loading (~~e.g. sfs_load~~ through frugalify)
- Support for generation of file system images containing the Chrome OS partition layout
- ~~Some fixes for cases where Puppy assumes that Puppy Linux on ARM == Puppy Linux on a Raspberry Pi~~ already in testing
- ~~Support for swap over [zram](https://www.kernel.org/doc/Documentation/blockdev/zram.txt), which can increase the usable life of the 2 GB RAM variant of the C201; swap on a slow, small and cheap SSD doesn't help much, it's inconvenient and can be unhealthy for the SSD~~ already in testing

~~I'm doing my best to cherry-pick fixes and features that are not specific to the C201, so this PR doesn't become huge and impossible to review.~~ already have

Known issues:
- [x] ~~General slowness - everything feels slow (window dragging, browsing, everything is unresponsive during package installation)~~ X is using the fbdev driver instead of using Panfrost and the screen doesn't update very often
- [x] Missing applications
- [x] ~~Boot partition not mounted with `noatime`~~ Needs testing
- [x] Wi-Fi interface is unstable?
- [x] Missing ath9k_htc firmware
- [x] Missing 16GB SSD image for [Devsus-style](https://github.com/dimkr/devsus#usage) installation
- [x] sfs_load fails because the SFS mount point is under aufs and must be outside of it - need to figure out a solution for this
- [x] quicksetup thinks this is a Raspberry Pi
- [x] ~~Consider upgrading to kernel 5.10, it's longterm too and Panfrost should have gotten better since 5.4~~ Aufs still doesn't support 5.10
- [x] For some reason boot fails when support/cros_image.sh creates the image using exactly the same procedure as Devsus does, so for now, cros_image.sh downloads blank images generated by Devsus CI and fills them with Puppy
- [x] Lots of junk and unneeded stuff in DISTRO_PKGS_SPECS
- [x] JWM menu is broken
- [x] ~~Does the webcam work?~~ The webcam doesn't work
- [x] ~~Does audio work?~~ Audio doesn't work
- [ ] Should [frugalify](https://github.com/dimkr/frugalify/) be moved to puppylinux-woof-CE?
- [x] Kernel panic when SFS files at / are symlinks
- [x] ~~CI job that builds C201 kernels/images?~~ *WIP* in [my fork](https://github.com/dimkr/woof-CE-c201/actions?query=workflow%3AC201)
- [x] Kernel building is slow, need to add ccache support
- [x] Consider implementing an "OTA update" mechanism that downloads the latest SFS+kernel from the latest GitHub release
- [x] When shutting down through the menu, there's a prompt and the laptop reboots (???)
- [x] The sparse 16 GB image inside the 2 GB one doubles the download size - the 2 GB image contains the SFS and a 16 GB sparse file containing exactly the same SFS; is there any way to cut the size by half without decompressing on first boot?
- [x] Check if 3D games other than Neverball work
- [x] Add zram-backed swap and make sure Firefox/Chromium can eat 4 GB of "RAM" on the 2 GB variant of the C201, without crashing or freezing the system
- [x] ~~The battery monitor tray icon doesn't work~~ The battery monitor applet requires ACPI
- [x] Add compton (or another compositor?) to fix tearing when dragging windows
- [x] Internet connection is lost when the firewall is started
- [x] delayedrun sleeps for 20 seconds before starting picom, and rendering is sluggish until then
- [x] After rebasing the kernel configuration against the Chrome OS 4.19 one, the kernel boots to a blank screen and the laptop reboots
- [x] Pre-built binaries and old packages need to be replaced with fresh, optimized packages built on Debian Bullseye, ~~a repo with CI/CD that produces binary packages would be nice~~ using petbuilds
- [x] JWM buttons are invisible
- [x] Some types of windows are semi-transparent because that's JWM's/picom's default, and rendering becomes slow
- [x] netmon_wce doesn't display icons
- [x] Firefox refuses to start
- [x] After the addition of fonts-liberation2, the image size has jumped from 187 to 232 MB and needs to be cut down again
- [x] Unused package repos need to be removed from DISTRO_*
- [x] Double menu entries for Geany, l3afpad, etc'
- [x] Some gtkdialog based applications don't work with GTK 3 - at least, sfs_load
- [x] No pulseudio
- [ ] I/O can be slow at times - it's unclear whether CONFIG_SQUASHFS_DECOMP_MULTI_PERCPU feels faster than CONFIG_SQUASHFS_DECOMP_SINGLE
- [x] The arrows in comboboxes and other symbolic icons are missing because adwaita-icon-theme is not installed, but this package is big
- [x] The update script (pm6_update) needs GUI
- [ ] rc.update is not triggered after update
- [x] pm6_update thinks it should update from cversion x to version x
- [x] pm6_update needs to use releases from the woof-CE repo cloned, not puppylinux-woof-CE/woof-CE
- [x] Firefox runs as root, not spot
- [ ] elogind is missing
- [x] ~The firewall is disabled by default~ ~- testing the fix~
- [x] ~/run is written to the save partition and should be a tmpfs~ ~- testing the fix~
- [x] #2074 - WIP
- [x] ~#2073~ ~- testing the fix~
- [ ] Login as finn doesn't work 
- [x] The iptables->iptables-legacy symlink is missing (?)
- [x] ~urxvt needs to be replaced with LXTerminal, for working copy-paste and tabs~ - ~WIP~
- [x] ~Audio in Firefox doesn't work #2118~ testing the fix
- [ ] Special keys like volume keys don't work
- [ ] Packages are downloaded over HTTP
- [ ] gtk3-nocsd is missing
- [x] Applications need to be moved to adrv
- [x] Once applications are moved to adrv, setup-spot fails and applications run as root
- [x] The default image viewer is not set
- [x] on ARM, pm6_update should choose an update for BOOT_BOARD and not any ARM release


woof-code/support/cros_image.sh
```diff
@@ -0,0 +1,74 @@
+wget -O- https://github.com/dimkr/devsus/releases/latest/download/devsus-templates.tar.gz | tar -xzf-
```
dimkr: TODO: produce the partition layout + ext4 file system here, instead of downloading a blank image

woof-code/rootfs-skeleton/usr/sbin/sfs_load
```diff
@@ -1756,7 +1756,7 @@ case "$PUPMODE" in
 		else SFSMODE="tmp"
 		fi
 		;;
-	12|13) SFSMODE="y";;
+	6|12|13) SFSMODE="y";;
```
dimkr, line 1806: TODO: get sfs_load to work; the whole point in the 2/6/12 distinction is allowing sfs_load to work in 6

woof-distro/arm/debian/bullseye/packages-templates/firmware-brcm80211/lib/firmware/brcm/brcmfmac4354-sdio.txt
```diff
@@ -0,0 +1,155 @@
+# Sample variables file for BCM94354Z NGFF 22x30mm iPA, iLNA board with PCIe/SDIO for production package
```
dimkr: TODO: see if I can download this file from the Chrome OS kernel repo during the build, instead of holding a copy

woof-distro/arm/debian/bullseye/_00build.conf
```diff
@@ -0,0 +1,150 @@
...
+
+## include devx SFS in ISO?
+DEVX_IN_ISO=no
+
+## Include the windows puppy installer LICK by Luke Lorimer aka <noryb009>
+LICK_IN_ISO=yes
+
+## compression method to be used (SFS files)
+#SFSCOMP='-comp xz -Xbcj x86 -b 512K'
+SFSCOMP='-comp xz -Xbcj arm,armthumb -b 512K'
```
dimkr: TODO: how does the block size affect compression ratio and decompression speed?

woof-code/rootfs-skeleton/usr/sbin/bootmanager
```diff
@@ -327,7 +327,7 @@ XML_LABEL=$(gettext 'Run pupswap')
 XML_ACTION='pupswap &'
 SWAPFILE_XML="$(generic_xml)" #--
 
-if [ $PUPMODE -eq 2 ];then #PUPMODE=2, full hd install, will not have this.
+if [ $PUPMODE -eq 2 -o $PUPMODE -eq 6 ];then #PUPMODE=2, full hd install, will not have this.
```
dimkr: TODO: this is the wrong check. The UI is misleading and says SFS loading is unsupported.

01micko:
I'm thinking that sound support may depend on pulse-audio. I have a HP tablet that has an atom SoC and sound fails on that and research seems to indicate  that pulse is needed.

dimkr:
> I'm thinking that sound support may depend on pulse-audio. I have a HP tablet that has an atom SoC and sound fails on that and research seems to indicate that pulse is needed.

Thanks for the help @01micko, pulseaudio is definitely in the backlog, and I'll take a look at sound and the webcam issues once I'm done with https://github.com/dimkr/woof-CE/pull/4.

dimkr:
@01micko pavucontrol is huge! It pulls in many libraries, and all of its alternatives need gtkmm 😠

01micko (commented):
I'm thinking to use dynamic svg (or cairo) icons in powerapplet by getting the percentage then figuring out the coordinates. Similar to how the various temperature applets I have made do it. I tend to lean toward svg, it can be a single method parsing appropriate integer input. (well they'll be unsigned shorts actually :smile_cat: )

woof-code/support/cros_image.sh
```diff
@@ -0,0 +1,152 @@
...
+
+	cp -f build/vmlinuz /mnt/legacyimagep1
+	extlinux -i /mnt/legacyimagep1
+	dd if=/usr/lib/EXTLINUX/mbr.bin of=${LOOP}
+	cat << EOF > /mnt/legacyimagep1/extlinux.conf
+DEFAULT puppy
+
+LABEL puppy
+	LINUX vmlinuz
+	APPEND root=PARTUUID=$PARTUUID init=/init rootfstype=ext4 rootwait rw
```
dimkr, line 146: @01micko Took some time to figure out how to achieve this, but now I have working initrd-less boot!

The build process for dpup Bullseye produces 2 GB images with the Chrome OS partition layout (ARM + x86_64) and the x86_64 version also produces a legacy BIOS image for Chromebooks that have SeaBIOS.

It boots just fine on "normal" PCs too, puts an overlayfs at / and it's very similar to a frugal installation, but without an initrd.

woof-code/packages-templates/firefox_FIXUPHACK
```diff
@@ -11,7 +11,6 @@ defaultPref("browser.uidensity", 1);
 
 // render website text while fetching images
 defaultPref("nglayout.initialpaint.delay", 0);
-defaultPref("layers.progressive-paint", true);
```
01micko: So _that_ is the culprit.

When I find time I'll do a new slacko 8 build from from `c201-v2`

zigbert:
while reading the known issues, I accidentally clicked a checkbox (missing applications). I clicked it again, hence the result of "zigbert edited the comment"... No intention to mess up your work ;)

PR #1957, opened 2020-12-15
hide the ndiswrapper option in SNS if ndiswrapper is missing

dimkr:


rizalmart:
Any network related stuff puppy was under @rserwin1 you need to message him in case you have changes on network components of puppy

rserwin1:
@dimkr,
Thank you for those fix recommendations.  I am adding the "Hide" fix to SNS 2.4.1.
Regarding the gtkdialog syntax error, I am looking at making the "buttons" line getting the error conditional, rather than greying out the 'Windows' button.
I will attribute the fix ti "dimkr" unless you prefer your puppy forum ID if you have one.

dimkr:
Thanks @rserwin1, feel free to this this differently; my concern is not wasting pixels and fooling users into thinking ndiswrapper is an option, when it's not. My first fix was to remove the "Windows" button, but this causes a syntax error because the `hbox` may have no children elements (i.e. if there are zero network interfaces), so I grayed it out.

Regarding credit, I don't have a forum ID in the new forums, and I find this UX fix trivial enough for me not to receive credit :)

PR #1964, opened 2020-12-20
snapmergepuppy: Exclude more folders

rizalmartin:
Exclude these folders when saving to save file/folder
/run
/var/run
/var/lock


s243a:
If I recall correctly, some packages create sub directories in "/var/run" and if these sub directories aren't saved then on the next boot one will get an error about the directory not existing.

01micko:
@s243a @rizalmartin `/run` also

This was the reason that `/run` was changed to it's own dir as used to be a symlink to `/tmp` which gets emptied every time at reboot.

PR #1969, opened 2020-12-21
in cross-builds, try to chroot into rootfs-complete using qemu-user-static

dimkr:
This makes the first boot on ARM faster, and makes the savefile cleaner.

### DIFF
--- a/woof-code/3builddistro
+++ b/woof-code/3builddistro
@@ -515,7 +515,11 @@ rm -f rootfs-complete/etc/profile.d/*.csh* 2>/dev/null # slackware 13.1
 #-----------------------------------------
 #sanity check...
 echo
-if [ "$WOOF_HOSTARCH" = "$WOOF_TARGETARCH" ];then #111123
+
+QEMU=`command -v qemu-${WOOF_TARGETARCH}-static`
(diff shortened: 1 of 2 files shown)


01micko:
So... have to build qemu statically.. do you have a recipe?

I like the idea so it gets merged but I want to try it.

dimkr:
> So... have to build qemu statically.. do you have a recipe?

There's a Debian package, qemu-user-static. These are static binaries, so you can steal them from Debian. But you'll also need to configure binfmt_misc to run qemu-arm-static for executables that start with the ELF header+architecture field==ARM.

In my ARM Chromebook port PR, I added a GitHub actions flow that runs woof-CE on Ubuntu, so I don't have to build after to each commit myself. [It installs qemu-user-static and binfmt-support](https://github.com/puppylinux-woof-CE/woof-CE/pull/1948/files#diff-acd220db5bbc5cc0c44fbe1cb66058cb1e75a2aea3f97eab1a0cd22fdf9f9b98R16), which takes care of configuring the kernel to run ARM executables via qemu-user-static. Each time I want to flash an image to my Chromebook, I have a CI/CD job that pushes it to https://github.com/dimkr/woof-CE/releases, and that's how I tested the qemu-user-static trick in 3builddistro.

PR #1972, opened 2020-12-21
make pup-advert-blocker use github.com/StevenBlack/hosts and run it in 3builddistro

dimkr:


01micko:
Ok, a problem running the `pinstall.sh` in chroot

```
--2021-01-03 05:18:37--  https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts
Could not seed PRNG; consider using --random-file.
Disabling SSL due to encountered errors.
```

No help searching. Can't we run the wget part in the running OS and concatenate the file into `sanbox3/rootfs-complete/etc/hosts`?

dimkr:
That's interesting. Maybe /dev needs to be bind-mounted to rootfs-complete/dev before doing this?

Somehow, it works for me with ARM builds, and I'm using a chroot.

dimkr:
OK, I'll see if I can make /usr/sbin/pup-advert-blocker write to ./etc/hosts and not /etc/hosts, then git rid of that chroot.

PR #1976, opened 2020-12-25
Create gvfs_FIXUPHACK

rizalmart:
Disable some gvfs components which cause some problems or conflicts on puppy


01micko:
`which` will only operate on the host system anyway - this needs a rethink so I'm closing.

PR #1989, opened 2021-01-03
Run pup-advert-blocker without chroot, to avoid SSL issues

dimkr:
Tested it by running pinstall.sh inside rootfs-complete, and it seems to work.


01micko:
Pretty much what I was thinking to do as a work-around.

PR #1992, opened 2021-01-03
restore the blank /etc/resolv.conf from before #1972

dimkr:
This is a purely cosmetic change.

One might argue that having a pre-configured DNS server may cause leak of private data (DNS queries) to a specific DNS server, if you set up your internet connection manually, in the short time window between the `ifconfig` and the modification of `/etc/resolv.conf`. Most people use DHCP, though.

### DIFF
--- a/woof-code/rootfs-skeleton/etc/resolv.conf
+++ b/woof-code/rootfs-skeleton/etc/resolv.conf
@@ -1,1 +1,1 @@
-nameserver 8.8.8.8
+# nameservers go in here


dimkr:
Closed by mistake

01micko:
Doesn't really matter because `dhcpcd` overwrites it anyway... here's the one from my fresh slacko 7 save after shutdown.

```
# Generated by dhcpcd
# /etc/resolv.conf.head can replace this line
# /etc/resolv.conf.tail can replace this line
```

PR #1997, opened 2021-01-10
alsawizard: Added redirection for pavucontrol

rizalmart:
If the puppy has pulseaudio was running Redirect to pavucontrol if it was installed

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/alsawizard
+++ b/woof-code/rootfs-skeleton/usr/sbin/alsawizard
@@ -14,7 +14,7 @@ if [ "$(pidof pulseaudio)" != "" ]; then
  if [ "$(which pavucontrol)" != "" ]; then
   exec pavucontrol
  else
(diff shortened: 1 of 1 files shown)


01micko (changes requested):
in the splash message perhaps change **was** to **is** 

Actually, pavucontrol should be mandatory if pulseudio is installed else sound cards are only configurable with `pactl`... but maybe someone can make a gui for that.

01micko:
What also annoys me about pulseaudio/pavucontrol is that there is no light weight tray status icon available. Sure there is [pasystray](https://github.com/christophgysin/pasystray) - which does ship in Slackware - however it requires QT which in turn requires mariadb. Hardly light. I suppose a systray launcher would suffice for pavucontrol, maybe indicating status in the tooltip by polling `pactl`.. a job for me I guess :laughing:

01micko:
Ok, :smile_cat: but of course pulse still needs work to be included in stock puppy.

PR #2000, opened 2021-01-15
add a picom package template with auto-start

dimkr:
Puppy feels smoother, faster and more luxurious with vsync :)

### DIFF
--- a/woof-code/packages-templates/picom_FIXUPHACK
+++ b/woof-code/packages-templates/picom_FIXUPHACK
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+mkdir -p root/.config/autostart
+
+cat << EOF > root/.config/autostart/picom.desktop
+[Desktop Entry]
+Encoding=UTF-8
+Type=Application
+NoDisplay=true
+Name=picom
+Exec=picom -b --backend=glx --unredir-if-possible --vsync
+EOF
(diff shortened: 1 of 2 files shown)


mrfricks:
hi @dimkr, good to see you back!!

i really like picom and have it in fossapup although this caused many a headache during development!

using glx with nouveau caused %50 cpu usage so if nouveau is detected it switches to xrender (no vsync).

also there's some oddities with the modesetting driver for both intel & nouveau where certain dialogs may not appear (puppy save dialog) and certain apps not work (googleearth)

i feel that puppy needs a more sophisticated graphics setup that includes picom but adjusts it accordingly depending on the driver. 

e.g. when the nvidia driver is installed it turns off vsync and suggests using the inbuilt nvidia driver vsync.

i made a basic picom gui for fossapup here http://distro.ibiblio.org/puppylinux/pet_packages-fossa64/picom-gtk-1.0.pet

dimkr:
Thanks @mrfricks, this is really good feedback.

Should I limit autostart to ARM then? Alternatively, I can check what Mesa driver is in use (don't know how to do that, but I'll figure this out) and decided whether or not to start picom, and with what options.

dimkr:
@mrfricks I restricted auto-start to ARM, at least until we come up with a better solution.

PR #2006, opened 2021-01-20
syntax error in ptheme pinstall.sh

dimkr:


dimkr:
Tested in https://github.com/dimkr/woof-CE/actions/runs/498449692

PR #2008, opened 2021-01-22
make the creation of a kernel sources SFS optional

dimkr:


01micko (commented):
Can you add a block in `build.conf` please, even if it is just a comment

PR #2013, opened 2021-01-26
rc.update: Add GTK4 support

rizalmart:
Use gtk4-update-icon-cache in gtk-update-icon-cache was missing


dimkr:
This needs to happen during 3builddistro as well!

CollaboratorGCM:
Wow! You guys are Great to consider this single move in PUPs from 2 to 4.

Issue #2011, opened 2021-01-26
zstd support on busybox rpm


rizalmart:
I noticed that PPM failed to install rpm's of newer distro because their rpm is now using zstd. We need to update the busybox in order to extract on zstd compressed rpm.

PR #2018, opened 2021-01-28
installpkg.sh: Improve installation of RPM's

rizalmart:
Use rpm2cpio to extract rpm packages


01micko:
Does this solve the zstd issue in #2011 ?

rizalmart:
> Does this solve the zstd issue in #2011 ?

Probably fixed if full rpm binaries was installed.

01micko:
Yes seems latest busybox isn't supporting zstd without patches.

PR #2031, opened 2021-02-04
Make PulseAudio+pa-applet work on Bullseye

dimkr:
There's a .desktop file that runs start-pulseaudio-x11, which runs pactl, which performs D-Bus calls.

Normally, D-Bus auto-starts pulseaudio via systemd, but we don't have that, and using a separate "session bus" instance of D-Bus is problematic when it's root that's logged in. Therefore, I forced start-pulseaudio-x11 to start pulseaudio itself with a wrapper script.

Also, I made sure pulseaudio is started before pa-applet, which doesn't display anything if this is not the case.

Firefox audio now passes through pulseaudio and volume can be changed via the pa-applet tray icon.

### DIFF
--- a/woof-code/packages-templates/pulseaudio_FIXUPHACK
+++ b/woof-code/packages-templates/pulseaudio_FIXUPHACK
@@ -17,6 +17,7 @@ chmod 755 usr/local/bin/pavu.sh
 
 cat > usr/local/bin/ppavol << EOF
 #!/bin/ash
+type pa-applet >/dev/null 2>&1 && exit
 #todo gettext
 if ! type yad >/dev/null 2>&1 ; then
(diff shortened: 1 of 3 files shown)


woof-code/packages-templates/pulseaudio_FIXUPHACK
```diff
@@ -17,6 +17,7 @@ chmod 755 usr/local/bin/pavu.sh
 
 cat > usr/local/bin/ppavol << EOF
 #!/bin/ash
+type pa-applet >/dev/null 2>&1 && exit
```
01micko, line 20: good idea

01micko (approved):
Changes look fine to me.

PR #2039, opened 2021-02-13
init: Improvements on Copy to RAM SFS

rizalmart:
Allocate only a quarter of RAM for copying SFS files to RAM.
When SFS file can no longer able to copy to ram. Just mount the SFS file from partition


peabee:
Why is this an "improvement"?
What problem are you trying to solve?
Where have the problem(s) been reported?
Why is this necessary if swap is available??

rizalmart:
> Why is this an "improvement"?
> What problem are you trying to solve?
> Where have the problem(s) been reported?
> Why is this necessary if swap is available??

This improvement is for low ram situation. Not all systems have swap partition. Also I add failsafe to avoid copy sfs on RAM recklessly as Puppy became bigger overtime due to software trends.

Setting quarter of physical ram as limit for copying SFS on ram is enough.

PR #2043, opened 2021-02-18
Cache the slacko64 kernel between CI runs

dimkr:


dimkr:
@01micko This PR is really hard to test, I'll trigger several builds of testing to make sure kernel-kit runs only once.

PR #2047, opened 2021-02-20
Cache petbuilds across CI runs and publish ISO+devx

dimkr:
Manually triggered builds produce a tag+GitHub release with ISO and devx:

Screenshot 2021-02-20 at 16 03 35

### DIFF
--- a/.github/workflows/slacko64.yml
+++ b/.github/workflows/slacko64.yml
@@ -25,6 +25,14 @@ on:
   schedule:
     - cron: '0 0 * * 4'
   workflow_dispatch:
+    branches:
+      - ci
+      - testing
+    inputs:
+      release_name:
+        description: 'Release name'
+        required: true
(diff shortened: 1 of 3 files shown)


01micko (approved):
Yes zstd needs to be the new compression default in distros that support - which is all modern ones.

dimkr:
Let's see how much faster it is, from what I see with dpup it's a huge improvement both in 3builddistro, and when starting large applications like Firefox.

With caching in place, a slacko64 build should be 8-11 minutes now, like https://github.com/puppylinux-woof-CE/woof-CE/actions/runs/584986297. I hope that zstd can save at least 2 minutes.

01micko (approved):

01micko (approved):
:+1:

01micko (approved):

Issue #2050, opened 2021-02-21
xdg-puppy-jwm build failure


dimkr:
```
make[2]: Leaving directory '/tmp/gnome-menus-2.14.3/desktop-directories'
make[1]: Leaving directory '/tmp/gnome-menus-2.14.3/desktop-directories'
Making install in util
make[1]: Entering directory '/tmp/gnome-menus-2.14.3/util'
if /ccache gcc -DHAVE_CONFIG_H -I. -I. -I.. -DGMENU_I_KNOW_THIS_IS_UNSTABLE -DGNOMELOCALEDIR=\""/usr/share/locale"\" -I./../libmenu -I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include  -DG_ENABLE_DEBUG -Wall -Wmissing-prototypes      -march=x86-64 -mtune=generic -Os -fomit-frame-pointer -ffunction-sections -fdata-sections -fmerge-all-constants -MT test-menu-spec.o -MD -MP -MF ".deps/test-menu-spec.Tpo" -c -o test-menu-spec.o test-menu-spec.c; \
then mv -f ".deps/test-menu-spec.Tpo" ".deps/test-menu-spec.Po"; else rm -f ".deps/test-menu-spec.Tpo"; exit 1; fi
/bin/sh ../libtool --tag=CC --mode=link /ccache gcc  -march=x86-64 -mtune=generic -Os -fomit-frame-pointer -ffunction-sections -fdata-sections -fmerge-all-constants  -Wl,--gc-sections -Wl,--sort-common -Wl,-s -o gnome-menu-spec-test  test-menu-spec.o  ../libmenu/libgnome-menu.la -lglib-2.0  
mkdir .libs
/ccache gcc -march=x86-64 -mtune=generic -Os -fomit-frame-pointer -ffunction-sections -fdata-sections -fmerge-all-constants -Wl,--gc-sections -Wl,--sort-common -Wl,-s -o .libs/gnome-menu-spec-test test-menu-spec.o  ../libmenu/.libs/libgnome-menu.so -lglib-2.0
creating gnome-menu-spec-test
make[2]: Entering directory '/tmp/gnome-menus-2.14.3/util'
test -z "/usr/bin" || mkdir -p -- "/usr/bin"
make[2]: Nothing to be done for 'install-data-am'.
[...]

PR #2051, opened 2021-02-21
Fix build failure of xdg-puppy-jwm on slacko64 (#2050)

dimkr:


01micko (approved):
Thanks for fixing the specs as well :smile:

PR #2055, opened 2021-02-23
Remove menu entries that don't work in barebones puplets

dimkr:
The ARM Chromebook images don't have many preinstalled applications and there's a number of menu entries that point to applications that don't work without third-party dependencies.

These applications either fail silently, or ask the user to install these dependencies. All Puppies have these dependencies and if they're left out in DISTRO_PKGS_SPECS, this is probably intentional and the menu entries should be removed in 3builddistro.


01micko:
21eadcf needs a better condition I think because raspup supports sfs loading and an sdcard image is built

dimkr:
Thanks! Removed that commit.

PR #2057, opened 2021-02-24
Remove empty submenus and extra separators in the JWM menu

dimkr:
In a barebones puplet, the games menu is present but has zero items, and if a submenu has zero items, the separator after it is still present:

menus


01micko:
Yeah I need that one too. Thanks

PR #2067, opened 2021-03-01
Automate the release of a slacko 7.0 successor with fixes

dimkr:
@01micko What do you think about releasing a 7.1 with fixes for critical issues in 7.0 soon? I can get this process automated, and 7.x can serve as additional testing ground for woof-CE.

(of course, the build workflows added in this branch need to pass first)

### DIFF
--- a/.github/workflows/slacko-7.yml
+++ b/.github/workflows/slacko-7.yml
@@ -72,14 +72,14 @@ jobs:
         (echo 2; echo 2; echo 3; echo; echo) | sudo -E ./merge2out
     - name: Set version
       if: github.event_name == 'workflow_dispatch'
-      run: sudo sed -i s/^DISTRO_VERSION=.*/DISTRO_VERSION=${{ github.event.inputs.version }}/ ../woof-out_x86_64_x86_slackware64_14.2/DISTRO_SPECS
+      run: sudo sed -i s/^DISTRO_VERSION=.*/DISTRO_VERSION=${{ github.event.inputs.version }}/ ../woof-out_x86_64_x86_slackware_14.2/DISTRO_SPECS
     - name: 0setup
       run: |
-        cd ../woof-out_x86_64_x86_slackware64_14.2
+        cd ../woof-out_x86_64_x86_slackware_14.2
         sudo ./0setup
     - name: 1download
       run: |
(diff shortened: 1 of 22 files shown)


01micko:
@dimkr the slacko-7 build fails because we are building on HOSTARCH=x86_64, TARGETARCH=x86 but ..

```
cd: ../woof-out_x86_x86_slackware_14.2: No such file or directory
```

dimkr:
> @dimkr the slacko-7 build fails because we are building on HOSTARCH=x86_64, TARGETARCH=x86 but ..
> 
> ```
> cd: ../woof-out_x86_x86_slackware_14.2: No such file or directory
> ```

Thanks, fixed!

I also made DISTRO_VERSION configurable (using a `sed`), so DISTRO_SPECS doesn't have to be changed every time we want to release a slacko 7.x bugfix release. I believe this should make 7.x a "zero-maintenance" release - every time a user complains about a critical bug we'll be able to issue a release after fixing the bug in woof-CE, without having to change anything in the woof-distro files of 7.0.

01micko:
Slacko 7 complains because FHSM is enabled in AUFS. Perhaps use the 4.19 DOTconfig?

dimkr:
Seems to work now, and CI is much faster! Published several releases in my fork and ~~the size looks correct~~ booted one. It's important not to forget to add that `-`, and I forgot to do that in one of my test releases 😊 

flow

screeny

The releases are marked with `prerelease: true`, because maybe we'd like to see if the release is sane on real hardware, then edit the release and untick the "this is a pre-release" checkbox:

pre

What do you think, @01micko? I can change to `prerelease: false` and the new release will appear at the top of the list of release, watchers of the repo will get notified, etc', without that manual step.

01micko:
Been following closely and it looks good! I think changing prerelease is a good idea for 7.

dimkr:
`build/boot/isolinux/isohybrid: error while loading shared libraries: libuuid.so.1: cannot open shared object file: No such file or directory`

Happens only in the 32-bit build 🤔

01micko (approved):
@dimkr , can you merge in `testing` (1 commit outstanding) and I'm happy to merge back this PR and deal with any consequences.

01micko:
Done!

Issue #1755, opened 2020-02-18
/run/sshd get's deleted upon reboot. 


s243a:
This is probably a woof-CE issue so I'll report it here:

> I installed ssh-server on DpupBuster. The folder /run/sshd seems to get deleted upon reboot. I recreated this folder with the standard permissions...that you get by creating a new folder in the file manager. Not sure if these permissions are correct. For the ssh server /run/sshd is needed when using the privilege separation option on your ssh server. Privilege separation is the default behavior for the ssh server in openssh. This is probably a woof-CE bug, so I'll also report it on github.
http://murga-linux.com/puppy/viewtopic.php?p=1050603#1050603 

The same issue probably applies to other folders in the /run folder. I remember a related issue on TazPup.

01micko:
/run is a symlink to /tmp  so this is why. In puppy /tmp has always been cleared between boots.

01micko:
I reinstated /run as a real directory. In my slackware install it has 0755 perms. If anyone knows better, feel free to adjust perms on /run.

PR #2080, opened 2021-03-05
Mount a tmpfs at /run

dimkr:
/run contains temporary files and they pollute the save file.


01micko:
`/run` was a symlink to `/tmp`. This was changed because there were complaints about settings being lost after reboot.

s243a:
> `/run` was a symlink to `/tmp`. This was changed because there were complaints about settings being lost after reboot.

Good catch :). For instance there was this issue:

`/run/sshd get's deleted upon reboot.  `#1755 

So maybe, rather than mounting tmpfs at /run, there could be some script to clean old pid files on shutdown or boot.

dimkr:
Something here is weird - if the files are supposed to be saved persistently, what are they under /run? It's a tmpfs in every distro.

Will investigate later, closing for now.

PR #2081, opened 2021-03-05
Allow EXTRA_COMMANDS to enable the firewall by default

dimkr:
File sharing through LAN, Puppy servers, etc' are uncommon, and having the firewall already enabled by the time you connect to the internet is a sane default.

### DIFF
--- a/woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
+++ b/woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
@@ -394,8 +394,28 @@ export GUI="<window title=\"Firewall Setup $version\" icon-name=\"$winicon\">
 		</hbox>
 	</vbox>
 </window>"
-eval $(gtkdialog -p GUI -c 2>/dev/null) 2>/dev/null
-echo $EXIT
+if [ "$1" = "enable" ]; then
+	echo "Enabling firewall"
+	EXIT=Doit
+	GENERIC=true
+	LOGGING=false
+	MAIN=true
+	EX=false
+	CONFIG_SSH=false
+	CONFIG_CUPS=false
+	CONFIG_SAMBA=false
+	CONFIG_SAMBA_CLIENT=false
+	CONFIG_DLNA=false
+	CONFIG_NTP=false
+	CONFIG_FTP=false
+	CONFIG_HTTP=false
+	CONFIG_DNS=false
+	CONFIG_DHCP=false
+	CONFIG_NFS=false
+else
+	eval $(gtkdialog -p GUI -c 2>/dev/null) 2>/dev/null
+	echo $EXIT
+fi
 case "$EXIT" in
 	Doit)echo "GENERIC=$GENERIC
 LOGGING=$LOGGING
@@ -1318,7 +1338,7 @@ echo "Load rules for mangle table ..."
 EOF_REST
 
 if [ -f "$TMPFW" ];then
-	[ $state = 1 ] && /etc/init.d/rc.firewall stop && sleep 1
+	[ $state = 1 -a "$1" != "enable" ] && /etc/init.d/rc.firewall stop && sleep 1
 	echo "copying firewall"
 	cp -af "$TMPFW" /etc/init.d/rc.firewall
 	chmod 755 /etc/init.d/rc.firewall
@@ -1327,6 +1347,8 @@ else
 	echo "Something went wrong"
 	exit
 fi
+
+[ "$1" = "enable" ] && exit
 
 # run it!
 gtkdialog-splash -bg yellow -timeout 2 -text "Starting firewall" &


CollaboratorGCM:
I am not so sure I agree, respectfully. There are some of us how run a sane home network without a need for firewall OOTB.

A better option **might** be to place a firewall option in FirstRUN where it is plainly visable and allows user to choose to leave disabled **OR** to enable a default; dont you agree?

I do understand the concern yu raise.

dimkr:
There are 4 reasons for doing so:

1. Even users with a home server benefit from having a firewall that blocks everything except the ports they really want to expose to other devices in the LAN. It's not true that people who run a home server must disable the firewall completely, although it is a way to prevent the firewall from blocking your Samba 🤷🏿‍♂️ 
2. Most users don't use Puppy as a home server.
3. Most users can benefit from being protected by default, when connecting to the internet for the first time and it happens to be a public WiFi network. Users can easily miss that firewall checkbox in SNS, and there is a small time window between the WiFi authentication and the moment the firewall is started that may be enough to attack a vulnerable service like Samba.
4. Quicksetup already has a checkbox for the firewall; I'm just proposing to change the default from off to on. The firewall can be disabled, and it's as easy as it is to enable the firewall right now.

CollaboratorGCM:
Not sure I would concur, but that's OK. I do understand you mean well.

Since it is a change from the traditional pattern it will be missed/overlooked leading to issues in PUP use.

If you follow-thru, then ALSO change FirstRUN to reflect the F/W is "ON"
such that users can see to turn off. Otherwise it is hidden and will be a
source of confusion in a number of ways leading to time not well spent in diags.

That will help all, I think.

dimkr:
> If you follow-thru, then ALSO change FirstRUN to reflect the F/W is "ON"

The checkbox reflects that the firewall is on by default, and so does the firewall tray icon (that big, green shield).

CollaboratorGCM:
Thx!

s243a:
As a side note, I personally run puppy with the firewall off. The router/gateway provided by my ISP blocks open ports to the internet by default and having the firewall off is convenient if I want to ran a lan services such as samba or ssh. That said, it is easy enough to uncheck the firewall checkbox on the first run wizard settings so this change is not big deal to me.

bigpup1:
The firewall 
If on when you first boot Puppy.
It can interfere with setting up the network connection for WIFI.
That is why firewall is off, the first time you boot Puppy.
In the past, the firewall was automatically on, and a lot of people had issues getting network connection setup correctly. Not everyone, but way to many.
So, turning off firewall for initial first boot, seemed to be the easy fix.
Depending on what network connection program you use.
After getting a working connection. It should be warning you to turn on the firewall.

CollaboratorGCM:
This is why I have brought up the idea of FirstRUN use. Maybe the time to
activate the F/W is after FirstRUN or in such a way that after FirstRUN the
user is reminded of F/W activation.

The idea is, up until now, is that the F/W is a user choice via FirstRUN
and beyond vs a non-user choice. Other present-day subsystems may be
impacted by this deliberate change from tradition. It took a lot of time,
obviously looking through the forum's past, of subsystems changes to get
Puppy to its current point of getting boot processing to stable desktop
operation.

If this moves forward, it is hoped that no undue efforts by development
will be needed to support this change, even as it may appear to be minor.

Again, I did not raise the concern because of tradition, rather, it is
about unforeseen disruption which may emerge requiring development
reaction. So,expressing it differently "what could be the penalty if it is
left "off" as it is?"

BTW, as I dont have 360 degree viewing: Is there a back-end subsystem that is being changed that requires a F/W to be running that is NOT mentioned, thus far?

dimkr:
> As a side note, I personally run puppy with the firewall off. The router/gateway provided by my ISP blocks open ports to the internet by default and having the firewall off is convenient if I want to ran a lan services such as samba or

Indeed, but this is not something you want to do with a laptop that's occassionally connected to untrusted WiFi networks.

I think that having the firewall on by default is the sane default, assuming the user is a non-technical user that doesn't use Puppy as a server. I might be wrong, and that's what I want this discussion to happen and put it in a pull request.

Maybe I'll just make it a configuration option, so Puppy builders will be able to turn on the firewall by default.

> It can interfere with setting up the network connection for WIFI.

This is very useful feedback. Do you have an example of a real world scenario? WPA2, DHCP, etc' work when the firewall is on, wired connection is just DHCP, and I wonder what exactly breaks.

CollaboratorGCM:
Again, for many/most, **is the unseen potentials and possible time+work effort** worth a change in tradition?

I am not aware of anyone complaining about an exposure that they have control over; ... yet.

This POST is a mere note for consideration. If you feel the effort is worth the development exposure, then let run with and we'll have to deal with issues that "may" arise.
 
Personally, I alway know to turn on the F/W when out of home over the years.
 
To be honest, with cell phones what they are, I have NOT used a laptop in a public location in 5 years because I use my phone. And for over a year, the Covid has negate any external use other than my phone use.

My kids/grands use their tablets, not PCs. Those I see in public places like coffee shops have work issued laptops.

WE are in a new age (smile).

dimkr:
Closing this PR until I make this change optional.

dimkr:
This PR now allows builders to enable the firewall by default, using `firewall_ng enable` in EXTRA_COMMANDS.

Objections, anyone?

PR #2091, opened 2021-03-09
rc.sysinit: Added isolated folder

rizalmart:
Due to unorthodox puppy scripts which has filenames having conflicts with the mainstream distro package. It can cause troubles when it was overwritten. The solution here was to put puppy scripts on a isolated folder and set it as first priority on $PATH this will prevent puppy scripts from being overwritten upon updating packages from mainstream distro

### DIFF
--- a/woof-code/rootfs-skeleton/etc/rc.d/rc.sysinit
+++ b/woof-code/rootfs-skeleton/etc/rc.d/rc.sysinit
@@ -37,7 +37,7 @@
 #130504 moved up, i think fbcon needs to be loaded before intel, nouveau modules load.
 #160609 rerwin: Add wait for USB3 driver.
 
-PATH=/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/X11R7/bin
+PATH=/lib/puppy/bin:/lib/puppy/sbin:/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/X11R7/bin
 . /etc/rc.d/functions_x
 
 #================================================================


01micko:
I think this idea has merit and would merge _but_ I want feedback from @dimkr @peabee @mrfricks @zigbert @gyrog

dimkr:
@rizalmart Do you have examples for this, other than conflicts between coreutils/findutils/util-linux/.. with busybox?

gyrog:
1. Are there any examples of this being a real problem?
2. Who's going to get all the contributors of Puppy scripts to use these new locations?
3. What about all the existing ".pet's" and existing ".sfs's"?

I guess my concern, is that this could create a "guilt" response on contributors to spend a significant amount of time going through all their scripts to relocate them, with maybe not much real benefit.
Should I be worried that someone might find my ancient 'shareinternet' pet?

Or do we accept that the benefit may be a long time comming, but at least the change is "cheap"?

gyrog:
Even with the patch there's still a problem.

Let's suppose I create a top Puppy script called "geany", (also assume that the real "geany" is not present by default), and place my binary in /lib/puppy/sbin to avoid any clashes.
I then download the "geany" application via PPM, everything is stil fine, no clashes.
But then when I click on the desktop icon for the downloaded "geany" I get my top Puppy script, not the downloaded application.
Why?
Because the downloaded ".desktop" file for the real "geany" contains:
"Exec=geany"

The only real solution to such a name clash is for the name of the Puppy script to be changed so that it is unique, e.g. "pgeany"

peabee:
I don't see any need for such a change - where are the problem reports to show it has been causing problems over these many years? And there may indeed be unintended consequences. I vote against.

s243a:
Aren't we just talking about a few core scripts such as df and mount as well as perhaps a few scripts (and config files) related to the init proceas?

The one exception I can think of is pmount. There is a conflicting debian utility called pmount.

Anyway, my understanding is the intent here is to reduce the likelyhood of error in the update process but it could also make it easier for people developing alternative build systems and/or package managers.

rizalmart:
> 1. Are there any examples of this being a real problem?
> 2. Who's going to get all the contributors of Puppy scripts to use these new locations?
> 3. What about all the existing ".pet's" and existing ".sfs's"?
> 
> I guess my concern, is that this could create a "guilt" response on contributors to spend a significant amount of time going through all their scripts to relocate them, with maybe not much real benefit.
> Should I be worried that someone might find my ancient 'shareinternet' pet?
> 
> Or do we accept that the benefit may be a long time comming, but at least the change is "cheap"?

@gyrog, @dimkr 

My concern here was the puppy core scripts there are core puppy scripts conflict with the mainstream linux package:
df
losetup
mount
umount
ps
reboot (symlinked to wmexit)
poweroff (symlinked to wmexit)
shutdown
init

If these scripts was overwritten by the scripts the puppy might fail to start. Slackware-based puppy has an update system. If  the puppy  core scripts  were overwritten by the mainstream package updates. puppy might fail to work.

Adding an isolated folder which prevents the puppy core scripts from overwriting when installing the updates from the parent distro where puppy is made of.

Also adding an isolated folder allows developers to override some components without getting overwritten unless the package directed to overwrite files on that isolated folder

> The only real solution to such a name clash is for the name of the Puppy script to be changed so that it is unique, e.g. "pgeany"

If it was the first place thatwill be no problem. Because it was not implemented in the first place, it's a different story.


> Let's suppose I create a top Puppy script called "geany", (also assume that the real "geany" is not present by default), and place my binary in /lib/puppy/sbin to avoid any clashes.
I then download the "geany" application via PPM, everything is stil fine, no clashes.
But then when I click on the desktop icon for the downloaded "geany" I get my top Puppy script, not the downloaded application.
Why?
Because the downloaded ".desktop" file for the real "geany" contains:
"Exec=geany"

About this geany scenario. You just override the geany by putting the script on the isolated folder. The isolated folder is for protecting puppy core scripts from being overwritten. But it can be also used for overriding some programs or applications. It's up on user/developer whether to override something or not.

dimkr:
Can't we solve this problem using fake installations? e.g. `yes|util-linux|util-linux|exe>null,dev>null,doc>null,nls>null` would prevent installation of util-linux and possible replacement of Puppy scripts

rizalmart:
> Can't we solve this problem using fake installations? e.g. `yes|util-linux|util-linux|exe>null,dev>null,doc>null,nls>null` would prevent installation of util-linux and possible replacement of Puppy scripts

@dimkr 
I think it doesn't help. Also it makes building puppy more hassle.
When installing package updates. It only reads /var/packages/user-installed-packages not /var/packages/woof-installed-packages
Creating isolated folders will enable puppies to update packages if they want especially those security updates.

dimkr:
> Creating isolated folders will enable puppies to update packages if they want especially those security updates.

Personally, I think updates are not a problem if updates are image-based (SFSs are replaced and not the individual packages inside them), and that's how updates straight from GitHub work in #1948. Actually, I don't see how to implement updates through PPM, if packages don't pass through FIXUP, etc'.

rizalmart:
> > Creating isolated folders will enable puppies to update packages if they want especially those security updates.
> 
> Personally, I think updates are not a problem if updates are image-based (SFSs are replaced and not the individual packages inside them), and that's how updates straight from GitHub work in #1948. Actually, I don't see how to implement updates through PPM, if packages don't pass through FIXUP, etc'.

For me, some users want to update some packages which was part of builtin not the whole sfs image.

I noticed that on TazPuppy when some users after they installed it. They applied packages updates immediately. Although TazPuppy was different from Puppy. I noticed that user behavior.

Package updates was offered by updater program not by the PPM. Best example here was updates_mgr which can be found on Slackware-based Puppies.

peabee:
Only Slacko has updates_mgr as far as I know....if this is dangerous then it should be fixed in Slacko.
Each Pup uses a different update mechanism....
I agree with @dimkr that updates should replace complete sfs components which have been tested to be consistent and working.

01micko:
`updates_mgr` is long deprecated.

gyrog:
My point with the geany example, was to show that simply using different directories and placng both the directories in the "PATH" does NOT solve the issue of what happens when  a "geany" command is entered in a console or a script, or a .desktop file, as many non-Puppy applications, and Puppy scripts, do.  The PATH will always resolve to the first one found., i.e. one of the programs can only be executed by specifying it's full path. The only real solution is to change the name of the Puppy script.

if we are talking about things like "mount", then this seems to be  heading in the direction of no longer having a "mount" and a "mount-FULL" in the normal location, to having a a Puppy " mount" in a Puppy location and a "mount-FULL", (now just "mount") in the normal location.

But it seems that we are not talking about these things, but rather a more basic Puppy thing, i.e. How close of an integratiion to the upstream Linux is Woof-ce supposed to achieve?
As I recall, it was never meant to be 100%, but a means to provide Puppy wilth access to the huge number of applications available in upstream repositories, but for the core of Puppy to remain Puppy, no matter if it's a uPUP or a dPUP or an sPUP, etc...
That's why Puppy does not support the use of upstream kernels, and has retained the PPM, (for all it's problems).
(Woof-ce's biggest problem is that it has been too successful.)
Perhaps we need to reflect on the significant difference between Puppy and "FatDog" etc...

As to the actual patch, I think it is just as likely to cause as much time wasting for current contributors to Pupp Linux, as it is to avoid problems in the future.

Issue #2074, opened 2021-03-04
Applications running as spot use the default GTK theme


dimkr:
Maybe only true to my dpup, Firefox or GTK 3

rizalmart:
> Maybe only true to my dpup, Firefox or GTK 3

Did you put gtkrc file on /home/spot?
gtk3 themes can be set by default by putting gtk.css on /etc/gtk-3.0

dimkr:
> Did you put gtkrc file on /home/spot?

Nope.

As a user, I expect applications to use the theme I chose, and the fact my browser runs as spot is an implementation detail.

PR #2102, opened 2021-03-11
Set the GTK theme for all users (#2074)

dimkr:


01micko:
Also cursor-theme and rox icon theme need looking at

dimkr:
> Also cursor-theme and rox icon theme need looking at

This is true, but setting the GTK and icon themes is enough to make browsers running as spot visually consistent with other applications, and that is my focus here.

I prefer to leave the problem of finn/real multi-user support in ptheme to somebody more involved in ptheme (@zigbert?).

PR #2103, opened 2021-03-11
Run Firefox, Sylpheed and Transmission as spot

dimkr:


bigpup1:
What is the problem you are trying to fix???

That has been possible, for a long time, in Puppy Linux.

**It is the choice, of the user, to do it or not do it.**

Login and Security Manager, usually offers the option to Run-AS-Spot, for different internet programs, that are installed in the Puppy version.
If program is not listed.
A terminal command:
`run-as-spot (name of program exec file) `will run it as spot.

I wonder if you truly, completely, understand, what run as spot actually does.

dimkr:
> What is the problem you are trying to fix???

Running the browser as root, as we all know, is a security nightmare. Running the browser as root is not a sane default, especially for users that don't understand what "login manager" and "run as spot" mean.

This is a PR, not a change pushed directly into testing without asking anyone. It is open for discussion, because I know some other developers have different opinions.

> I wonder if you truly, completely, understand, what run as spot actually does.

Yes, I do.

Ideally, I'd like to run everything as non-root, but it's a large change and I'm busy with other things now.

01micko:
@bigpup1 I have  a high opinion of @dimkr 's work and since the new year he has brought an _enormous_ improvement to puppy, and that isn't just my opinion.

That said, he also has commit rights to this repo (and others owned by `puppylinux-woof-CE`) but is offering this as a pull-request so other devs have a chance to review the code _and_ the ethos (as stated).

He has been around puppy ~11 years? (correct me if wrong @dimkr ) and has written many utilities for Puppy including some low level woof utilities in C and other programming languages.

He was known on the old forum as 'Iguleder'. Yes he is a young bloke (in the context of us old blokes :smile_cat: ) but that can only be a good thing for Puppy and Linux in general going forward. We need more interest from young people. We've lost much talent over the years for being a bit too conservative (in the context of distro development) - I could rattle off a list - search the old forum - a start is `tombh`, `pizzasgood` [both young talented people]. I would call @bkauler 's ethos to Linux development rather more _radical_.

I'd almost wave this through but it does deserve some decent discussion.

peabee:
I agree with @bigpup1 - we should retain the status-quo - the Puppy ethos is to constrain users as little as possible unless THEY decide they want to be constrained. The mechanisms to constrain apps to running as spot exist and work but should be a user choice to use.

> Running the browser as root, as we all know, is a security nightmare.

This is just theory - has there been any instance of an actual problem on Puppy Linux?

> I'd like to run everything as non-root

This would be a terrible change and completely unecessary for a single user desktop system. It would make Puppy just like any other Linux distro - putting artificial barriers into the way we work for no good reason.

dimkr:
> @bigpup1 I have a high opinion of @dimkr 's work

😊, but this is software engineering, not art. My work should be judged by its usefulness (i.e. protection of the user's private data) and not by the credentials of its author.

> This is just theory - has there been any instance of an actual problem on Puppy Linux?

Every time you download an email attachment, every time you download a torrent, etc', as root, there is a small chance that someone ran malicious code on your machine and installed persistent malware that records your keystrokes. Maybe you haven't seen any evidence for this, but this doesn't mean it never happened, and I can't prove that this risk has ever manifested in actual violation of your privacy.

> This would be a terrible change

This change has advantages, and they come at the price of some inconvenience here and there (root vs. /home/spot).

Personally, I understand that most users spend most of their time in the browser, and have to download files, or open downloaded files, considerably less frequently than 10 years ago, when stuff like Google Drive and Google Docs wasn't as popular as today.

Other than this home directory annoyance, I don't see big downsides to this change and truly believe the increased security/privacy is worth it.

Consider the fact this change will be part of the Chromebook port of Puppy that I'm working on, its intended audience and the high degree of security and robustness this audience is used to, when using Chrome OS.

--- 

If running applications as spot has zero value, why do we support this? Telling people it's safe to run the browser as root, then expecting them to tick the not-so-easily-accessible "run as spot" checkbox and assuming they understand what "running as spot" means, doesn't make much sense to me. I bet that if we start publishing beta releases of slacko, with some telemetry mechanism that collects user settings and data like the list of installed packages, we'll see that almost nobody uses this feature.

If you object strongly to this PR, even if it's for a reason I disagree with, I'll close it. No offense.

peabee:
> If you object strongly to this PR, even if it's for a reason I disagree with, I'll close it. No offense.

@01micko is the arbitrator

dimkr:
(I don't mind enabling spot in EXTRA_COMMANDS, only for dpup)

peabee:
That would be a good approach ;-)

01micko:
This should be a global option, like the firewall solution

PR #2109, opened 2021-03-12
Make the GTK+ 3 gtkdialog more backwards-compatible

dimkr:
This should many (most? all?) old gtkdialog packages work with the GTK+ 3 build of gtkdialog, by replacing combobox with combobxotext, and tree or list with table. Same as the fixes that went into woof-CE, but at runtime.

https://github.com/puppylinux-woof-CE/gtkdialog/pull/92

### DIFF
--- a/woof-code/rootfs-petbuilds/gtkdialog/compat.patch
+++ b/woof-code/rootfs-petbuilds/gtkdialog/compat.patch
@@ -1,0 +1,75 @@
+From e570075d19cf2e90a2d5160e510e661deb5aaa30 Mon Sep 17 00:00:00 2001
+From: Dima Krasner <dima@dimakrasner.com>
+Date: Fri, 12 Mar 2021 09:21:02 +0200
+Subject: [PATCH] automatically replace deprecated widgets
+
+---
+ src/automaton.c | 13 +++++++++++++
+ 1 file changed, 13 insertions(+)
+
+diff --git a/src/automaton.c b/src/automaton.c
+index 080284c..af27140 100644
+--- a/src/automaton.c
++++ b/src/automaton.c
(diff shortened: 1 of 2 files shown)


zigbert:
<table> can not replace <tree> in general - at least not in gtk2. If script is taking advantage of the many attributes and options of the tree widget, table will fall short - if not the table in gtk3 is more mature than the one in gtk2. Test the Tray manager in jwm_config. It will be a worthy testbench.

- Move items in list by drag'n' drop
- click will allow manual editing of item
- right-click will show menu

The tree-widget seems to work ok with gtk3, - or am I missing someting. If tree is not supported by gtk3, and the only option is to use table, I only see gtkdialog as a cut-down version of itself.

dimkr:
@zigbert This change is meant to increase the number of old gtkdialog based scripts that work with the GTK 3 build of gtkdialog. Right now, gtkdialog just exits if a script uses a widget removed in GTK 3.

Instead, this patch makes gtkdialog replace the missing widgets with the alternative that's most likely to work:

- combobox -> comboboxtext
- list -> tree
- table -> tree

If this works well, we don't need to rename gtkdialog to gtkdialog-gtk3 or something, use this name everywhere in woof-CE, and keep a separate GTK 2 build named "gtkdialog" for backwards-compatibility.

EDIT: sorry, now I see I have a mistake in my original description of the change. We replace list and table with tree, not tree and list with table.

01micko:
I tagged a release `0.8.4c`

zigbert:
No problem if List and Table becomes Tree. After reading first comment, I understood it the other way around...

dimkr:
@01micko I'll make the petbuild use the new release, and drop the patches.

Issue #1860, opened 2020-08-25
builtin_files problem


rizalmart:
I noticed that on the entries of DISTRO_PKGS_SPECS like this for example

`yes|xserver_xorg|xorg-server,package-a,package-b|exe,dev
`

When the packages become a builtin, Its database entries are successfully entered on the /var/packages/woof-installed-packages file. However the list of files installed by the packages merged as one package entry based from the entry in DISTRO_PKGS_SPECS

for example if a sample DISTRO_PKGS_SPECS entry shown above was installed. The builtin-files was
`
/var/packages/builtin_files/xserver_xorg
`

Instead of
`
/var/packages/builtin_files/xorg-server
`
`
/var/packages/builtin_files/package-a
`
`
/var/packages/builtin_files/package-b
`

Which is the correct one.

Is there anyway to fix this problem on builtin_files.sh of woof? If the problem was fixed it was much easier to audit files and repack a packages.

sc0ttj:
I think I have encountered this issue when using Pkg too - trying to find which package a file comes from is made tricker in these cases, if the package in question is a built-in.. 

I would generally argue that the package names listed in ./built-ins/, and the listed file contents of each package, should reflect the repo packages as accurately as possible.

sc0ttj:
I think I have encountered this issue when using Pkg too - trying to find which package a file comes from is made tricker in these cases, if the package in question is a built-in.. 

I would generally argue that the package names listed in ./built-ins/, and the listed file contents of each package, should reflect the repo packages as accurately as possible.

sc0ttj:
I think I have encountered this issue when using Pkg too - trying to find which package a file comes from is made tricker in these cases, if the package in question is a built-in.. 

I would generally argue that the package names listed in ./built-ins/, and the listed file contents of each package, should reflect the repo packages as accurately as possible.

sc0ttj:
Puppy version: FossaPup64 9.5

I was trying to find out which package the `/usr/sbin/puppyapps` script comes from:

- it is not listed as an installed package (built-in)

![puppyapps-notfound](https://user-images.githubusercontent.com/2726610/97730256-895a1e80-1acb-11eb-9dcb-2d234e50cbaf.png)

I tracked it down, by knowing that the 'noarch' repo "is a thing"... Other users (beginners) might not know about it.. 

[...]

PR #2135, opened 2021-03-18
Fix and enable fossa64 adrv support

dimkr:
I couldn't understand why every time I built fossa64, GCC failed during 3builddistro because libgmp was missing. gmp wasn't copied into rootfs-complete, and it does include the correct, libgmp10 dependency of GCC:

```
yes|gmp|libgmp10,libgmpxx4ldbl,libgmp-dev,libgmp3-dev|exe,dev,doc,nls| #in precise, this was only in devx, but abiword needs it.
```

I couldn't understand why it's not copied, if there's a `yes` there, and why every fossa64 build fails when I enable adrv.

Now, I finally found it. The problem is here:

```
        DUMMY=XXXX #unless they make a Qld beer proggy 'XXXX' will never match; besides need a trailing field
        EXTRAS_LIST="${EXTRAS_LIST// /|}${DUMMY}"
        EXTRAS_LIST="${EXTRAS_LIST//^|}" #strip leading "|"
        EXTRAS_LIST="${EXTRAS_LIST//||/|}" #replace double "||" with "|"
fi
if [ -n "${DUMMY}" ];then
        PKGLIST="`echo "$PKGS_SPECS_TABLE" | grep '^yes' grep -vE "${EXTRAS_LIST}" | cut -f 2 -d '|' | tr '\n' ' '`"
```

`EXTRAS_LIST=abiword|gnumeric...` and the `gmp` line contains one of these substrings, `abiword`, **in a comment**, so gmp is removed from the list of packages that go into rootfs-complete.

### DIFF
--- a/woof-code/3builddistro
+++ b/woof-code/3builddistro
@@ -265,9 +265,9 @@ else
 	EXTRAS_LIST="${EXTRAS_LIST//||/|}" #replace double "||" with "|"
 fi
 if [ -n "${DUMMY}" ];then
-	PKGLIST="`echo "$PKGS_SPECS_TABLE" | grep '^yes' | grep -vE "${EXTRAS_LIST}" | cut -f 2 -d '|' | tr '\n' ' '`" 
+	PKGLIST="`echo "$PKGS_SPECS_TABLE" | grep '^yes' | sed s/#.*//g | grep -vE "${EXTRAS_LIST}" | cut -f 2 -d '|' | tr '\n' ' '`"
 else
-	PKGLIST="`echo "$PKGS_SPECS_TABLE" | grep '^yes' | cut -f 2 -d '|' | tr '\n' ' '`"
+	PKGLIST="`echo "$PKGS_SPECS_TABLE" | grep '^yes' | sed s/#.*//g | cut -f 2 -d '|' | tr '\n' ' '`"
 fi
 PKGLIST="${PKGLIST}"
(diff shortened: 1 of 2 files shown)


rizalmart:
@dimkr 
Hello can you also please fix this woof-ce issue #1860 ?

dimkr:
> @dimkr
> Hello can you also please fix this woof-ce issue #1860 ?

I'll take a look later, got some things to fix after this fix is merged.

rizalmart:
> > @dimkr
> > Hello can you also please fix this woof-ce issue #1860 ?
> 
> I'll take a look later, got some things to fix after this fix is merged.

Thanks a lot. The woof-ce builtin script was too cryptic to me.

dimkr:
> Thanks a lot. The woof-ce builtin script was too cryptic to me.

Yes, many scripts in woof-CE are hardly readable. Automated builds of each pull requests make it easier to fix them without breaking something by accident, so I'm not afraid to touch that area once I stabilize adrv-related stuff.

dimkr:
There are multiple forum topics that discuss modularity and attempts to build a minimal fossa64. IMHO most are pointless theoretical discussions.

@mrfricks This PR and #2138 make woof-CE capable of building a fossa64 that has the same set of libraries as 9.5, but with almost all applications moved to adrv. This is done without manual intervention in the build process, and this build has the latest "huge" 5.4.x kernel and fdrv, so it should be as modular as 9.5, but reproducible.

What do you think?

PR #2137, opened 2021-03-18
Convert README to markdown and add build badges

dimkr:


01micko:
Thanks! I've been meaning to do this.

Issue #2033, opened 2021-02-05
Support for symlinks in merge2out


dimkr:
I've started working on dpup Bullseye for x86_64, which takes advantage of all the work I put into #1984 and #1948. Almost pure GTK 3 setup, zstd compression for all SFS images, initrd-less boot, pulseaudio working out of the box and more goodies.

The only differences between the package lists of x86_64 and ARM are one firmware package needed for the ARM laptop I'm targeting as the first supported ARM device, and the binutils package name.

I added an architecture-dependent part to DISTRO_PKGS_SPECS that appends these packages. However, merge2out doesn't support symlinks, so I can't make the two DISTRO_PKGS_SPECS point to a common location and reduce the amount of duplication.

PR #2148, opened 2021-03-20
Add a barebones dpup bullseye template

dimkr:
This should be almost identical to what's produced by #1948, but a regular ISO with savefiles, etc'.

Once #2033 is fixed, the DISTRO_PKGS_SPECS files can become symlinks.

There **will** be many missing packages required for core functionality like the installers - for example, syslinux.

### DIFF
--- a/.github/workflows/bullseye64.yml
+++ b/.github/workflows/bullseye64.yml
@@ -1,0 +1,209 @@
+name: bullseye64
+
+on:
+  push:
+    branches:
+      - ci
+      - testing
+    paths:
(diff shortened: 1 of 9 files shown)


dimkr:
Publishing a test build in https://github.com/dimkr/woof-CE/runs/2155759280

EDIT: https://github.com/dimkr/woof-CE/releases/tag/dpupbullseye64-9.0.0-prepreprealpha1

01micko:
Looking good!

dpup

PR #2150, opened 2021-03-20
Make the retro 4.3.1 theme more useful

dimkr:
Minimal_Labelled is the "new standard" and the huge number of desktop icons will probably drive users away from 431.

Blue-Moon is old and has many missing icons. If we fall back to any other icon theme for those missing icons, applications look inconsistent. And even if we do that, the pinboard and the JWM menu use different icons.

The only problem is that the 431 desktop background doesn't work great with the blue-green folder icon, but it's a small problem compared to the current visual mess.

EDIT: also switched to the xf tray layout, which is another "new standard". It just works better.


01micko:
I'll try it soon :)

PR #2155, opened 2021-03-21
Add busybox vi and set EDITOR=vi if nano is missing

dimkr:
#2041, #2110, #2131, #2151, etc', make dpup a very nice Puppy for advanced users BUT there's no vi!

### DIFF
--- a/woof-code/support/busybox_symlinks.sh
+++ b/woof-code/support/busybox_symlinks.sh
@@ -58,7 +58,7 @@ do
 		# klibc-utils
 
 		# Editors
-		vi|ed)    continue ;;
+		ed)       continue ;;
 		patch)    continue ;; # too primitive
 
 		# Finding Utilities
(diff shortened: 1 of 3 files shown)


woof-code/support/busybox_symlinks.sh
```diff
@@ -100,7 +100,7 @@ do
 	esac
 	#-
 
-	if [ -f bin/$N -o -f sbin/$N -o -f usr/bin/$N -o -f usr/sbin/$N ] ;then
+	if [ -e bin/$N -o -e sbin/$N -o -e usr/bin/$N -o -e usr/sbin/$N ] ;then
```
dimkr, line 103: This change prevents collision with other `vi`s like vim, where /usr/bin/vi may be a symlink

01micko: Makes sense

Issue #2156, opened 2021-03-21
The LXTerminal menu should be hidden only when -e


dimkr:
https://github.com/puppylinux-woof-CE/woof-CE/commit/8a2308b5b131f4d1c511ba2294afa93bdab14a8f

PR #2157, opened 2021-03-21
Restore the lxterminal menu bar (#2156)

dimkr:
At first, I tried to hide the menubar only if `-e` is specified. However, we can't symlink rxvt->lxterminal and get rid of rxvt-unicode because there are extra arguments lxterminal doesn't understand, like `-bg`. The motivation behind hiding the menu was making it easier to get rid of rxvt-unicode without breaking anything, but it looks like it won't be that easy.

lxterminal is only used interactively and rxvt-unicode is still used for all `-e` use cases, so restoring the menubar, which is useful only for interactive shells, makes sense.


peabee:
Looks OK but bit surprised that a patch is still needed - surely the default is to show the menu?

dimkr:
> Looks OK but bit surprised that a patch is still needed

The other patch changes the default background and foreground colors to match those of rxvt-unicode, for visual consistency.

> surely the default is to show the menu?

Yes.

01micko:
@peabee I think you misunderstood .. the patch was to _remove_ the menubar and that patch is now removed.

PR #2159, opened 2021-03-21
Default to the 431 global theme in dpup

dimkr:
This will make screenshots distinguishable from Slacko 8.x screenshots, and give the GTK+ 3 port of Polished-Blue extra testing.


dimkr:
Too many missing applications, the bottom tray is not useful at all.

PR #2160, opened 2021-03-21
If remove_sublevel=yes, change the bugfix version to 999

dimkr:
.0 is confusing or scary for users, who might think they have zero bug and security fixes. .999 looks weird, but at least it's not scary.

### DIFF
--- a/kernel-kit/build.sh
+++ b/kernel-kit/build.sh
@@ -606,7 +606,7 @@ fi
 cp Makefile Makefile-orig
 if [ "$remove_sublevel" = "yes" ] ; then
 	log_msg "Resetting the minor version number" #!
-	sed -i "s/^SUBLEVEL =.*/#&\nSUBLEVEL = 0/" Makefile
+	sed -i "s/^SUBLEVEL =.*/#&\nSUBLEVEL = 999/" Makefile
 fi
 ## custom suffix
 if [ -n "${custom_suffix}" ] ; then


01micko:
I think 'scary' is subjective. I agree that `0` may mislead users but maybe `999` would be scary to some. Just `99`?

We could probably grok out the _real_ kernel version easy enough.

```
# UNAME_R=`ls /etc/modules | tail -n1 | cut -d '-' -f 2-`
# echo $UNAME_R 
```

dimkr:
The problem with .99 is that it can be a real version. For example, 5.4 is now at 5.4.107. It has to be either 0, or something super high.

01micko:
Yeah, I that's true.

peabee:
I don't understand why you don't want to tell people which kernel they are using????

dimkr:
If the kernel version is, say, 5.4.1, people who install third-party drivers like nvidia drivers won't be able to update to kernel 5.4.2, because they use different directories and verification of the driver version will fail.

If we change the version number to 5.4.0, the user can upgrade to any later 5.4.x version without breaking anything.

Distros like Debian and Ubuntu have special mechanisms like dkms, that rebuild third-party drivers during installation, or every kernel version x has a matching driver y for version x package. We don't have this.

01micko:
Here's a hacky idea... using code to similar to what I posted above we could append the 'real' kernel to `uname -a` but leave everything else for backwards compatibility
```
Linux ralf11028 5.10.0 #1 SMP Sun Mar 21 12:58:29 AEST 2021 x86_64 AMD Ryzen 5 1600 Six-Core Processor AuthenticAMD GNU/Linux (real kernel version 5.10.25)
```
This could be activated _only_ if we use a 'sublevel=yes' kernel in `3builddistro`

dimkr:
Yes, putting the real version in the hostname or something is a good idea. Need to see if there's a clean way to patch the Makefile.

PR #1984, opened 2021-01-01
Add optional support for building missing packages from source in 3builddistro

dimkr:
Can cross-compile ARM packages on x86_64, thanks to #1969!

@sc0ttj @01micko @peabee What do you think about this feature? I find it much easier to build packages directly from woof-CE (with cache, of course, so it doesn't happen in every 3builddistro), instead of having to manage a repo.

Extra packages (= not pre-installed ones) can be handled by the petbuilds repo, of course, and I think we can set up a CI/CD flow in GitHub Actions that builds all PET packages, then pushes a GitHub release that contains .pet packages plus the Packages-* file (all pet.specs, concatenated), and use the latest release URL as a PET repo.

PR #2170, opened 2021-03-23
petbuilds.sh: Some improvements

rizalmart:
* Added more paths to clean
* Added /usr/local/share path for DOC  and NLS
* Added gtk-doc and gnome help files on DOC path

### DIFF
--- a/woof-code/support/petbuilds.sh
+++ b/woof-code/support/petbuilds.sh
@@ -210,7 +210,7 @@ for i in ../rootfs-petbuilds/busybox ../rootfs-petbuilds/*; do
         rm -f ../petbuild-output/${NAME}-${HASH}/usr/share/icons/hicolor/icon-theme.cache
         rm -rf ../petbuild-output/${NAME}-${HASH}/lib/pkgconfig
         rm -rf ../petbuild-output/${NAME}-${HASH}/usr/lib/pkgconfig
-        rm -rf ../petbuild-output/${NAME}-${HASH}/usr/lib/girrepository-1.0
+        rm -rf ../petbuild-output/${NAME}-${HASH}/usr/lib/girepository-1.0
         rm -rf ../petbuild-output/${NAME}-${HASH}/usr/share/pkgconfig
         rm -rf ../petbuild-output/${NAME}-${HASH}/usr/share/gc
         rm -rf ../petbuild-output/${NAME}-${HASH}/usr/share/gir-1.0
--- a/woof-code/support/petbuilds.sh
+++ b/woof-code/support/petbuilds.sh
@@ -200,29 +200,82 @@ for i in ../rootfs-petbuilds/busybox ../rootfs-petbuilds/*; do
         fi
 
         rm -rf ../petbuild-output/${NAME}-${HASH}/root/.ccache
+        rm -rf ../petbuild-output/${NAME}-${HASH}/root/.cache/*
+        rm -rf ../petbuild-output/${NAME}-${HASH}/root/.dbus/*
+        rm -rf ../petbuild-output/${NAME}-${HASH}/root/.thumbnails/*
+        
         rm -rf ../petbuild-output/${NAME}-${HASH}/tmp
         rm -rf ../petbuild-output/${NAME}-${HASH}/etc/ssl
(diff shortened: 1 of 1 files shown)


woof-code/support/petbuilds.sh
```diff
@@ -275,15 +279,21 @@ for NAME in $PKGS; do
...
-        mv ../packages-${DISTRO_FILE_PREFIX}/${NAME}/usr/share/${DOCDIR} ../packages-${DISTRO_FILE_PREFIX}/${NAME}_DOC/usr/share/
+    for usrfld in usr/share usr/local/share; do
+    
+     if [ -d ../packages-${DISTRO_FILE_PREFIX}/${NAME}/${usrfld}/locale ]; then
+        mkdir -p ../packages-${DISTRO_FILE_PREFIX}/${NAME}_NLS/${usrfld}
+        mv ../packages-${DISTRO_FILE_PREFIX}/${NAME}/${usrfld}/locale ../packages-${DISTRO_FILE_PREFIX}/${NAME}_NLS/${usrfld}/
+     fi
+    
+     for DOCDIR in doc man info gtk-doc help gnome/help; do
+        if [ "${NAME}" != "cups" ]; then
```
dimkr: There's no cups petbuild in woof-code/rootfs-petbuilds (and shouldn't be, because CUPS is always a compat distro package), I'm confused.

woof-code/support/petbuilds.sh
```diff
@@ -200,25 +200,82 @@ for i in ../rootfs-petbuilds/busybox ../rootfs-petbuilds/*; do
...
-        rm -rf ../petbuild-output/${NAME}-${HASH}/usr/share/info
-        rm -f ../petbuild-output/${NAME}-${HASH}/usr/share/icons/hicolor/icon-theme.cache
         rm -rf ../petbuild-output/${NAME}-${HASH}/lib/pkgconfig
-        rm -rf ../petbuild-output/${NAME}-${HASH}/usr/lib/pkgconfig
-        rm -rf ../petbuild-output/${NAME}-${HASH}/usr/share/pkgconfig
-        rm -rf ../petbuild-output/${NAME}-${HASH}/usr/include
-
+        
+        for usrfld2 in usr usr/local
+        do
```
dimkr, line 215: Is there any specific file or directory that's added by any existing petbuild and needs to be deleted, or these changes are supposed to handle any existing and future leak of extra files? I mean, is this big change supposed to solve a concrete, existing problem, or protect against future problems?

rizalmart: Protect against future problems and improve flexibility. Also make petbuild output even lighter.

dimkr: I tried very hard to keep petbuilds.sh as small as possible, because the collection of 3builddistro petbuilds should remain small if we want to be able to maintain it (it's just Puppy-specific applications, traditional Puppy stuff like Geany with Puppy customization, and Puppy applications we want at a specific version, like JWM).

The whole idea behind #1984 is to make it possible to build a Puppy from other distro libraries, plus a traditional Puppy desktop with a basic set of applications, built from source. Building 'core' stuff like glibc, OpenSSL, CUPS or PulseAudio, or the machinery to support that, doesn't belong here. IMHO, it only adds complexity that may hide real problems, and why change something that achieves its goal?

I'm not obsessed with petbuilds.sh and trying to protect it against changes or something, only trying to keep it short and readable, at least while it's still new and buggy here and there.

rizalmart: However after compiling the program and installed with make install. Sometimes it leaves residual files which adds weight to the petbuild output. So additional cleanup to newly compiled petbuild output is needed.

dimkr: The question is, does petbuild.sh need to change so much. Is there any specific file or directory that's not cleaned up today that needs to be dealt with? If yes, petbuilds.sh should delete that, and only that. Otherwise, we should look at builtin_files, see what needs to be cleaned up, and deal with that.

rizalmart: However its hard to deal with builtin_files at this moment due to merging of packages which supposed not to be  happen

PR #2171, opened 2021-03-23
.xinitrc: Some improvements

rizalmart:
* Added libcanberra support
* Remove hard-coded path for commands

### DIFF
--- a/woof-code/rootfs-skeleton/root/.xinitrc
+++ b/woof-code/rootfs-skeleton/root/.xinitrc
@@ -12,7 +12,7 @@
 
 #v4.01 BK daemon to handle kernel uevents...
 echo '' > /var/local/pup_event_icon_change_flag
-/sbin/clean_desk_icons #v4.02 tidy up drive icons first.
+clean_desk_icons #v4.02 tidy up drive icons first.
 #/sbin/pup_event_frontend_d &
 #130525 ...start this further down.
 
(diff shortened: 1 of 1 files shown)


woof-code/rootfs-skeleton/root/.xinitrc
```diff
@@ -153,19 +159,19 @@ if [ "$CURRENTWM" = "xfwm4" ];then
...
- if [ -f /usr/bin/fbpanel ];then
+ if [ "$(which fbpanel)" != "" ];then
   #a bit of a hack: when 3builddistro runs fixmenus, which calls variconlinks,
   #which populates /var/local/icons with symlinks, /usr/local/lib/X11/pixmaps
   #is not yet populated (happens at first boot, from default icon theme)...
   [ ! -e /var/local/icons/home48.png ] && ln -fs /usr/local/lib/X11/pixmaps/* /var/local/icons/
   fbpanel &
  fi
- [ "$desktop" = "rox" ] && [ -f /usr/bin/lxpanel ] && lxpanel &
+ [ "$desktop" = "rox" ] && [ "$(which lxpanel)" != "" ] && lxpanel &
```
dimkr: btw, you can use `command -v lpanel` instead of `which` - it should be faster because it's a shell builtin and not a separate process.

rizalmart: Thanks for  the info. You're free to improve my pulled reuqest

dimkr:
@rizalmart I swear didn't try to convince you to close this PR

woof-code/rootfs-skeleton/root/.xinitrc
01micko: ..and you can check the exit status instead of test and comparison = faster. I ~prefer~ have been using `type` (redirecting the output to /dev/null) but it is not POSIX and output as far as I know is the same between `type` and `command -v`

PR #2173, opened 2021-03-23
.xinitrc: Some improvements

rizalmart:
* Added libcanberra support
* Remove hard-coded path for commands

### DIFF
--- a/woof-code/rootfs-skeleton/root/.xinitrc
+++ b/woof-code/rootfs-skeleton/root/.xinitrc
@@ -12,7 +12,7 @@
 
 #v4.01 BK daemon to handle kernel uevents...
 echo '' > /var/local/pup_event_icon_change_flag
-/sbin/clean_desk_icons #v4.02 tidy up drive icons first.
+clean_desk_icons #v4.02 tidy up drive icons first.
 #/sbin/pup_event_frontend_d &
 #130525 ...start this further down.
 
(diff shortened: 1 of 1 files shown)


01micko (changes requested):
Hi @rizalmart 

A few things.
- first the canberra block - that's fine, leave it
- calling things by path is faster than referring to $PATH variable. Remember, I maintain raspup and Pi 1 and Pi Zero take several minutes to boot and any time saving is highly desirable
- instead of `[<<test $(command -v executable)<<subshell != ""<<comparison ]<<close test` its faster to use `command - v executable >/dev/null 2>&1 <<portable redirection` which uses slightly less processes by just using the exit code. See #1715 

I'd like to @dimkr 's opinion on this.

woof-code/rootfs-skeleton/root/.xinitrc
```diff
@@ -41,14 +41,20 @@ fi
 
 xhost +local: 2>/dev/null
 
+if [ "$GTK_MODULES" != "" ] && [ "$(echo "$GTK_MODULES" | grep "canberra\-gtk\-module")" == "" ];then
```
dimkr, line 44: Is sound enabled by default, or ptheme needs to modified to enable sound (https://wiki.archlinux.org/index.php/Libcanberra#Configuration)?

dimkr:
I understand that Puppy needs to have a sound theme for this to work, and this PR doesn't touch DISTRO_PKGS_SPECS* of configurations in development (slacko, dpup).

01micko:
> I understand that Puppy needs to have a sound theme for this to work, and this PR doesn't touch DISTRO_PKGS_SPECS* of configurations in development (slacko, dpup).

Yes, but a puplet builder might want a sound theme, but it doesn't interest me.

woof-code/rootfs-skeleton/root/.xinitrc
rizalmart: Nope it is not enabled by default. But it depends on the Desktop Environment or user if they want sounds.

rizalmart:
> > I understand that Puppy needs to have a sound theme for this to work, and this PR doesn't touch DISTRO_PKGS_SPECS* of configurations in development (slacko, dpup).
> 
> Yes, but a puplet builder might want a sound theme, but it doesn't interest me.

Sounds is a part of Desktop Environment especially i  multitask activity for notifying something. Especially when running on a battery or deleting something. Major DE's supports sounds

01micko:
I don't really understand _why_ someone want's a major DE on Puppy? Sound scheme is ok, but I tried big stuff before on Puppy and it defeats the purpose of the OS. I baulked at diving into pulseaudio but people want bluetooth and that is the _only_ reason I did it.

IMO, small filers (pcmanfm and derivatives) small WM's (fluxbox, Icewm, windowmaker etc are cool) but we start bending things a little too much with xfce. Nothing wrong with it but maybe some of these huge things should be blacklisted in package management instead of accommodated. I think a line has to be drawn somewhere.

PR #2178, opened 2021-03-25
Add a basic Gitpod development environment 

dimkr:
This should help people like @zigbert, who modify individual files through GitHub's UI and push a commit for every modified file.


dimkr:
Gitpod can't build a Puppy because mounts are not allowed (0setup, 1download and 2createpackages work), but Gitpod can boot a Puppy built on GitHub inside QEMU.

Screenshots of Puppy development on a $100 Chromebook:

image

image

image

PR #2179, opened 2021-03-25
Remove unneeded dependencies from bullseye64

dimkr:


peabee:
Hi @dimkr 
Would be interesting to know how you determined which dependencies were "unneeded"?
Thanks

dimkr:
@peabee Something like this (not sure about the 5) - 
```
cat ~/.packages/woof-installed-packages | cut -d \| -f 5 | while read x; do [ -z "`fgrep +$x ~/.packages/woof-installed-packages`" ] && echo $x; done
```

PR #2182, opened 2021-03-27
"Improve" gtk3 display a tiny bit

peabee:
Just a tiny smidgen better in gtk3 - but still pretty awful..............


01micko:
Yeah, it's a pain but no matter how much gtkdialog you throw at it I still reckon it's a bug in the source to do with `pango` line wrapping. When I get my latest 'to-do' out of the way I'll dive into the gtkdialog sources.

dimkr:
> When I get my latest 'to-do' out of the way I'll dive into the gtkdialog sources.

Line wrapping has indeed changed in GTK+ 3, it's very likely this is a gtkdialog bug. I'll take a look too (after my TODOs) 😄

dimkr:
Took a quick look and I don't see GTK+ 2 vs. 3 difference inside gtkdialog itself that seem relevant 🤔

01micko:
ooh...
```
max-width-chars="1"
```
Try that in all `<text>` tags

eg
```
<text space-expand="false" space-fill="false" max-width-chars="1">
```

peabee:
max-width-chars="80"
works OK in gtk3 but not so good in gtk2...........
gtk3:
Screenshot-gtk3
gtk2:
Screenshot-gtk2

01micko:
I have a test for gtkdialog built with gtk3 in `/usr/lib/gtkdialog/box_splash`

I'm _not_ saying that it should be applied globally but would be good in  `welcome1stboot` and `quicksetup` and perhaps a few others in woof.

eg: (pseudo code)
```shell
# use $MAX when needed
MAX=''
if [ "$GTKDIALOG_GTK3" = true ];then
    MAX=' max-width-chars="1"' #preceding space
fi
```

Usage:
```
<text space-expand="false" space-fill="false"'${MAX}'>
```

Actually, with the correct name, it could be an environment variable set in `/etc/profile`

01micko:
Woops, closed by mistake.

The whole gtkdialog gtk3 test **should** be an environment variable.

dimkr:
It looks like every dialog a hbox with a label is horizontally maximized without wrapping. Is that right?

dimkr:
> Woops, closed by mistake.
> 
> The whole gtkdialog gtk3 test **should** be an environment variable.

Maybe we should patch gtkdialog to report what GTK version it uses, or, add a file under /etc/profile.d to the petbuild.

dimkr:
`space-expand="false" space-fill="false" max-width-chars="80"` looks OK here, in both GTK+ 2 and 3.

01micko:
With this one, on a local install I created this file in `/etc/profile.d/gtkdialog.sh`

```shell
if ldd $(which gtkdialog) | grep -iq 'gtk\-3' ; then #gtk3
	GTKDIALOG_BUILD=GTK3
elif ldd $(which gtkdialog) | grep -iq 'gtk\-X11\-2' ; then #gtk2
	GTKDIALOG_BUILD=GTK2
else
	GTKDIALOG_BUILD=GTK1
fi
export GTKDIALOG_BUILD

```

I added a test to `welcome1stboot`
```diff
--- /initrd/pup_ro2/usr/sbin/welcome1stboot	2020-07-09 08:32:28.000000000 +1000
+++ /usr/sbin/welcome1stboot	2021-04-01 15:12:06.035625694 +1000
@@ -4,6 +4,11 @@
 export TEXTDOMAIN=welcome1stboot
 export OUTPUT_CHARSET=UTF-8
 
+MAX_CHARS=''
+if [ "$GTKDIALOG_BUILD" = 'GTK3' ] ; then
+	MAX_CHARS=' max-width-chars="80"'
+fi
+
 read -r ISSUE ETC < /etc/issue
 W_MSG="$(gettext "Welcome,
 This is the first time you are running ${ISSUE}!...")"
@@ -32,7 +37,7 @@ export WELCOME_DIALOG='
     </hbox>
     <hbox border-width="7" space-expand="true" space-fill="true">
       <text space-expand="false" space-fill="false"><label>""</label></text>
-      <text xalign="0" use-markup="true" space-expand="true" space-fill="true"><label>"'"${LABEL_INTERNET}"'"</label></text>
+      <text xalign="0" use-markup="true" space-expand="true" space-fill="true"'$MAX_CHARS'><label>"'"${LABEL_INTERNET}"'"</label></text>
       <vbox space-expand="false" space-fill="false">
         <button>
           '$(/usr/lib/gtkdialog/xml_button-icon /usr/local/lib/X11/pixmaps/connect48.png huge)'
@@ -43,7 +48,7 @@ export WELCOME_DIALOG='
     <hseparator></hseparator>
     <hbox border-width="7" space-expand="true" space-fill="true">
       <text space-expand="false" space-fill="false"><label>""</label></text>
-      <text xalign="0" use-markup="true" space-expand="true" space-fill="true"><label>"'"${LABEL_SETUP}"'"</label></text>
+      <text xalign="0" use-markup="true" space-expand="true" space-fill="true"'$MAX_CHARS'><label>"'"${LABEL_SETUP}"'"</label></text>
       <vbox space-expand="false" space-fill="false">
         <button>
          '$(/usr/lib/gtkdialog/xml_button-icon /usr/local/lib/X11/pixmaps/configuration48.png huge)'
@@ -55,7 +60,7 @@ export WELCOME_DIALOG='
 
     <hbox border-width="7" space-expand="true" space-fill="true">
       <text space-expand="false" space-fill="false"><label>""</label></text>
-      <text xalign="0" use-markup="true" space-expand="true" space-fill="true"><label>"'"${LABEL_NEEDHELP}"'"</label></text>
+      <text xalign="0" use-markup="true" space-expand="true" space-fill="true"'$MAX_CHARS'><label>"'"${LABEL_NEEDHELP}"'"</label></text>
       <vbox space-expand="false" space-fill="false">
         <text space-expand="false" space-fill="false"><label>""</label></text>
         <button>

```

Similar works OK in `/usr/bin/pupdialog` but it needs more tweaks which gtk2 should be happy with.

Can't do a thing with `quicksetup`. Is what it is.

GTK3 welcome1stboot
welcome-gtk3

GTK2 welcome1stboot
welcgtk2


_The original without the patch is identical in gtk2_

dimkr:
IMHO, having readable dialogs is more important than consistency between GTK+ 2 and 3, and I hate those popups on first boot anyway.

01micko:
Love 'em or hate 'em they're there by popular demand! Consistency isn't really the end goal. A more professional look is more the aim.

If @peabee doesn't mind I might commit some changes to `peabee-patch-1` branch and do some gtk2 and gtk3 builds.

peabee:
I have no problem with that - but would mention that I haven't kept my fork up-to-date with testing as I don't use it for builds, only for occasional patches......
also just realised my laptop clock hasn't been adjusted for BST  :-()

01micko:
> but would mention that I haven't kept my fork up-to-date with testing as I don't use it for builds, only for occasional patches......

That's ok I just updated it

peabee:
> I hate those popups on first boot anyway.

I think a lot of people find them very useful - having all common settings presented in one dialog without having to search them scattered through the menus - why do you "hate" them? They are easy enough to quickly dismiss....

dimkr:
> why do you "hate" them? They are easy enough to quickly dismiss....

I prefer a more smooth experience: less questions, better default settings. For example, if we do the IP-to-location translation, we can choose the timezone/locale automatically. Also, I really hate the fact you have two popups and not one.

But if the wizards exist, we should make them look better.

peabee:
> if we do the IP-to-location translation, we can choose the timezone/locale automatically.

That would imply network connectivity on bootup which most systems (wifi) do not have.....??

dimkr:
> That would imply network connectivity on bootup which most systems (wifi) do not have.....??

This problem is easy to solve. But anyway, streamlining the first boot experience is a different discussion.

01micko:
Ok, not committed to `peabee-patch-1` yet but will be soon.

Here is the current state in picture form..

Below are snaps of `/usr/sbin/welcome1stboot`, `/usr/sbin/filemnt`, the gtkdialog environment variables for gtk3 and gtk2 respectively.

**Note**: `filemnt` was using `/usr/bin/pupdialog` but now uses `/usr/lib/gtkdialog/box_ok` by @zigbert 

### **GTK 3**

welcome-gtk3

filemnt-gtk3

gtkdialog-env-gtk3

### **GTK 2**

welcom-gtk2

filemnt-gtk2

gtkdialog-env-gtk2

**_I think the gtk3 versions look better!!!_**

dimkr:
They look much better!

01micko:
I think we're done here.

Issue #2194, opened 2021-03-31
Kernel-kit should run periodically instead of in every build


dimkr:
Building the latest 4.19.x, 5.4.x, etc' kernel once a week (or even two weeks) is enough.

This will make CI much much much faster, and pave the way for automated kernel updates in an already installed Puppy, independent of other components of Puppy.

(It's true that kernel-kit output is cached, but the cache gets wiped pretty frequently).

Also, a common place where every Puppy builder can get up-to-date kernels with matching firmware, would protect against regressions like #2060, where a driver or a firmware blob that got accidently removed in Puppy version x+1.

dimkr:
Not sure how this issue interacts with #2168, but I didn't have to fix it to make #2195 work in my fork 🤔

Issue #2060, opened 2021-02-27
WIFI card Intel Advanced-N 6200 AGN not detected


z3huti:
BionicPup64 8.0 does not detect WIFI card Intel Advanced-N 6200 AGN

but BionicPup 32bit does

also Slacko Puppy 7.0 64bit does not detect it, but 32bit works

dimkr:
Hi @z3huti, can you share the output of `cat /proc/version` and `dmesg` from at least one scenario where it works and another where it doesn't?

z3huti:
> Hi @z3huti, can you share the output of `cat /proc/version` and `dmesg` from at least one scenario where it works and another where it doesn't?

https://drive.google.com/file/d/1q0D2HGWJIcn-jAb86ubnRD4oBUIBoxag/view?usp=sharing

dimkr:
The difference is here:

```
iwlwifi 0000:03:00.0: Direct firmware load for iwlwifi-6000-6.ucode failed with error -2^M
iwlwifi 0000:03:00.0: Direct firmware load for iwlwifi-6000-5.ucode failed with error -2^M
iwlwifi 0000:03:00.0: Direct firmware load for iwlwifi-6000-4.ucode failed with error -2^M
iwlwifi 0000:03:00.0: no suitable firmware found!^M
iwlwifi 0000:03:00.0: minimum version required: iwlwifi-6000-4^M
iwlwifi 0000:03:00.0: maximum version supported: iwlwifi-6000-6^M
```

```
iwlwifi 0000:03:00.0: Direct firmware load for iwlwifi-6000-6.ucode failed with error -2^M
iwlwifi 0000:03:00.0: Direct firmware load for iwlwifi-6000-5.ucode failed with error -2^M
iwlwifi 0000:03:00.0: loaded firmware version 9.221.4.1 build 25532 op_mode iwldvm^M
iwlwifi 0000:03:00.0: CONFIG_IWLWIFI_DEBUG disabled^M
iwlwifi 0000:03:00.0: CONFIG_IWLWIFI_DEBUGFS disabled^M
iwlwifi 0000:03:00.0: CONFIG_IWLWIFI_DEVICE_TRACING disabled^M
iwlwifi 0000:03:00.0: Detected Intel(R) Centrino(R) Advanced-N 6200 AGN, REV=0x74^M
```

[...]

PR #2195, opened 2021-04-01
Build longterm kernels once a week and disable PR builds (#2194)

dimkr:
This will reduce build times and make woof-CE's CI more scalable (need to make room for the next Ubuntu-based Puppy).

This PR will never pass CI because now Puppy builds try to pull kernel-kit output that's built once a week on the testing branch, so this PR must be merged before I can debug any issues. (Tested in a fork).

Because now slacko/fossa/dpup all use the same kernel packages (so fdrv and zdrv are renamed before inclusion in ISO), I added the package name suffix to /proc/version.

### DIFF
--- a/.github/workflows/bullseye.yml
+++ b/.github/workflows/bullseye.yml
@@ -7,37 +7,11 @@ on:
       - testing
     paths:
       - 'woof-distro/x86/debian/bullseye/**'
-      - 'kernel-kit/**'
-      - 'woof-code/**'
-      - '!woof-code/support/arm_**'
-      - 'woof-arch/x86/**'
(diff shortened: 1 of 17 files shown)


dimkr:
The failing builds demonstrate what the "fail early, fail fast" engineering principle means 😺 

CI is so much faster now, it's 15 mins or less to build an x86/x86_64 Puppy on GitHub with good compression, petbuilds and all. The build should fail if there's no fdrv, so we're protected against stuff like #2060, and non-release builds are even faster (can easily take <=5 mins) because they still use LZO, and petbuilds are cached.

PR #2201, opened 2021-04-03
Fix empty woof-installed-packages if adrv is enabled

dimkr:
There's a big pile of bugs in this loop, hope I fixed all of them.


dimkr:
~TODO: firefox-esr is skipped because we look for `firefox`, it's `yes|firefox|firefox-esr` and not `yes|firefox-esr|firefox-esr`~ fixed

PR #2202, opened 2021-04-03
Add a dmz-cursor-theme petbuild

dimkr:
The `aero` cursor theme is not present in a vanilla woof-CE build, and the default X cursor is horrible - it's ugly, small and unfamiliar.

Until now, only dpup had dmz-cursor-theme (the neutral cursor theme from the GNOME 2 days) and it's the default cursor for Dark Gray, while all other themes either use the default X cursor, or fall back to it.

This PR adds a petbuild, so even Slackware-based puppies have this theme, and makes it the default.


01micko:
I like the idea.. are all breakages fixed?

dimkr:
@01micko as far as I see, yes - tested on dpup and slacko

01micko:
ok, merging

PR #2203, opened 2021-04-03
pup-advert-blocker hangs forever when starting and stopping

dimkr:


01micko:
Why just dpup? I think it should be global.

dimkr:
> Why just dpup? I think it should be global.

Me too, but I think the only reason why this change doesn't meet opposition is that some Puppy traditionalists don't know about it.

01micko:
Ok, I'll merge but I might make it a config setting later.

PR #2204, opened 2021-04-03
Put the working directory in the shell prompt, but only for dpup 

dimkr:

### DIFF
--- a/woof-distro/x86/debian/bullseye/_00build.conf
+++ b/woof-distro/x86/debian/bullseye/_00build.conf
@@ -124,4 +124,5 @@ chroot . /usr/sbin/setup-spot transmission-gtk=true
 mv -f ../adrv/usr/bin/transmission-gtk{,.bin}
 mv -f usr/bin/transmission-gtk ../adrv/usr/bin/
 rm -f usr/bin/transmission-gtk.bin
+echo 'PS1=\"\\\\w\\\\$ \"' > etc/profile.d/ps1.sh
 "
--- a/woof-code/rootfs-skeleton/etc/profile
+++ b/woof-code/rootfs-skeleton/etc/profile
@@ -26,7 +26,7 @@ else
 fi
 
 USER=`id -un`
-PS1="\w\$ "
+PS1="# "
 LOGNAME=$USER
 HISTSIZE=1000
 HISTFILE="$HOME/.history"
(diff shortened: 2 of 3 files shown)


peabee:
Thanks

01micko:
This should be a config setting too.

eg:

```
#Prompt: leave empty to keep default or set the preferred prompt here for the CLI
#use whatever prompt you like
PROMPT="\w\$\"

```

01micko:
Oooh, conflict since last PR

woof-distro/x86/debian/bullseye/_00build.conf
```diff
@@ -124,4 +124,5 @@ chroot . /usr/sbin/setup-spot transmission-gtk=true
 mv -f ../adrv/usr/bin/transmission-gtk{,.bin}
 mv -f usr/bin/transmission-gtk ../adrv/usr/bin/
 rm -f usr/bin/transmission-gtk.bin
+echo 'PS1=\"\\\\w\\\\$ \"' > etc/profile.d/ps1.sh
```
01micko, line 127: No need to escape because you are echoing literal with single quotes.
```
 echo "PS1=\"\\w\\\$ \""
PS1="\w\$ "
```

with double quotes: escape`"`escape`\w`escape`\`escape`$ `escape`"` or are you using a different shell?

01micko:
making this a config setting

01micko:
it's hard tp please everyone but there shouldn't be any complaints about e7b7f79 , the fix for f3d2b51

woof-distro/x86/debian/bullseye/_00build.conf
dimkr: The huge amount of escaping here is needed because we read line by line and `eval` each line in 3builddistro.

01micko: Ah I see, so I hope you like my 'peaceful' resolution.

dimkr: Definitely! Much cleaner. I thought you were in the 'old prompt camp' too 😊

01micko: On the fence :person_fencing: 
:rofl:

PR #2209, opened 2021-04-05
xorgwizard: Added video mode detection

rizalmart:
Instead of  predefined screen resolution. It just read the supported screen resolutions of a monitor. To enable this feature just create xorg-detect-screen-size file in /var/local before running the xorgwizard.


01micko:
Looks ok and with weird tablets and stuff might be useful.

PR #2210, opened 2021-04-05
When building a PR, keep the output for 3 days

dimkr:
@01micko This will make it easier to test powerapplet_tray, and make you less hesitant to publish alpha builds.

Now, every build puts the woof-CE output in a build artifact and it's deleted after 3 days:

image

(It's a zip file containing the files that previously, went into the release, and it's uploaded even when it's a PR build, so the PR can be tested easily without having to build locally first)
(All CI workflows are exactly the same now, except the kernel package name!)

Because now publishing a release is not fun, in addition, I'm working on this (will be a separate PR I think, I'm still testing it):

image

This job does the same, but without caching, and creates a release **draft** at the end (so it can be tested before it's published, or discarded if it's not good).

### DIFF
--- a/.github/workflows/bullseye.yml
+++ b/.github/workflows/bullseye.yml
@@ -1,13 +1,6 @@
 name: bullseye
 
 on:
-  push:
-    branches:
-      - ci
-      - testing
-    paths:
-      - 'woof-distro/x86/debian/bullseye/**'
-      - 'woof-distro/x86_64/debian/bullseye64/**' # bullseye/* are symlinks to bullseye64/8
   pull_request:
     branches:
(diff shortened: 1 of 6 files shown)


peabee:
Only problem is the download size if you only want the .iso for testing.......maybe *3 - 1GB vs 400MB.....

dimkr:
That's true @peabee, but I had to choose between uploading `*.iso|*.sfs` (in that case, GitHub Actions puts all files in a .zip file) or uploading the files individually, but then I'd have to specify their exact names.

The download is pretty fast with `aria -j4 -x4 -s4 `.

PR #2212, opened 2021-04-05
Add a parameterized workflow that release a draft

dimkr:
Tested with a 64-bit slacko64 8.0 and a 32-bit 7.0.

~You can't really choose the kernel version, it must be the one in _00build.conf, but this should be easy to fix.~ fixed and it's pretty nice, you can choose the 4.19.x kernel to build a "retro" flavor, or upgrade some old Puppy you like to 5.10.x to increase its shelf life.


dimkr:
@01micko I triggered https://github.com/puppylinux-woof-CE/woof-CE/actions/runs/721386335, just for testing purposes. I'll discard the release.

01micko:
I'll give it a run when it's ready.

PR #2225, opened 2021-04-10
Add nlsx to the dpup ISO

dimkr:


dimkr:
Not very useful until #2227 is fixed

PR #2228, opened 2021-04-10
Remove pmirror if rsync is missing

dimkr:


zigbert:
pMirror is depending on mirdir - not rsync

dimkr:
> pMirror is depending on mirdir - not rsync

https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-code/rootfs-skeleton/usr/sbin/pmirror#L11
https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-code/rootfs-skeleton/usr/sbin/pmirror#L140

@zigbert Please correct me if I'm wrong

zigbert:
Whot!? - This is not the pMirror I made...
History log explains it... rewritten by wdlkmpx in 2019.
Maybe it should have another name - it is actually another app... cloned gui - different backend
But... who cares...

CollaboratorGCM:
In addition, isn't "rsync" a part of the standard PUP command set? Why would it be missing. That's almost like "cp" or a similar command is missing???

Merely a comment.

dimkr:
> In addition, isn't "rsync" a part of the standard PUP command set?

I don't know what "PUP" means, but I rarely use rsync and prefer to be able to leave it out.

PR #2229, opened 2021-04-11
Add "ndrv" support for automatic loading of nlsx

dimkr:
The main question is, did I miss anything? I grepped the entire woof-code for `adrv`, `ADRV`, etc' and adjusted all places I found.


initrd-progs/0initrd/init
```diff
@@ -969,6 +977,10 @@ find_onepupdrv "$A_PART" "$A_BP_FN" "$A_DEF_FN" "a"
 [ "$ONE_FN" ] && ADRV="$ONE_PART,$ONE_FS,$ONE_FN"
 [ "$ADRV" ] && { LOADMSG="adrv"; setup_onepupdrv "$ADRV" "a" "p"; }
 
+find_onepupdrv "$N_PART" "$N_BP_FN" "$N_DEF_FN" "a"
+[ "$ONE_FN" ] && NDRV="$ONE_PART,$ONE_FS,$ONE_FN"
+[ "$NDRV" ] && { LOADMSG="ndrv"; setup_onepupdrv "$NDRV" "n" "p"; }
```
01micko, line 982: I was thinking whether the third arg should have been added or not before this PR but at the end of the day it doesn't matter because NLS files won't cover anything anyway.

gyrog: Why does it need it's own ".sfs" file? (Why not include it in puppy....sfs?)
Why does it need to be "above" the 'puppy...sfs'?
Could it be appended just like an extra-sfs? (It could be done fairly easily in 'rc.sysinit'.)
If it must be a Puppy system '.sfs', where is the definition of the filename?
Is there any code in woof-ce to write the extended 'DISTRO_SPECS' in 'initrd.gz' and '/etc/DISTRO_SPECS' in 'puppy....sfs'?

Will the code work?
Probably, but we've always used the filename definitions in DISTRO_SPECS in the past.

Why is this a problem now?
Do we need to be creating an ndrv?
I don't understand why we would, but then I don't know what it is, and under what situation it is needed.
There was a suggestion that it needs to be loaded automatically because the user would not have enough English to navigate to load it as an extra-sfs. If that is so, then isn't their lack of English going to cause lots more significant problems?
If this is supposed to be for a translator, then I would have thought that they would have to know 2 languages, 1 in which Puppy is already available, and 1 in which it is not. So why would they not be able to mount this as an extra sfs in the Puppy running in the already translated language that they know?

dimkr: Traditionally, locales are left out in Puppy, so it's English-only. I want to add support for building a Puppy with locales, but make them optional because some users don't need them, hence the SFS. But it needs to be loaded by default, because that's what makes non-English speaking users able to change the language.

gyrog:
@dimkr,
Then why "n", should it be an "l" (locale) drive?

I don't see any advantage in adding this "ndrv" above the Puppy...sfs, apart from some folk who would simply see it as "now there are 3 sfs's available above the puppy...sfs for us to fiddle with".
But there's no problem if it is.

Slightly off-topic:
Should English be a locale? Should Puppy by default be released with an English "ndrv".

As an ordinary optional system .sfs, this is not really available until after the initial build of the stack,
so there's a significant amount of 'init' happening with this "locale file" unavailable.
Do we need a special solution that makes it availablle sooner in 'init'?
Perhaps one that adds a "locale file" to the 'initrd.gz'?

initrd-progs/0initrd/init
gyrog: @dimkr,
Is there woof-ce code to produce DISTRO_SPECS files with the "ndrv" information?

dimkr: Yes, I changed the relevant parts in 3builddistro (please correct me if I'm wrong).

dimkr:
> @dimkr,
> Then why "n", should it be an "l" (locale) drive?

I can replace it with "ldrv" if it really matters.

> I don't see any advantage in adding this "ndrv" above the Puppy...sfs, apart from some folk who would simply see it as "now there are 3 sfs's available above the puppy...sfs for us to fiddle with".

I'm not sure I understand. Are you against the idea of supporting applications in languages other than English (this feature is optional, off by default), or prefer this aufs layer to be in a different position (can be changed easily)?

01micko:
> > @dimkr,
> > Then why "n", should it be an "l" (locale) drive?
> 
> I can replace it with "ldrv" if it really matters.
> 

Whatever.  'l' would be my preference, but that isn't important.

> > I don't see any advantage in adding this "ndrv" above the Puppy...sfs, apart from some folk who would simply see it as "now there are 3 sfs's available above the puppy...sfs for us to fiddle with".
> 
> I'm not sure I understand. Are you against the idea of supporting applications in languages other than English (this feature is optional, off by default), or prefer this aufs layer to be in a different position (can be changed easily)?

@gyrog , do you mean  that the `ndrv` might be open for abuse by being higher on the stack than the puppy sfs. I agree actually if so, and it really should be below the main sfs. Makes no difference except that it isn't as attractive for casual puppy hackers to add an 'ndrv' for their own purposes rather than for translations.

Correct me if I'm wrong @gyrog but if there is a 3rd arg to `setup_onepupdrv()` (apparently 'p')  it puts the sfs above the main sfs in the aufs stack?

gyrog:
@dimkr,
The drive letter is not important,  and now I get it, the "n" comes from NLS.

But we should be clear about the position of the ndrv...sfs branch within the stack.
Should it be at the top, the bottom or somewhere in the middle.
It could work just as well at the bottom, as @01micko observed "it doesn't matter because NLS files won't cover anything anyway".
Your doco and most of the code suggests that it resides in between the adrv...sfs and the ydrv...sfs,
I suggest that it should reside between the ydrv...sfs and the puppy...sfs,
we view it as part of the Puppy core, (along with puppy...sfs, fdrv...sfs, and zdrv...sfs),
so it should be able to be patched by the 'ydrv...sfs'.

To do this, would you please move the blob of code starting with "find_onepupdrv", to above the corresponding 'ydrv...sfs' code.
Your current code places the 'ndrv...sfs' at the top of the stack, between the RW branch and the 'adrv...sfs'.
The "p" parameter at the end of the "setup_onepupdrv" command keeps pushing the branches into the stack at this position. So the 'adrv...sfs'  and the 'ydrv...sfs'  (and the 'ndrv...sfs') get added to the stack in the reverse order.

gyrog:
@01micko,
(I was composing while you commented).
I wouldn't use the word "abuse", since the 'init' script doesn't care what's actually in each sfs file, just that there must be a 'puppy...sfs' file.
Some folk have made very eclectic use of this, by renaming the 'puppy...sfs' as 'zdrv...sfs' to give themselves the maximum number of defined branches above the main puppy stuff.

dimkr:
Closing until I fix this. Thanks @gyrog and @01micko.

01micko:
@dimkr there is a compelling reason to put `nlsx` on top of the stack to do with the `shared-mime-info_FIXUPHACK` template introduced in 4ed4f77 only 3 weeks after this PR was closed.

PR #2232, opened 2021-04-13
Enable support for ext4 encryption in kernel 4.19.x

dimkr:
Needed by https://github.com/dimkr/frugalify/pull/4: it's an easy way to encrypt the save folder; good to have in case a Puppy flash drive gets lost.


dimkr:
@01micko This is pretty cool! https://github.com/dimkr/frugalify/pull/4 adds support for encrypted save folders. If you take the latest dpup OS from https://github.com/dimkr/woof-CE/releases, replace the "init" executable with the latest aufs+fscrypt variant of frugalify from https://github.com/dimkr/frugalify/releases and boot for the first time, /save is encrypted, and frugalify is only 8K bigger 😸 

I wonder if we can use the same technique to encrypt the Puppy save file itself, while the save file is attached to a loopback device and contains an unencrypted ext4.

PR #2245, opened 2021-04-24
network-wizard 2.2: Display corrected progress windows, fixes, etc.

rserwin1:
Corrects failure to connect due to failure of wpa_supplicant progress dialog and pauses between progress updates so they can be seen.
Restores display of progress windows during reconnections.
Adds interface name to dhcpcd progress dialog, because all interfaces get reconnected.
During reconnection and at boot-up, when multiple interfaces are present, the last-configured interface is tried first, providing control and predictability of the selected interface.
Various fixes: Corrects pcmcia check; simplifies link detection; sets IS_WIRELESS for use in a particular message; removes v411 BK hack to remove old network wizard configs (*[0-9]mode); corrects iwconfig check to test for an associated AP address; prevents multiple psk= lines in wpa profile; corrects a PID test; corrects the wait for initialized interfaces, to retry /sys/class/net/*/address; corrects test for X active.


rserwin1:
Also corrects failure to start a connection due to a progress dialog input timing issue, resolved by delaying the first output to the dialog.

PR #2246, opened 2021-04-27
0setup: Remove slackware base packages as deps

rizalmart:


peabee:
@rizalmart @01micko 
Seems dangerous to plumb in specific Slackware package names into 0setup.....
For instance aaa_glibc-solibs was only name changed recently by Slackware......
Does this have to be inline in 0setup?

rizalmart:
@peabee 
> @rizalmart @01micko
> Seems dangerous to plumb in specific Slackware package names into 0setup.....
> For instance aaa_glibc-solibs was only name changed recently by Slackware......
> Does this have to be inline in 0setup?

packages started in aaa_ prefix are slackware base files. It conflicts with glibc, util-linux, coreutils and many other packages.

01micko:
Except for aaa_elflibs none of these are listed as dependencies for anything.
I can't see any problem so far and I have run slacko since based on slackware 13.0.
Closing.

PR #2250, opened 2021-05-01
rc.shutdown: Unmount timeshift mountpoints

rizalmart:
Timeshift is a system restore utility for linux. Sometimes it mounts on /run/timeshift/. For safety reason, it must be unmounted upon shutdown

More info
[https://github.com/teejee2008/timeshift](https://github.com/teejee2008/timeshift)

### DIFF
--- a/woof-code/rootfs-skeleton/etc/rc.d/rc.shutdown
+++ b/woof-code/rootfs-skeleton/etc/rc.d/rc.shutdown
@@ -111,6 +111,13 @@ sync
 umount-FULL -a -t cifs,smbfs,nfs,sshfs
 ###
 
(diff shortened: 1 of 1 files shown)


01micko:
Can't a package manage that?

gyrog:
I would have thought a little scipt in '/etc/init.d' could do that, when it's called with 'stop' parameter.

CollaboratorGCM:
Timeshift, rsnapshot, BackInTime and TimeVault - ALL would have this/similar problem especially when Save-session functions are present in the running system. Correct?

1. So is this a Puppy problem to solve via package (PET) builds for those, as well? 
2. And is it a WoofCE need or a forum builder's need?

PR #2260, opened 2021-05-15
Sync focal64 with bullseye

dimkr:
Haven't tested these changes yet.

- [x] Does sound work in the latest Firefox + ~~Chrome~~? Firefox seems happy, I think this is good enough
- [x] Does pa-applet work?
- [x] ~~Is the rendering of websites and fonts Ubuntu-like, or Puppy-like?~~ Ubuntu-like
- [x] Does nscd run? Is DNS cache working?
- [x] Does bash completion work? I.e. `gunzip <tab>` should offer only *.gz
- [x] Is the font rendering issue (squares instead of letters, at least in the terminal) fixed?
- [x] Add pavucontrol


dimkr:
@01micko Anything else I should add here, from slacko or dpup?

- [ ] Upgrade to the same 5.10.x kernel as slacko and dpup, or upgrade to the latest 5.4.x (we have automated builds of both)?

mrfricks:
hi @dimkr 
see this about fixing the font rendering issues https://github.com/puppylinux-woof-CE/woof-CE/issues/2263#issue-892495843

dimkr:
@mrfricks I'll add fixing the messed up rendering to the TODO at the top

mrfricks:
stopping that old fonts.conf overwriting the fontconfig one should also fix website rendering

01micko:
A feature of all @mrfricks pups is `quickpet`. We should add icons and templates to `ptheme` to accommodate that.

mrfricks:
is pavucontrol needed for pa-applet?

dimkr:
> is pavucontrol needed for pa-applet?

It's as essential as alsamixer, when you have pulseaudio. I'll add it.

01micko (approved):

dimkr:
Building this branch in https://github.com/dimkr/woof-CE/actions/runs/849268452 (it's mixed with other branches, but nothing that affects focal64)

dimkr:
#2263 is resolved and looks like Fossapup64 is now configured to use PulseAudio properly.

dimkr:
@mrfricks I see the quickpet package installs some icons, which probably conflict with the PMaterial directory from woof-CE. Other than re-packaging quickpet without the 9.5 updates thing and without the conflicting icons, I think it's not much work to stabilize this "9.6" and release it.

IMHO this PR can be merged even before quickpet is fixed.

mrfricks:
there are still some issues that need resolving.  i need to break apart the zz_fossafix64.pet into individual packages as it contains lots of fixes for various items.  e.g. custom hacks-postinstall.sh, fixes for older apps,  missing libs, all default settings, all app config settings.
osmo also needs repacking as the one released with fossa crashed and was fixed with updates

the build also seems to grab gtk2 versions of geany & mtpaint where as fossapup has gtk3 versions

/root/.config/gtk-3.0/settings.ini isn't populated so gtk3 apps look bad (missing icons etc)

libllvm may have upped a version as well

dimkr:
> the build also seems to grab gtk2 versions of geany & mtpaint where as fossapup has gtk3 versions

Done

Issue #2252, opened 2021-05-06
Support for running JWM over Cage+Xwayland


dimkr:
https://source.mnt.re/reform/reform-system-image/-/blob/mmdebstrap/reform2-imx8mq/template-skel/bin/reform-windowmaker

It's a very neat trick!

I've built a dpup bullseye with wlroots and all the required Wayland bits, then built Cage and used it to run a fullscreen Geany or ROX-Filer.

I couldn't run a non-rootless Xwayland, though, because there's no elogind and Cage doesn't like running as root. I'm trying to add PAM, PAM support in busybox login and elogind, to see if now it's possible to run the normal Puppy desktop inside a full screen Xwayland.

In addition to the cool factor, JWM over Xwayland would provide tear-free rendering, and allow us to remove many packages needed by Xorg (i.e. stuff like xserver-xorg-video-*). In some ways, despite of this "translation layer", JWM over Wayland could be more elegant than a proper Xorg desktop, and I wouldn't be surprised if it's smaller in terms of ISO size.

dimkr:
Got the Puppy desktop running over Xwayland and cage. Had to build Xwayland, seatd, wlroots (with a patch that allows it to build against Wayland 1.18.0) and cage (with a patch that allows it to run as root).

The keyboard wizard doesn't do anything because Xwayland doesn't read Xorg.conf.

Applications seem to work just fine and everything is rendered without tearing, on an old x86_64 laptop. Haven't tried 3D games yet, and haven't tested this setup on ARM.

dimkr:
Look ma, no Xorg!

![xwayland](https://user-images.githubusercontent.com/1471149/118363444-0e56e200-b59d-11eb-8bb5-de34bf82b8ee.png)

CollaboratorGCM:
+1!

CollaboratorGCM:
I have 3 configurations for testing an ISO when one is available that will interest you in your wayland progress.
1. Touchscreen AIO AMD 2018 desk
2. TV-HDMI-nVidia Intel 64bit desktop
3. 2008 HP touchscreen Intel 64bit Laptop

dimkr:
Published https://github.com/dimkr/woof-CE/releases/tag/cagepup64-8.0-prepreprealpha1, for early testing.

The keybouard/mouse wizard should work, but there's no way to configure the resolution and I'm sure there are more issues.

[...]

PR #2265, opened 2021-05-16
Add support for running JWM under Xwayland using Cage #2252

dimkr:
A Wayland compositor is responsible for rendering, device management (displays, input) and window management. A Wayland compositor combines some of the responsibilities of the X server and a window manager.

For example, Cage is a Wayland compositor that runs one fullscreen application ("kiosk"), and its main use case is stuff like POS systems.

To allow X applications to work on Wayland, most Wayland compositors (including Cage) support running Xwayland, a Wayland application that maps every X window to a Wayland window. X doesn't have direct control over input devices: instead, X sees a "virtual" mouse, etc', and the compositor is responsible for input device settings. Another difference compared to regular X.Org is that Xwayland doesn't run a traditional X window manager and there's no X desktop window.

Traditionally, Puppy uses JWM (X window manager) and ROX-Filer, which draws the pinboard on the X root window. They can't be used with Wayland, because they need an X server. Unlike other applications, can't run in a Xwayland window. Also, unlike some X window managers, which have a Wayland equivalent (i.e. i3 and sway), we don't have a Wayland equivalent of JWM, and ROX-Filer can't run on Wayland because it still uses GTK+ 2.

By running Cage and a fullscreen, "rootful" Xwayland that talks to a X window manager and provides a root window, we can run JWM and ROX-Filer. Basically, it's "X in a window", but it's a fullscreen window.

This PR adds:
- A `cage` package that contains Xwayland and Cage, built from source
- (Cage is built against wlroots and seatd's "builtin" backend, allowing Cage to run without systemd's logind, and without an additional seatd daemon; also, by avoiding the logind/elogind dependency, we don't have to switch from busybox login to shadow login, with proper PAM and XDG environment variables/directories support and D-Bus user sessions)
- `startxwayland`, a script that runs Xwayland under Cage, then passes control to xinitrc
- Support for running `startxwayland` if `/usr/bin/Xorg` is missing
- A `cage` patch (https://github.com/Hjdskes/cage/pull/188) that adds support for touchpad and mouse configuration via environment variables
- Support for specifying these environment variable values in ~/.Xwaylandrc, using Puppy's traditional keyboard/mouse wizard
- `swayidle`, to run xlock (Puppy's traditional screensaver) when idle

This is not proper Wayland migration, but:
- It's a good opportunity to iron out issues like RAM consumption, dedicated GPU support, etc', without changing the way the Puppy desktop looks or behaves, or breaking applications people use today
- It's a good opportunity to provide features like tear-free rendering, but without having to use something like picom (and only for some GPUs where it doesn't produce visual artifacts); the overhead of Xwayland should be very small, and the performance gains compensate for this overhead
- ROX-Filer is the only remaining must-have application that still uses GTK+ 2 and can't run on Wayland directly (everything else uses GTK+ 3, which supports Wayland natively); porting it to GTK+ 3 won't be trivial

---

Motivation:

1. X.Org is ~unmaintained
2. Tear-free rendering and better utilization of hardware acceleration
3. Reduction in size: XWayland is smaller than X.Org
4. Preparation for proper migration to Wayland, possibly with a (future) Wayland compositor that resembles JWM
5. Better support for ARM Chromebooks that work great with Wayland, but suffer from slow rendering under X with the same open-source GPU drivers

---

TODO:
- [x] Test xkbconfigmanager
- [x] Enable touchpad tap by default
- [x] Make mouse/touchpad settings configurable
- [x] Restart X when changing keyboard layout, etc'
- [x] Test with multiple monitors
- [x] Organize Cage patches
- [ ] ~~Upstream Cage patches~~ Will take time and shouldn't block this PR
- [x] Make sure all CI builds pass
- [x] ~~Test against NVIDIA/AMD/multi-GPU setups~~ Will take time and shouldn't block this PR, NVIDIA support for accelerated Xwayland is still rather new
- [x] Check what packages other than Xorg its video/input drivers can be removed
- [x] ~~Check if Debian's Xwayland package is good enough~~ If we already build Cage, swayidle, etc' and need up-to-date versions, we should also build Xwayland
- [x] Compare RAM consumption to good old Xorg, on multiple machines
- [x] See what parts of the cage package can be removed (the seatd executable?)
- [x] Check what functionality other than the keyboard/mouse wizards depends on having xorg.conf
- [x] Fix restartwm, X exits but doesn't start again
- [x] Make the screensaver work
- [x] Fix DPI inconsistency bewteen X, GTK+ 2 and GTK+ 3 applications - the latter seems to ignore Xft.dpi
     
Testing results:
- X needs to be restarted to detect a newly connected monitor
- Flickering and junk when running Tux Racer while connected to a second monitor

118363444-0e56e200-b59d-11eb-8bb5-de34bf82b8ee


01micko:
This will likely work in Slacko too.

dimkr:
> This will likely work in Slacko too.

Right now, the "activation switch" for building Xwayland+seatd+wlroots+cage is having no Xorg.

Once I get https://github.com/Hjdskes/cage/issues/138 fixed and make Puppy's keyboard/mouse to work well, we can start on testing on Slacko and maybe Fossa too.

dimkr:
Keyboard and pointer device settings are now configurable. It works great!

input

EDIT: now the dialog allows the user to change more settings, like button emulation, disable-while-typing, etc'

dimkr:
~~I'm still unable to get X to run in a VM~~, but all the basics are in place: keyboard and mouse configuration, screensaver support, ability to change DPI, support for applications running as spot, support for Firefox and Chrome out of the box, and full bypass of xorgwizard and the creation of /etc/X11/xorg.conf.

Everything else can happen in separate PRs.

EDIT: fixed

01micko:
Do you want to fix #2260 before I merge this?

dimkr:
> Do you want to fix #2260 before I merge this?

Fixed the conflict there. The order in which these PRs get merged doesn't matter to me - do what doesn't interfere with your Slacko work. Both should not affect Slacko at all, but this PR touches more files.

01micko:
> > Do you want to fix #2260 before I merge this?
> 
> Fixed the conflict there.

merged

> The order in which these PRs get merged doesn't matter to me - do what doesn't interfere with your Slacko work. Both should not affect Slacko at all, but this PR touches more files.

I'll take a closer look tomorrow (my time - 10 - 12 hours or so)

01micko (approved):

Issue #2268, opened 2021-05-17
LXTask doesn't show processes running as spot


dimkr:
It needs to be patched to enable that option by default!

PR #2269, opened 2021-05-17
Patch LXTask to show spot processes by default (#2268)

dimkr:


woof-code/rootfs-petbuilds/lxtask/spot.patch
```diff
@@ -0,0 +1,12 @@
+diff -rupN lxtask-0.1.10-orig/src/functions.c lxtask-0.1.10/src/functions.c
+--- lxtask-0.1.10-orig/src/functions.c	2021-05-17 21:25:16.446720499 +0300
++++ lxtask-0.1.10/src/functions.c	2021-05-17 21:27:03.713389812 +0300
+@@ -231,7 +231,7 @@ void load_config(void)
+ 
+     show_user_tasks = key_file_get_bool(rc_file, group, "show_user_tasks", TRUE);
+     show_root_tasks = key_file_get_bool(rc_file, group, "show_root_tasks", FALSE);
+-    show_other_tasks = key_file_get_bool(rc_file, group, "show_other_tasks", FALSE);
++    show_other_tasks = key_file_get_bool(rc_file, group, "show_other_tasks", TRUE);
```
01micko, line 9: You would think that would be a config option.

01micko (commented):
I'll look more tonight.

woof-code/rootfs-petbuilds/lxtask/spot.patch
dimkr: It is, but the default doesn't make sense for Puppy: it shows processes running as root, but not those running as other users (exactly the opposite of what you want in other distros).

Issue #2267, opened 2021-05-17
.desktop files are opened with Geany when xdg-utils is installed


dimkr:
xdg-open from rootfs-skeleton is replaced with the real deal, which opens .desktop files with Geany instead of running what's in Exec through `rox`.

When I install Chromium in dpup, it pulls this package.

dimkr:
http://forum.puppylinux.com/viewtopic.php?f=146&t=2971

01micko:
Ok, approving.

PR #2270, opened 2021-05-17
Make sure rootfs-skeleton xdg-open is not replaced with the xdg-utils one (#2267)

dimkr:


woof-distro/x86_64/debian/bullseye64/DISTRO_PKGS_SPECS-debian-bullseye
```diff
@@ -618,6 +618,7 @@ yes|xclip|xclip|exe,dev,doc,nls
 no|xcur2png||exe #pcur needs this
 no|xdelta||exe
 no|xdg-puppy-jwm||exe
+yes|xdg-utils|xdg-utils|exe,dev,doc,nls
```
01micko, line 621: I did it this way in raspup. Is there any advantage in your way?

```
yes|xdg-utils|xdg-utils|exe>null,dev>null,doc>null,nls>null #fake install
```












'

dimkr: xdg-utils contains more than just xdg-open. I considered a "fake install", but decided to fix this issue without breaking stuff that needs the other executables in xdg-utils.

PR #2271, opened 2021-05-19
Small display improvement aimed at GTK3 builds

peabee:


01micko (commented):
Screenshot please?

peabee:
Before:
Screenshot
After:
Screenshot(1)

dimkr:
I'm not a UX engineer by trade, but I find it absurd that half of this window is reserved for one sentence and a checkbox. Maybe, the resolution label and the checkbox should be a row under the frame containing the three comboboxes.

Generally, it seems to me that the combination of both horizontal and vertical elements in one UI makes it harder to navigate and less convenient with a small screen.

01micko:
This is just a 2 line deletion

quicksetup

peabee:
Seems good to me - go for it!

01micko:
Actually have it better now :) -- just testing in gtk2 as well (the main window _should_ be `scrollable="true"` for lower res)

quick2

01micko:
A bit crappy in gtk2 :( -- should be fixable. (using qemu)

quick2(1)

01micko:
I went through the tedious task of testing and tweaking in gtk2 and gtk3 at my native res and 2 really small res - 640x480 and 800x600.

The upshot is that `--geometry="WIDTHxHEIGHT` must be used.

Anyway, I'll post just a couple of screenshots

640x480 gtk2 initial screen (gtk3 looks very similar but scrollbars fade in)
quick-gtk2-small res 0

640x480 gtk3 scrolled bottom right (gtk2 looks similar but line wrapping is different - plus fixed scrollbars)
quick-gtk3-small res 1

800x600 gtk2
quick-gtk2-800x600

800x600 gtk3
quick-gtk3-800x600

So @peabee I'll close this and issue a new PR.

peabee:
Thanks @01micko

PR #2272, opened 2021-05-19
Reduce the squashfs block size when using zstd

dimkr:
https://forum.puppylinux.com/viewtopic.php?p=25809#p25809

### DIFF
--- a/woof-distro/x86/slackware/14.2/_00build.conf
+++ b/woof-distro/x86/slackware/14.2/_00build.conf
@@ -60,7 +60,7 @@ LICK_IN_ISO=yes
 #SFSCOMP='-comp xz'
 #SFSCOMP='-comp gzip'
 #SFSCOMP='-noI -noD -noF -noX'
-SFSCOMP='-comp zstd -Xcompression-level 19 -b 512K'
+SFSCOMP='-comp zstd -Xcompression-level 19 -b 256K'
 
 ## if "$WOOF_HOSTARCH" = "$WOOF_TARGETARCH"
 ## This is usually not needed
--- a/woof-distro/x86/slackware/14.2/_00build.conf
+++ b/woof-distro/x86/slackware/14.2/_00build.conf
@@ -57,10 +57,9 @@ LICK_IN_ISO=yes
 
 ## compression method to be used (SFS files)
 #SFSCOMP='-comp xz -Xbcj x86 -b 512K'
-#SFSCOMP='-comp xz'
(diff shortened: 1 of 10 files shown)


woof-distro/x86/slackware/14.2/_00build.conf
```diff
@@ -60,7 +60,7 @@ LICK_IN_ISO=yes
 #SFSCOMP='-comp xz'
 #SFSCOMP='-comp gzip'
 #SFSCOMP='-noI -noD -noF -noX'
-SFSCOMP='-comp zstd -Xcompression-level 19 -b 512K'
+SFSCOMP='-comp zstd -Xcompression-level 19 -b 256K'
```
01micko: A problem with slackware(64)-14.2 is lack of support for `zstd` due to the age. I should have mentioned this awhile ago. Sorry!

01micko (changes requested):
Please change slacko and slacko64 based on slackware 14.2 back to xz compression for squash please Yes I know it messes up the CI too.

woof-distro/x86/slackware/14.2/_00build.conf
dimkr: I've tested it with 4.19.x, and it boots just fine!

```
$ grep SQUASHFS_ZSTD kernel-kit/configs_x86*/DOTconfig-4.19* | grep =y
woof-CE/kernel-kit/configs_x86_64/DOTconfig-4.19.160-x86_64:CONFIG_SQUASHFS_ZSTD=y
woof-CE/kernel-kit/configs_x86/DOTconfig-4.19.160-i686-pae:CONFIG_SQUASHFS_ZSTD=y
```

Both 32-bit and 64-bit.

01micko: It's not a kernel issue at all - userspace. `zstd` compressed sfs mount and unmount just fine - but try to squash or unsquash with `zstd`. Slacko-7.0x must be able to build itself!

There is no `zstd` support _at all_ in slackware-14.2.

peabee:
Rockedge has problem building FossaPup64 using BionicPup64 as it doesn't have zstd....
https://forum.puppylinux.com/viewtopic.php?p=26333#p26333

dimkr:
If you ask me, it's unrealistic to build 2020 tech on a 2018 distro, and super unrealistic to expect woof-CE developers to test it on an old distro with 3+ years old packages

PR #2273, opened 2021-05-20
xorg-autoconf: Improve Xorg detection

rizalmart:
If Xorg was not installed. Stop generating configuration file


dimkr:
@rizalmart This duplicates some of the work in #2265. There are multiple code paths that lead to xorgwizard and the generation of xorg.conf.

rizalmart:
@dimkr  it's not a duplicate. It's just a fail safe for xorg-autoconf script itself

PR #2274, opened 2021-05-22
Improve quicksetup UI re #2271

01micko:
Thanks to @peabee and @dimkr


01micko:
Here is a screenshot

Screenshot

dimkr:
Huge improvement! Efficient and organized.

Maybe the "Run Video Wizard" checkbox can be moved below "Run Internet apps as spot", to save some vertical space.

01micko:
> Maybe the "Run Video Wizard" checkbox can be moved below "Run Internet apps as spot", to save some vertical space.

Well, I think the 'Run Video Wizard' checkbox needs to stay in the 'Graphics' field but I can still save a line. Also tested on raspi, however GUI is a lot taller because there is a 'sound' field so I get the scrollbars - no problem, that's what they are for.

quick-gtk3-new(1)



I ~can~ have actually now ~shorten~ shortened the dimensions by ~10 - 12 px or so~ 22px.

Also, if a user increases global fontsize it might look a bit ugly with the scrollbars but it will still be functional.

PR #2280, opened 2021-05-23
Generate UEFI bootable x86_64 Chrome OS style images

dimkr:
These make it possible to test Chromebook builds of Puppy on recent, regular x86_64 machines that support only UEFI boot.

This special UEFI support uses the "EFI stub" feature of the kernel, which allows it to be used instead of a boot loader, but at the cost of having to hardcode the kernel command-line. To make these images able to auto-update in the future, but without using the same partition UUID all the time, the boot partition is identified by a special label.


01micko:
Interesting, so you ust rebuild the bzImage then install it to the image. :+1:

PR #2289, opened 2021-05-26
Avoid many sleep calls during init and X startup on x86_64 and ARM

dimkr:
They're still there on 32-bit x86, to avoid breaking Puppy on ancient hardware.

I don't understand the comment in `delayedrun` ... it's a problem that tray icons start after `welcome1stboot`, but we make the problem even worse by adding that `sleep 3`.

With this PR, rc.syinit takes 7 seconds in a VM on my not-very-new laptop.

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/delayedrun
+++ b/woof-code/rootfs-skeleton/usr/sbin/delayedrun
@@ -40,14 +40,14 @@ if [ ! -f /var/local/delayedrun_firstboot_flag ];then
   else
    welcome1stboot
   fi
-  sleep 3 # problem as tray applets only got launched after welcome msg exited.
+  [ "$WOOF_TARGETARCH" = "x86" ] && sleep 3 # problem as tray applets only got launched after welcome msg exited.
   touch /var/local/delayedrun_firstboot_flag
 fi
 
 if [ -d /root/Startup ];then
  for A in /root/Startup/* ; do
   $A 2>/dev/null & 
-  sleep 0.2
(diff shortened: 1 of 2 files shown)


woof-code/rootfs-skeleton/usr/sbin/delayedrun
```diff
@@ -40,14 +40,14 @@ if [ ! -f /var/local/delayedrun_firstboot_flag ];then
   else
    welcome1stboot
   fi
-  sleep 3 # problem as tray applets only got launched after welcome msg exited.
+  [ "$WOOF_TARGETARCH" = "x86" ] && sleep 3 # problem as tray applets only got launched after welcome msg exited.
```
01micko, line 43: Probably can get rid of that sleep entirely as most tray applets are started by the WM with `/etc/xdg/autostart/*.desktop` or `$HOME/.config/autostart/*.desktop`

01micko (approved):

woof-code/rootfs-skeleton/usr/sbin/delayedrun
dimkr: That's true, I wonder how many PETs still use /root/Startup and suffer from slow start ...

01micko: Not many I suspect but the ones that count are numlockx, but who cares? I'm not worried about a delay in my numlock - if configured!

01micko: Just ditch it I say, just commit directly - it's trivial.

Issue #2278, opened 2021-05-23
proprietary firmware build in 3builddistro is broken


01micko:
see my [comment](https://forum.puppylinux.com/viewtopic.php?p=26121#p26121)

It's not just a legality issue, the actual build is broken. I'll take a look as time permits.

- [x] fix logic in `3builddistro`
- [ ] if an `fdrvXXX.sfs` is present add the proprietary firmware to that.
- [ ] update links
- [ ] ditch old firmwares
- [ ] add bluetooth firmwares - many of these _are_ free but haven't made it to kernel.org
- [ ] add sound firmwares - ditto
- [ ] cross reference with `zdrv*/etc/modules/firmware.lst*`

01micko:
`support/get_fw` harks back to the days when we builtin the firmware to `zdrv`. There is also a logic problem in `3builddistro`

PR #2292, opened 2021-05-30
fix broken extra firmware

01micko:
re #2278

Adding this to a `ydrv` as this routine was written when we used an `fdrv` optionally. It (fdrv) was actually only loaded with a boot parameter,


01micko:
I'm going to put this in kernel-kit

PR #2291, opened 2021-05-29
Optional, automatic, recursive dependency resolution in 0setup 

dimkr:
Right now, dpup suffers from three issues that can be solved quite easily once 0setup can automatically add missing dependencies:
1) Many "base packages" that are assumed to be present in any Debian installation are missing, because we don't always add dependencies of packages in the package list. Therefore, Debian package x that depends on package y that depends on package z may be broken, because z is missing in dpup and x assumes its presence.
2) Firefox can't play many online videos, because the "recommended" libavcodec58 dependency of Firefox is missing. However, adding it is not fun, because it pulls in a large tree of dependencies. It's impractical to make sure that all of its dependencies are included in dpup, because dependencies can change over time, and the manual process of adding all dependencies can be automated quite easily, so why not? _(by the way, libasound2-plugins pulls in libavcodec58)_
3) Some "core" Debian packages can break Puppy, if PPM installs them (for example, packages that replace /bin/init). #2267 is an example.

Most of the packages in the dpup package list can be removed after this PR, because now they're added automatically. For example, binutils dependencies were resolved manually, and now the binutils package can specify just the top-level "binutils" package.

I'll handle this in a separate PR, because this can be a breaking change if a package that nothing depends on was added accidentally, and petbuilds work only because it's there.

To opt-in for automatic dependency resolution, a DISTRO_PKGS_SPECS, ~~replace `yes` with `auto`. Not sure if it's super clear what "auto" means in this context, but I think I prefer it to `yes-recursive` or something~~ add `deps:yes` to the new, sixth field. Right now, wildcards like `firefox-esr-l10n-*`, ~~explicit repo choice~~, etc', conflict with this feature but this can be fixed in the future.

PR #2293, opened 2021-05-30
Add pavucontrol to dpup

dimkr:
Behold the power of automatic dependency resolution (#2291)!


01micko:
You could have just committed this!

PR #2301, opened 2021-06-05
Use the compact Firefox tab bar in Firefox 89

dimkr:
https://support.mozilla.org/en-US/kb/compact-mode-workaround-firefox


dimkr:
This is what Firefox 89 looks like with this PR:

ff

Firefox isn't great at drawing the tile bar because the window can't be dragged. It's a dilemma between a thick tab bar plus a separate title bar that waste lots of vertical space, or inability to drag the window unless pressing Alt. If most users maximize the browser most of the time, I think the first option is better, assuming the tab bar will expand in the future when the "compact" option is finally removed.

PR #2302, opened 2021-06-05
Remove the finn user

dimkr:
It's not "experimental" as loginmanger says, it's simply broken. It doesn't work, and I don't see anyone working on fixing it.


01micko (approved):
This is fine.

`loginmanager` does need some apps added such as many internet facing apps, but there lies a problem with hyphenated apps in the checkbox variable but this is the subject of another PR so I'll approve.

peabee (commented):
Unsure how to respond to this....
Removal of the non-root user might play into the hands of people who say "don't use Puppy because it is insecure"......however.....
I've never even tried to use finn.....
I've not seen forum posts from people with problems with finn until very recently....
If finn is provided it needs to work....but how broken is it and how much effort is needed to fix it?
There will still be lots of builds out there with broken finn even if its removed for future builds.
A discussion on the forum may surface somebody who is interested enough to put the effort into making it work?
On balance I'd probably leave it be until we're sure it can't be fixed.....there doesn't seem to have been much downside to shipping builds with a broken finn so maybe strengthen the warnings as an interim bodge.....

dimkr:
> Removal of the non-root user might play into the hands of people who say "don't use Puppy because it is insecure"......however.....

But ... what's the use of a broken feature? If you can use Puppy as a non-root user, but it doesn't work, it's the same as not having this feature, but at least, it's not misleading.

I think removing it is the right choice, for two reasons: 1) it's misleading users, so even if it's going to get fixed in the future, keeping it there is causing damage and 2) if it's been like this for so much time, I think nobody will ever fix it, or at least, nobody will fix it and maintain afterwards. There's no point in calling it "experimental" when in reality, it's rotten to the core and might leave you unable to start X and turn it off.

01micko:
Why not branch it off then and maybe it can get fixed later?

dimkr:
Does anyone volunteer to fix it?

01micko:
I would if I had the time ...

At one point a while ago I had code that started X as Fido. I never merged it. It's floating around the old forum somewhere.

peabee:
> Does anyone volunteer to fix it?

I think we need to widen the potential contributors by raising it as an issue on the forum and asking for help from anyone interested in retaining the "finn" facility........

01micko:
I'll just branch it off and cherry pick from testing I suppose.

dimkr:
> I think we need to widen the potential contributors by raising it as an issue on the forum and asking for help from anyone interested in retaining the "finn" facility........

I'm fine with giving somebody in the forums the chance to save finn, but do you agree to approve this PR if this attempt fails? This PR can be squash merged, so it's a single commit that can be reverted in the future when someone wants to restore finn and start fixing issues.

peabee:
> do you agree to approve this PR if this attempt fails?

Yes indeed - as said earlier it should work if it is to continue into the future....

Which build should we ask a volunteer to work on? A stable established build like FossaPup64 or a newer more fluid build like Slacko64-8?
And how should we ask them to present their fixes so that they can be incorporated into Woof-CE once completed?
Where is the best place on the forum to post our invitation?

gyrog:
I've always been quite content to run always as "root".
I recently installed a real 'devuan' Linux to test uefi multi booting, and after years of being able to do simply anything to Puppy, I found having to use 'sudo' quite annoying.

I've never tried to use 'spot' or 'finn', so i've no idea if they are at all useful.

Having a broken 'finn' is doing Puppy no favours, I wouldn't miss it if it were removed.
But if @01micko, wants to put it on his back burner, could it be archived, removed, and then possibly re-introduced as a workinng 'finn'.

dimkr:
Closing this until nobody objects to this PR.

01micko:
I don't object, and think it _should_  be merged, but in saying that it _should_ be fixed too. I really don't have the time (#2324) to fix it :disappointed: but I wish I did.

dimkr:
Like you, @01micko, I have my time/effort/fun/attention/etc' budget, and until @peabee decides to give up on the idea of finding a brave volunteer to fix finn, my plan is to merge this PR (plus other PRs that didn't get merged for similar reasons) into my woof-CE fork and always build from there.

Convincing everyone to believe in my software engineering values (focus, stability of important features > number of unpopular features, honesty, etc') is not a priority for me, and I don't have the time for this.

peabee:
Would a compromise be to leave all the code in place but disable the feature until such time as effort can be found to fix it?
Bit like full install functionality is still there but needs to be activated with a symbol export in a terminal to function......

There seems to be some indication of interest from @rockedge on the forum.....

dimkr:
Merging this PR and reverting my commit would achieve the same result as adding an on/off switch that defaults to off, and removing it later.

(squashed the branch so it's a single commit that can be reverted)

peabee:
Somewhat different - build time vs runtime............
Anybody wanting to work on it now has to do a build beforehand......
Oh well..........

01micko:
Created branch `finn-fix` so merging this.

peabee:
> Created branch `finn-fix` so merging this.

I don't see this branch - is it hidden??

PR #2304, opened 2021-06-05
Allow the display settings wizard to run on Xwayland

dimkr:
Right now, the button in wizardwizard is grayed out, but it's still possible to run the display settings wizard directly from the menu and nothing works. Instead, every feature that doesn't work on Xwayland should be grayed out, while the wizard itself is still accessible.

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/wizardwizard
+++ b/woof-code/rootfs-skeleton/usr/sbin/wizardwizard
@@ -57,7 +57,6 @@ export WizardWizard='
       </button>
       '$BUTTON_22'
       <button image-position="2">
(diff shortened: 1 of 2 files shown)


01micko (commented):
What about the glx-gears and report-video buttons?

I think report-video can be made wayland compliant but IDK if glx-gears would run.

Using a dummy value in screeny for `WAYLAND_DISPLAY`

xw

dimkr:
report-video, glxgears and screensaver are the three buttons that are not grayed out. They work under Xwayland.

The screensaver runs on Xwayland via swayidle, and glxgears should work under any Wayland compositor that supports Xwayland.

PR #2306, opened 2021-06-06
fix quicksetup + and - bug for timezones that use GMT+ or GMT-

01micko:
also fixes psync
added new script 'fix_gmt'

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/fix_gmt
+++ b/woof-code/rootfs-skeleton/usr/sbin/fix_gmt
@@ -1,0 +1,20 @@
+#!/bin/sh
+
+# timezone fixer upperer
+# helper for quicksetup (and friends) and psync
+# inverts the '+' or '-' for gui display because in 
+# /usr/share/zoneinfo/Etc GMT[+-] are inverted!
+
+fix_gmt_display() {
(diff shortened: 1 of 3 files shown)


peabee (approved):
Looks OK to me (not really understanding what the problem is - living in the UK GMT+0 and GMT-0 are the same..)

01micko:
See [greengeeks problem](https://forum.puppylinux.com/viewtopic.php?f=118&t=2893&p=27330#p27330)

gyrog:
@01micko,
Does this code eventually use the actual TZ variable?

I assume that you are aware that the actual TZ variable has the signs backwards, e.g. TZ=AEST-10 is correct for Queensland, at GMT+10.
And only the signed number of the TZ variable is significant in setting the timezone.

01micko:
> @01micko,
> Does this code eventually use the actual TZ variable?
> 

No

> I assume that you are aware that the actual TZ variable has the signs backwards, e.g. TZ=AEST-10 is correct for Queensland, at GMT+10.
> And only the signed number of the TZ variable is significant in setting the timezone.

Yes. That is what this PR addresses.

gyrog (approved):
Ok

PR #2321, opened 2021-06-12
Fix deletion of big Mesa drivers

dimkr:


peabee (commented):
Sounds good if it doesn't have any consequences ..... what do the deleted drivers do? what will be the effect of deleting them?

rizalmart:
Oh boy deleting vmware dri maybe a bad idea.

dimkr:
The drivers this template already tried to delete are all used by some VMs. But OpenGL should still work, using swrast. I only added zink to the list, because I don't think it's in use today.

01micko:
Why closed?

dimkr:
@01micko I think @peabee and @rizalmart oppose this change 🤔

Issue #2325, opened 2021-06-14
GTK+ 3 pnote eats 60 MB of RAM on x86_64


dimkr:
It's on by default, and it's a pretty big waste of RAM.

It's true that dialog and especially its GTK+ 3 variant can be resource hungry, but pnote deserves special attention because it's the only long-running gtkdialog process in a clean Puppy.

#2326 is only a partial solution, because it only reduce pnote's RAM consumption, and only on x86.

dimkr:
@01micko How about disabling pnote by default? (instead of this hack, https://github.com/dpupos/woof-CE/blob/59911efa1ca8f43e47738ac14554a1f6f5c1ed50/woof-distro/x86_64/debian/bullseye64/_00build.conf#L144)

01micko:
Done in 5e6ccb77b

PR #2326, opened 2021-06-14
Use X.Org and GTK+ 2 petbuilds on x86, to save RAM

dimkr:
Testing in https://github.com/dimkr/woof-CE/actions/runs/936219054


dimkr:
@01micko This can save RAM on old x86 machines that don't benefit much from GTK+ 3 (because Wayland won't work, won't work well, or the GPU is too slow to justify the extra RAM usage of GTK+ 3). Also, this will make future 32-bit builds more familiar to those who stick to some old 32 bit Puppy only because it consumes less RAM and the latest versions of browsers still work on it.

However, this will make 32 bit and 64 bit builds different. ROX-Filer (GTK+ 2) and browsers (GTK+ 3) will look the same, but applications (including gtkdialog based stuff) won't. This increases the bug surface and doubles the testing work in some cases.

What do you think? Is it worth it?

01micko:
> What do you think? Is it worth it?

Probably. Let's face it, why would you even run an x86, i?86 distro? Because you are trying to keep an old clunker going - no other valid reason. This a strength of Puppy I suppose but the users must report bugs or they won't get fixed. 64 and arm/aarch should be priorities really but I think the bugs should be fairly minimal in x86, at least at our end. Third parties? IDK, but 32 bit users must understand that we haven't the dev base to fully test.

dimkr:
Pretty cool, built a 32-bit dpup now, with GTK+ 2 petbuilds and X.Org. It builds a bit faster, uses only 80 MB of RAM (with pnote disabled, #2325) and the GTK+ 3 applications (Firefox, Transmission ...) look consistent.

PR #2327, opened 2021-06-14
Use overlayfs instead of aufs

dimkr:
This might speed up woof-CE!

Testing in https://github.com/dimkr/woof-CE/actions/runs/936219054

### DIFF
--- a/woof-code/_00func
+++ b/woof-code/_00func
@@ -175,16 +175,16 @@ EOD
 }
 
 function set_layer_type() { # sets $LAYER_TYPE
+	LAYER_TYPE='overlay'
+	modprobe overlay >/dev/null 2>&1
+	cat /proc/filesystems | grep -q 'overlay' && return
+	lsmod | grep -q '^overlay ' && return
+	#-
 	LAYER_TYPE='aufs'
 	modprobe aufs >/dev/null 2>&1
 	cat /proc/filesystems | grep -q 'aufs' && return
 	mount | grep ' / ' | grep -q 'aufs' && return
 	lsmod | grep -q '^aufs ' && return
-	#-
-	LAYER_TYPE='overlay'
-	modprobe overlay >/dev/null 2>&1
-	cat /proc/filesystems | grep -q 'overlay' && return
-	lsmod | grep -q '^overlay ' && return
 	#-
 	echo "ERROR: need aufs or overlay"
 	exit 1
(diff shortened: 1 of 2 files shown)


gyrog:
I notice that you have removed the line of code to delete aufs whiteout files,
but I don't see any code to delete overlayfs whiteout files, why?
From memory, overlayfs whiteout files are 0 length character device files with a name identical to the "deleted" file, no ".wh." prefix.
Somewhere in there I decided that mucking around with whiteout files was "hairy" stuff that I want to always avoid.

I've never tested the speed of aufs vs overlayfs, though "they" claim it is very efficient because it does only directory processsing, but I wonder if it will be noticeable, did you notice it?

dimkr:
> but I don't see any code to delete overlayfs whiteout files, why?

I'll take a look at petbuild output, thanks. As far as I understand, we don't delete files during the `make install`, so there shouldn't be any whiteout files. Maybe I added that deletion because some of the source code we use is polluted with .wh files and we copy directories recursively.

> I've never tested the speed of aufs vs overlayfs, though "they" claim it is very efficient because it does only directory processsing, but I wonder if it will be noticeable, did you notice it?

It's really hard to tell if overlayfs makes builds faster, because build time vary a lot, even if you run the same build twice. But it looks like some people believe overlayfs is faster, at least in some scenarios, so I hope they're right. Anyway, the main reason to use overlayfs is using the mainline kernel file system instead of a large third-party patch,.

gyrog:
@dimkr,
I've used overlayfs local stacks to merge various directories for some time now and never worried about whiteout files. When I'm compiling a kernel I use a local directory as the top RW layer, and my local woof-ce clone as a RO lower directory, this keeps the woof-ce clone clean.
So you are probably correct about the whiteout files.

The only problem with the "using the mainline kernel file system instead of a large third-party patch" argument is that if that patch doesn't work, then Puppy don't run, aufs simply has to be there.

Assuming that it works, I don't have aproblem with your code, for what it's worth.

rizalmart:
Overlayfs can't add/remove layers on the fly. Some great features of Puppy will be lost. Also most of the Puppy core scripts are optimized for AUFS. If Overlayfs offers compatibility and same feature with AUFS, then go jump aboard the Overlayfs

01micko:
So @dimkr is it faster? If you're happy that it is I'll merge.

dimkr:
> So @dimkr is it faster? If you're happy that it is I'll merge.

It's hard to tell, the build times vary a lot even without any changes 🤷🏿 

I think this PR deserves more testing before it's merged.

01micko:
> > So @dimkr is it faster? If you're happy that it is I'll merge.
> 
> It's hard to tell, the build times vary a lot even without any changes 🤷🏿
> 
> I think this PR deserves more testing before it's merged.

Ok.

Issue #2328, opened 2021-06-14
Volume keys don't work with Xwayland and PulseAudio


dimkr:
Not sure if this problem affects all builds, or just dpup, only when using PulseAudio, only when using Xwayland ...

It's tempting to add JWM key bindings for `pactl set-sink-volume @DEFAULT_SINK@ +5%` and `pactl set-sink-volume @DEFAULT_SINK@ -5%`, but I'd like to know what is causing this and what builds are affected.

PR #2329, opened 2021-06-14
Fix volume key bindings when using PulseAudio (#2328)

dimkr:
~Not tested yet~


dimkr:
Tested under QEMU and looks good. When `pactl` fails, we fall back to `amixer`, so the issue with PulseAudio is fixed without breaking ALSA.

PR #2335, opened 2021-06-17
Use the dpup or slacko logo for the thin tray menu button

dimkr:


01micko:
Ok, I don't need to think about this one :+1:

dimkr:
JWM is horrible at scaling, it's pixelated. Maybe I'll open another PR that adds `puppylogo16`, 48x48 scaled down to 16x16 during 3builddistro using rsvg-convert.

PR #2349, opened 2021-06-21
Bluetooth audio support in dpup

dimkr:
- [x] Test on x86_64
- [ ] ~Test on ARM~ Will test later
- [ ] ~Test with headphones~ Will test later, forgot the headphones at the office 😨 
- [x] Test with speakers
- [x] Test in Firefox running as root
- [x] Test in Firefox running as spot
- [x] Fix audio in Firefox running as spot

---

This PR changes how D-Bus and PulseAudio work in Puppy.

Some applications, like Blueman, consist of an unprivileged daemon that runs through a "session" instance of D-Bus, and a separate GUI process. Traditionally, Puppy has only a "system" D-Bus instance and these applications just don't work. Therefore, this had to be fixed.

Secondly, because root and spot see different PulseAudio output devices when they use different instances of PulseAudio, and spot can't use a privileged PulseAudio instance, we must switch to a single PulseAudio instance to allow both users to use the same list of output devices. By making the "session bus" used by PulseAudio exit when the X server exits, we make sure PuleseAudio is stopped or restarted with the X server.

Third, because normally, the "session bus" uses the logged in user's credentials (so the daemon and the GUI components of applications like Blueman can talk to each other), we have to make the "session bus" that runs as spot the "session bus" of root too, otherwise processes running as root can't talk to PulseAudio running as spot.

And finally, this PR makes pa-applet run as spot in dpup, for no reason other than consistency with Blueman. Volume and mute keys work fine, and there's no reason to run it as root.

### DIFF
--- a/woof-distro/x86_64/debian/bullseye64/DISTRO_PKGS_SPECS-debian-bullseye
+++ b/woof-distro/x86_64/debian/bullseye64/DISTRO_PKGS_SPECS-debian-bullseye
@@ -62,7 +62,7 @@ no|bin86|bin86|exe>dev,dev,doc,nls
 yes|binutils|binutils|exe>dev,dev,doc,nls||deps:yes
 yes|bison|bison|exe>dev,dev,doc,nls||deps:yes
 yes|blueman|blueman|exe,dev,doc,nls||deps:yes
-yes|bluez|bluez|exe,dev,doc,nls||deps:yes
+yes|bluez|bluez,bluez-obexd|exe,dev,doc,nls||deps:yes
 no|boehm-gc|libgc1,libgc-dev|exe,dev,doc,nls||deps:yes
 no|busybox||exe
 yes|bzip2|bzip2,libbz2-1.0|exe,dev,doc,nls||deps:yes
--- a/woof-distro/x86_64/debian/bullseye64/DISTRO_PKGS_SPECS-debian-bullseye
+++ b/woof-distro/x86_64/debian/bullseye64/DISTRO_PKGS_SPECS-debian-bullseye
@@ -61,6 +61,8 @@ yes|bdb|libdb5.3|exe,dev,doc,nls||deps:yes
 no|bin86|bin86|exe>dev,dev,doc,nls
 yes|binutils|binutils|exe>dev,dev,doc,nls||deps:yes
 yes|bison|bison|exe>dev,dev,doc,nls||deps:yes
+yes|blueman|blueman|exe,dev,doc,nls||deps:yes
+yes|bluez|bluez|exe,dev,doc,nls||deps:yes
 no|boehm-gc|libgc1,libgc-dev|exe,dev,doc,nls||deps:yes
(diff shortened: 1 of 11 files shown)


woof-distro/x86_64/debian/bullseye64/DISTRO_PKGS_SPECS-debian-bullseye
```diff
@@ -172,6 +175,7 @@ yes|glibc_l10n|libc-l10n|exe,dev,doc,nls||deps:yes
 no|gmeasures||exe,dev>null,doc,nls
 yes|gmp|libgmp10,libgmpxx4ldbl,libgmp-dev|exe,dev,doc,nls||deps:yes #in precise, this was only in devx, but abiword needs it.
 no|gnome-doc-utils|gnome-doc-utils|exe>dev,dev,doc,nls|+python-libxml2
+yes|gnome-icon-theme|gnome-icon-theme|exe>null,dev>null,doc>null,nls>null
```
01micko, line 178: Is this really needed? What icons are missing? I can fix that.

01micko (approved):
Interesting, especially since my motivation for including pulseaudio was to better support bluetooth.

woof-distro/x86_64/debian/bullseye64/DISTRO_PKGS_SPECS-debian-bullseye
dimkr: I wanted to prevent installation of this huge package. It's a dependency of Blueman. 

PMaterial is not 100% complete (i.e some missing pavucontrol icons), but close enough and much smaller, so there's no reason to install this package.

01micko: Ok, I'll try to add icons to all 3 themes over the week end so we don't need packages like adwaita icons or gnome icons.

01micko: I do know that some microphone icons are missing - gtk3iconbrowser is my friend there :smile:

01micko: Sorry, totally missed the `>null` redirect - oops

dimkr:
@01micko I don't think this will break audio in slacko64, but I'm leaving the decision when to merge this PR, to you.

01micko:
Ok, I think this can be merged and deal with the fallout in due course.

Some (most?) of the specific dpup hacks can probably be woof-wide, but it will be probably weeks (not too many I hope) before I can find the time to get bluetooth working in slacko.

dimkr:
I think I'll add a Blueman template that dues the `setup-spot` part, so Bluetooth audio works in any Puppy with zero tinkering if `blueman` and PulseAudio's Bluetooth module are added with `deps:yes`.

Issue #2344, opened 2021-06-19
Early microcode loading


01micko:
In [this post](https://github.com/puppylinux-woof-CE/woof-CE/pull/2316#issuecomment-864349820) in #2316 I performed a little experiment to see if early microcode loading would work in Puppy with our `initrd.[gx]z` and it worked! (Tested Intel only - will test AMD later).

**EDIT**: new code posted

What I propose is an entirely optional mechanism to concatenate the generated `ucode.cpio` archive to the current `initrd.gz` at the woof level and in userspace. It uses [this guide](https://www.kernel.org/doc/Documentation/x86/microcode.txt).

In woof, it can be done quite easily if you have built a kernel and added firmware from [kernel.org](https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/) for AMD but Intel microcode needs to be grabbed from [Intel's github repo](https://github.com/intel/Intel-Linux-Processor-Microcode-Data-Files/releases).
~In userspace, it is easier to grab the AMD microcode from [googlesource (direct link to tar archive)](https://kernel.googlesource.com/pub/scm/linux/kernel/git/firmware/linux-firmware.git/+archive/refs/heads/main/amd-ucode.tar.gz)~ ~(yes I can hear the sighs :scream: "No, not Google!" ).~ EDIT: **now getting code from kernel.org** :tada: 
It is a mirror of the kernel.org firmware repo and will be current. The advantage is a direct link to the tar archive. _Note: it may be possible to do a sparse checkout from kernel.org with the right pattern_
The Intel files are updated regularly.

The userspace implementation could be part of *Boot Manager* and give the options:
- AMD, Intel or both.
- whether to backup the stock `initrd.gz` or not

Anyway, here is a script, useful for userspace and woof that just gets the microcode (AMD, Intel or both) and builds the `ucode.cpio` in `/tmp`.

```sh
#!/bin/sh

# gets the latest intel and amd microcode for cpu bugs and builds
# a cpio archive containing this code in the correct hierarchy
# for early loading of microcode.
# refer https://www.kernel.org/doc/Documentation/x86/microcode.txt
# GPLv2

# the intent is for the script output to be fairly quiet

set -e
[ -e /etc/rc.d/PUPSTATE ] && . /etc/rc.d/PUPSTATE
rm -rf 
ver=0.1
PROG="${0##*\/}"

usage_() {
	echo "$PROG-$ver"
	echo
[...]

PR #2353, opened 2021-06-26
Ucode early and late loading plus build cdboot locally.

01micko:
See #2344 for all the latest discussion

To DO

- [x] integrate into `3builddistro`
- [x] Wait for updates to `frugalpup`
- [x] test early loading with other boot ~managers~ loaders like ~refind, grub,~ grub4dos. `grub4dos` works - with `errorcheck off` in `menu.lst` stanza and calling both ucode.cpio and initrd.gz on the `initrd` line. 
- [x] basic auditing - eg record when files are built; check if new ones are available.
- [x] ~get individual blob versions for Intel and record them so not downloaded twice. AMD is more problematic but they haven't been updated in a while.~ not worth the effort on my side or the end user side as it involves more grokking out from downloaded files which can be slow on on a poor connection
- [x] update code in the woof iso routine
- [x] update 'full' install routine


01micko:
Here are some screenies of the respective GUI

Bootmanager

bootmanager


My Ryzen box with ucode.cpio loaded

micro2pup_amd

My Intel box (i686)

micro2pup_intel


(woops - just let it slip that I have slacko32-8.0 in development! :laughing: )

EDIT:

I haven't forgotten other architectures too.

Raspi shots follow (gtk2)

Bootmanager

raspi-bootmanager


micro2pup

micro2pup

gyrog:
1  In 'latest_microcode.sh' you store it in '/mnt/home/$PSUBDIR',
this might not be a great idea, it might be better to extract the actual save directory from $PUPSAVE.
or save it in the / filesystem, like' /usr/etc/' or somewhere.
Current Puppy supports having a save location that has no relation to the install location,
and $PSUBDIR defines the install location.

2 In 'init.d/rc.ucode' you attempt to download microcode;
On my machines, I don't have a working network until several seconds after the desktop appears,
so even thinking of doing network stuff in 'init.d' seems extremely optimistic.

3 FrugalPup already coded, 'frugalpup-bootentry' change, and tested Ok.
It will insert the "if" statement into every grub2 boot entry it generates, quite harmless if the microcode initrd file does not exist in the install directory.
Just the resolution of an unrelated issue before release of v34.

01micko:
> 1 In 'latest_microcode.sh' you store it in '/mnt/home/$PSUBDIR',
> this might not be a great idea, it might be better to extract the actual save directory from $PUPSAVE.
> or save it in the / filesystem, like' /usr/etc/' or somewhere.
> Current Puppy supports having a save location that has no relation to the install location,
> and $PSUBDIR defines the install location.
> 

That is fully intentional. `ucode.cpio` is essentially an 'initial ram disk' and is supposed to be right next to `initrd.[gx]z`. Maybe I'm missing your point. :thinking: .. But maybe you are right. Feel free to offer advice.

> 2 In 'init.d/rc.ucode' you attempt to download microcode;
> On my machines, I don't have a working network until several seconds after the desktop appears,
> so even thinking of doing network stuff in 'init.d' seems extremely optimistic.
> 

There is a `check_route()` function that waits 30 seconds (approx- `sleep 1` is not exactly a second). If it's not enough the script fails. It can be called later like any other service script, and it only tries to download if the files aren't there. (that's the intention anyway - and this is for late loading too - nothing to do with creating the `ucode.cpio`)

> 3 FrugalPup already coded, 'frugalpup-bootentry' change, and tested Ok.
> It will insert the "if" statement into every grub2 boot entry it generates, quite harmless if the microcode initrd file does not exist in the install directory.
> Just the resolution of an unrelated issue before release of v34.

Great! Sounds good to me. release when you are comfortable.. that is why this PR is marked 'Draft' :) - so reviewers can pick it to bits and we end up with better code.

gyrog:
The problem is that '/mnt/home/$PSUBDIR' is in fact meaningless, '/mnt/home' is a symbolic link to the mount point for the save partition, '$PSUBDIR' is the sub-directory for the install location.
We've gotten away with this sort of thing, for many years because in a default frugal install, the directory that contains the "puppy...sfs" and the directory that contains the savefolder are one and the same.
BUT, this is not necessarily so, the save location can be on a different drive, in a sub-directory that is not the same as the one used for the install directory.

To write the microcode beside the "initrd.gz", beside the "puppy...sfs", you need to parse $PUPSFS.

But then we have the second part of the problem, the install directory could be on an ntfs partition, that is the Windows C: drive. In this case the user needs to be very aware that they are writting a file to that drive, before it happens. 
There is no way we can check that "Fast Startup" is in fact de-activated in Windows, so we can't be sure.
(The save location could easily be some where else, on a Linux partition.)

gyrog:
I guess I'm suggesting that copying the microcode initrd file into the install directory of a particular Puppy should be a separate deliberate step.
I'm also thinking of folk who have multiple Puppies, (I currently have 20), I would want to download the latest microcode for my CPU once, then copy it to the individual install directories when and if I so choose.

01micko:
> I guess I'm suggesting that copying the microcode initrd file into the install directory of a particular Puppy should be a separate deliberate step.
> I'm also thinking of folk who have multiple Puppies, (I currently have 20), I would want to download the latest microcode for my CPU once, then copy it to the individual install directories when and if I so choose.

No problem , I can pop a message that says 'ucode.initrd is in /tmp. Do you want to install it? [yes button] [no button]' perhaps with suggestions that they may want to store it and be careful if the partition is ntfs, however using this messes with my preferred method of creating a config file with a time stamp. Maybe I'll append the time stamp the the cpio archive? Then tailing it can retrieve the time stamp, but I'll have to test whether appending it breaks the archive or not.

> you need to parse $PUPSFS.

Will do.

This should be ok;
```
~$ PUPINSTALL=/initrd${PUP_HOME}${PUPSFS##*,}
~$ echo ${PUPINSTALL%/*} 
/initrd/mnt/dev_save/pups/slacko64-8.2.1

```

gyrog:
@01micko,
Sorry, but it won't do, again '/initrd/$PUP_HOME' is the mountpoint of the save partition, not the install partition. You would have to extract the partition name ${PUPSFS%%,*} and then find it's mountpoint, if it is mounted, if it's not, you would need to mount it.
In pupmodes 5,77,and 66, nothing is mounted by default, in pupmode=12,13, the save partition is always mounted "rw", but the install partition is mounted by default only if it happens to be the same as the save partition. 

I would suggest that rather than trying to put it in the install directory of the current running Puppy, your "download/chek-for latest" code gets a "microcode-storage-directory" from the user. Then if a 'ucode.cpio' already exists, you can directly check it's "modify" date and time, or store a config file beside it. And you have somewhere permanent to store the file.
Copying this microcode initrd to a particular Puppy install directory, could be a separate process.

01micko:
Ok @gyrog installing is optional now

```sh
~$ latest_microcode.sh -h
latest_microcode.sh-0.2

 latest_microcode.sh -h|--help : show this message
 latest_microcode.sh amd : build AuthenticAMD.bin microcode
 latest_microcode.sh intel : build GenuineIntel.bin microcode
 latest_microcode.sh [amd|intel|b|''] noinstall : build microcode, DON'T install
 latest_microcode.sh ucode-r : get the existing microcode release date and exit
 latest_microcode.sh remote-r : get the remote microcode release date and exit
 latest_microcode.sh b | OR with no arguments build both AMD and Intel microcode
```

The ` latest_microcode.sh $arg noinstall` command achieves this where are is amd, intel, '', or b (both).

The GUI calls this and the result is an option window after the `ucode.cpio` is built 

micro2pup_install


Also the version is recorded _inside_ `ucode.cpio` and if needed is extracted. The GUI does this too calling `latest_microcode.sh ucode-r` and the remote versions are called too with `latest_microcode.sh remote-r`. That way recommendations in the GUI are made according to what's installed and what the remote versions are.

Committing shortly.

01micko:
So, now if the ${PUPSFS%%,*} isn't mounted then a mount point is made and it is mounted and ucode.cpio copied if 'install' is selected on cli or if 'yes' selected in GUI.

gyrog:
Just 1 simple question:
Approximately, how big is the download for microcode?

01micko:
4MB

Most of it is Intel. The one one for Late loading only gets 1 firmware so it's small.

If you choose only AMD it is 30K

dimkr:
Is it enough to copy ucode.cpio to /mnt/home? If /mnt/home is the partition where the Puppy SFSs are, this won't work with UEFI because this file needs to be in the ESP.

woof-code/rootfs-skeleton/usr/sbin/micro2pup.sh
```diff
@@ -0,0 +1,252 @@
...
+}
+EARLY=0 # true
+
+#### exclusions and special cases
+MSG1=$(gettext 'Your initrd.gz resides on an unsupported file system, however late loading may work if your main filesystem is writeable.')
+MSG2=$(gettext 'Unsupported CPU Architecture.')
+MSG3=$(gettext 'This program does not support a Virtual Machine.')
+MSG4=$(gettext 'The puppy home drive is read only, however if saving session late loading is supported.')
+MSG5=$(gettext 'Your CPU vendor is unsupported.')
+PUPDIR=/initrd${PUP_HOME}${PUPSFS##*,}
```
dimkr, line 26: What about 'full' installations? The fixes I'll need to do for `frugalify`/PUPMODE=6 images built in https://github.com/dpupos/woof-CE are the same or almost the same as those needed to support early microcode loading in 'full' installations.

jamesbond3142:
> Is it enough to copy ucode.cpio to /mnt/home? If /mnt/home is the partition where the Puppy SFSs are, this won't work with UEFI because this file needs to be in the ESP.

@dimkr The only constraint is that the bootloader need to be able to find and load it. If the existing bootloader can only load from ESP, then, yes, of course it has to be in ESP, but otherwise it can be anywhere.

> What about 'full' installations

Same, the bootloader still have to load it, and then it will automatically get used. My tests indicated that if you have "root=" line, full-install functionality will be used (the kernel won't attempt to run /init from initramfs if "root=" is specified).
For frugalify, if I get it right, you just need to specify "init=/sbin/frugalify" (to override the default /sbin/init), right? Everything else will be as full-install setup.

dimkr:
> For frugalify, if I get it right, you just need to specify "init=/sbin/frugalify" (to override the default /sbin/init), right? Everything else will be as full-install setup.

You need to specify `root` and `init=/init`, because the frugalify executable sits alongside the SFSs. The kernel needs to mount the partition and run frugalify, which chroots into an overlayfs of SFSs at /.

After frugalify runs, Puppy behaves sort of like a full installation (it doesn't know it runs under a union file system) but with stuff like /initrd and /mnt/home (bind mounts created by frugalify to simulate a 'frugal' installation) present.

If I understand correctly, initrd.gz is loaded to memory by the boot loader, so ucode.cpio must be in the same partition as the boot loader, and it's neither /initrd nor /mnt/home, the two bind mounts to the `root` partition, created by frugalify.

gyrog:
With current Puppy uefi booting, neither the 'initrd.gz' nor the 'vmlinuz' are in  the ESP,
so I don't expect any problems with 'ucode.cpio' not being in the ESP.

dimkr:
> With current Puppy uefi booting, neither the 'initrd.gz' nor the 'vmlinuz' are in the ESP,
> so I don't expect any problems with 'ucode.cpio' not being in the ESP.

Thanks @gyrog, didn't know that. I use efilinux in my builds, so vmlinuz is alongside bootx64.efi. There's no initrd.gz with frugalify, but I'll need to put ucode.cpio there.

woof-code/3builddistro
```diff
@@ -561,9 +561,17 @@ if [ ! "$CROSFLAG" ]; then
 		./build.sh -prebuilt -auto -arch ${WOOF_TARGETARCH:-default} ${INITRD_LANG} ${INITRD_KM}
 	)
 	mv -fv ../initrd-progs/initrd.[gx]z build/ || exit 1
+	
 
 	case $WOOF_TARGETARCH in x86*)
-		cp -fv ../initrd-progs/0initrd/README.txt build/ ;;
+		cp -fv ../initrd-progs/0initrd/README.txt build/ 
+		# ucode.cpio
```
dimkr: This part should be outside of the `if [ ! "$CROSFLAG" ]; then`

woof-code/rootfs-skeleton/usr/sbin/bootmanager
```diff
@@ -346,13 +391,38 @@ else
 	XML_ACTION='sfs_load &'
 	SFSFRAME_XML="$(generic_xml)" #--
 fi
+case $KARCH in
+	x86_64|i?86|amd64);;
```
dimkr: It's always i?386 or x86_64, never amd64 AFAIK

01micko: Yeah we don't use $DEBARCH

woof-code/rootfs-skeleton/usr/sbin/micro2pup.sh
01micko: Full installs put `vmlinuz` in `/boot` IIRC, so if `ucode.cpio` is there and bootloader calls it shouldn't be an issue.

woof-code/3builddistro
01micko: Ok in #2369 I see

woof-code/rootfs-skeleton/etc/init.d/rc.ucode
```diff
@@ -0,0 +1,98 @@
+#!/bin/sh
+
+# service script to identify if ucode is present and load it if so
+# if not download it and load it if possible
+# GPLv2
+
+[ -e "$HOME/.ucode.rc" ] && . $HOME/.ucode.rc # written by micro2pup.sh
```
dimkr: is HOME set when running the init script? Why not /etc or ~?

01micko: Fixed.

gyrog:
The scripts in '/etc/init.d' are NOT run by the 'init' script.
These are the "services", $HOME is set by 'rc.sysinit', well before the scripts in '/etc/init.d' are run.

01micko:
@gyrog ~I think~ the `ucode.cpio` isn't getting copied to the `initrd.gz` location in `frugalpup`, if it exists.

gyrog:
The only changes I made to FrugalPup is for it to generate boot-entires that will load the ucode.cpio file as an initrd if it exists in the installation directory.
Since the presence or absence of the ucode.cpio file is the trigger, surely it should be the responsibility of the user to "move" it there.
I must have missed the memo indicating that FrugalPup would be expected to do anything more.

01micko:
My intention was to issue a separate PR for `cdboot` but it is tightly integrated in to the function of this PR so doing it here - re #2391

01micko:
lol .. `bottmanager` - wish I had one!

Anyway, almost ready.

01micko:
I think this PR is complete. If @dimkr @gyrog @jamesbond3142 can take a look to see if I missed anything that would be good. I intend to merge in about 24 hours.

dimkr (approved):

PR #2364, opened 2021-06-30
Speed up woof-CE and PPM by running sync less often

dimkr:
sync between two package downloads is pointless, because it's not a disaster if the battery dies during package download. It only slows down PPM, especially if the user is doing I/O intensive things while waiting for PPM to finish installation. PPM is super slow when booting from a slow flash drive, and these sync calls are one of the main reasons for that.

### DIFF
--- a/woof-code/2createpackages
+++ b/woof-code/2createpackages
@@ -286,7 +286,6 @@ do
 			rm -f packages-${DISTRO_FILE_PREFIX}/${PETPKG}.tar
 			continue
 		fi
-		sync
 		rm -f packages-${DISTRO_FILE_PREFIX}/${PETPKG}.tar
 		if [ -d ${TARGET_DIR}/usr/share/applications/ ] ; then
 			sed -i 's| %u|| ; s| %U|| ; s| %f|| ; s| %F||' ${TARGET_DIR}/usr/share/applications/*.desktop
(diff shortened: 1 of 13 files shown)


01micko:
hmmm.. there's more flushing there than a broken toilet! </attempted humour>

IDK why so many, but to that extent I've rarely touched PPM.

dimkr:
> IDK why so many, but to that extent I've rarely touched PPM.

Same here, I hardly touch PPM. I think it's a horrible thing that must be destroyed and I hate almost everything about it, especially the package list format. As people spend more and more time in the browser, PPM and package management become more irrelevant, or less important.

But users still ... well, use it, and expect it to work. This PR is a small and safe change that can speed up PPM dramatically, so I don't mind doing this. (Making woof-CE itself faster is 90% of the value of this PR anyway.)

01micko:
`pkg` is the future but that needs plenty of improvement, however I'd rather put time into hat than PPM!

I reckon just merge this.

sc0ttj:
I'll be happy to pull all the `sync` calls from Pkg - they're only in there cos I copied the PPM and assumed they must be useful/needed.. 

Also, feel free to suggest any bugs/fixes etc to Pkg at the Pkg gitlab or github, I'll pick up what I can understand ;)

dimkr:
Maybe we can move PPM to rootfs-packages, default to `false` in future Puppies, and always add Pkg if PPM is out.

01micko:
Ok, I'm going to merge but as a suggestion, when we get time we should help @sc0ttj more on `pkg`. I'm sure it can be a more than good replacement for PPM.

01micko (approved):

dimkr:
I'll start contributing to Pkg - I'll do the same sync cleanup there, and add support for the Pre-Depends field.

PR #2365, opened 2021-06-30
Remove microcode packages from dpup

dimkr:


01micko (approved):
Makes sense since we get them in the ucode.cpio or on the fly for late loading.

01micko:
Of course depends on #2353 but it's almost ready, just need to test grub4dos and update the code in the iso booting routine.

Issue #2363, opened 2021-06-28
dbus problem with latest woof-ce builds 27-june-2021


peabee:
In both Slacko64-8.2.2 and CagePup64 (details below) problem with dbus:
```
~$ dbus-monitor --session
Failed to open connection to session bus: Did not receive a reply. Possible causes include: the remote application did not send a reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken.
```
Welcome to Slacko64 Puppy version 8.2.2, released Jun 2021
BUILD_FROM_WOOF='testing;6649757;2021-06-27 17:35:10 +1000;ALTERNATE_BUILD=15.0;EXTRA_MERGE='

CagePup64
BUILD_FROM_WOOF='testing;a795917;2021-06-26 18:25:56 +0300;ALTERNATE_BUILD=bullseye64;EXTRA_MERGE='

peabee:
@dimkr asked me to try running gvfsd-trash as spot..........
All I can report is that both gvfsd-trash and dbus-monitor --session do not produce error messages if run-as-spot....
```
/usr/libexec$ run-as-spot dbus-monitor --session
signal time=1624848266.982306 sender=org.freedesktop.DBus -> destination=:1.8 serial=2 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=NameAcquired
   string ":1.8"
signal time=1624848266.982351 sender=org.freedesktop.DBus -> destination=:1.8 serial=4 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=NameLost
   string ":1.8"
^C

/usr/libexec$ run-as-spot ./gvfsd-trash
^C

/usr/libexec$ dbus-monitor --session
Failed to open connection to session bus: Did not receive a reply. Possible causes include: the remote application did not send a reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken.

/usr/libexec$ ./gvfsd-trash
Error connecting to D-Bus: The connection is closed (g-io-error-quark, 18)
/usr/libexec$ 

```

dimkr:
I wonder why applications running as root can't connect to spot's session bus. I'll do some `strace` to see what's going on.

[...]

PR #2368, opened 2021-07-02
Use the session bus only for spot (#2363)

dimkr:
- [x] Test Bluetooth audio on dpup, when Blueman runs as spot
- [x] Test Bluetooth audio on dpup, when Blueman runs as root (should fail)
- [x] Check if dbus-monitor runs as root 
- [x] Check if gvfd-trash runs as root


peabee (approved):
My tests show:
https://github.com/puppylinux-woof-CE/woof-CE/issues/2363#issuecomment-873370397

dimkr:
Looks good on dpup, Bluetooth audio works.

01micko (approved):

PR #2379, opened 2021-07-10
Bluetooth audio support in focal64

dimkr:
This is https://github.com/puppylinux-woof-CE/woof-CE/pull/2349, but for focal64. Now, PulseAudio works for both root and spot, with Bluetooth audio support.

### DIFF
--- a/woof-code/packages-templates/bluez_FIXUPHACK
+++ b/woof-code/packages-templates/bluez_FIXUPHACK
@@ -1,0 +1,5 @@
+# hack for fossa64: /etc/init.d/bluetooth looks for /etc/default/bluetooth
+if [ "$DISTRO_BINARY_COMPAT" = "ubuntu" -a ! -e etc/default/bluetooth ]; then
+    mkdir -p etc/default
+    echo BLUETOOTH_ENABLED=1 > etc/default/bluetooth
+fi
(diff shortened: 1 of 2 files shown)


01micko (approved):
looks fine to me

mrfricks:
happy to see how blueman / pulseaudio compares to bluez-alsa.

mrfricks (approved):

dimkr:
@mrfricks I'm leaving the decision whether or not to merge this PR, to you.

The main benefit of using Blueman is the use of PulseAudio. Because most graphical audio editors, browsers, mixers, etc' use PulseAudio noawadays, it's more elegant (in some ways) to use PulseAudio across the board and avoid direct use of ALSA. For example, the way woof-CE handles PulseAudio now makes audio work for spot too, and both spot and root can see Bluetooth output devices. It's non-trivial to achieve this consistency with ALSA, as far as I'm aware.

It's true that PulseAudio pulls in some dependencies, and it's an extra daemon that needs to run, but I think it's not too bad because 1) it's just one process for both root and spot and 2) you need bluetoothd anyway 3) Bluetooth is ubiquitous and that's enough to justify good support for it, even if it costs extra RAM.

rizalmart:
@mrfricks 

> happy to see how blueman / pulseaudio compares to bluez-alsa.

I used blueman and pulseaudio, it works nicely and easy to setup bluetooth audio devices

01micko:
Ok, I'm happy to merge this. If no objection by the weekend I will do so.

PR #2390, opened 2021-07-17
Restore xz compression of vmlinuz and, zdrv and fdrv

dimkr:
5.10.x, x86_64:

- vmlinuz: 7.8 ->  6.9 MB (-11.7%)
- zdrv: 52 -> 48 MB (-7.9%)
- fdrv: 36 -> 33 MB (-9.6%)

Is the super fast decompression speed worth it? Maybe not.

- We're saving about 7 MB in Puppy's ISO size here, and 7 MB of RAM if we boot with pfix=ram.
- Files from fdrv and zdrv are needed only rarely (during boot, and when plugging in external devices that need a driver that's not loaded already). Therefore, zstd compression of fdrv and zdrv won't make Puppy feel faster; maybe only boot slightly faster. 
- We're talking about small files here, and in most computers from the last decade, the bottleneck will probably be the slow drive (a hard drive or a USB 2.0 flash drive) the SFSs sit on.


dimkr:
Double checked and zstd (main SFS and adrv) does have a noticeable impact on responsiveness on my 5th gen i5, even when booting with pfix=ram, overlayfs at / and a tmpfs as the writable layer. With zstd, after the first start, Firefox starts almost instantly, even after `echo 1 > /proc/sys/vm/drop_caches`. With xz, after the first start, there's a small delay before Firefox starts and `echo 1 > /proc/sys/vm/drop_caches` makes it start slowly. In addition, there's a small delay (<= 1s) every time I launch an application, and that includes tiny applications like l3afpad. Also, CPU is busier with xz, especially in I/O intensive workloads.

I think zstd for frequently accessed files (i.e. main SFS, adrv, nlsx) and xz for the rest (i.e. fdrv, zdrv, docx) it the right balance of size (usually +8-12% in the case of dpup) vs. responsiveness (much more responsive).

I'll do another test on my 2nd gen i5 - I'm pretty sure the results will be similar.

PR #2394, opened 2021-07-18
Move libicudata to nlsx and replace it with a tiny stub

dimkr:
It's a huge library (27 MB on bullseye64), and I don't know if it's needed in an English-only system, like Puppy by default (i.e. without nlsx loaded).

This library exports only one symbol and there's a stub library included in the icu source code, which hasn't changed for a very long time.

### DIFF
--- a/woof-code/packages-templates/icu_FIXUPHACK
+++ b/woof-code/packages-templates/icu_FIXUPHACK
@@ -2,3 +2,7 @@
 
 # remove ICU static libs (huge)
 find ../icu_DEV -type f -name '*.a' -delete 2>/dev/null
+
+# copy libicudata to NLS so libicudata_stub petbuild can replace it in EXE
+SO=`find . -name 'libicudata.so.*' -type f`
+[ -n "$SO" ] && install -D -v -m 644 "${SO}" "../icu_NLS/${SO}"
--- a/woof-code/rootfs-petbuilds/libicudata_stub/pet.specs
+++ b/woof-code/rootfs-petbuilds/libicudata_stub/pet.specs
@@ -1,0 +1,1 @@
+libicudata_stub-69.1|libicudata_stub|69.1||BuildingBlock|348||libicudata_stub-69.1.pet||libicudata stub|puppy|||
--- a/woof-code/rootfs-petbuilds/libicudata_stub/petbuild
+++ b/woof-code/rootfs-petbuilds/libicudata_stub/petbuild
@@ -1,0 +1,8 @@
+download() {
+    [ -f stubdata-0e7b4428866f3133b4abba2d932ee3faa708db1d.cpp ] || wget -t 1 -T 15 -O stubdata-0e7b4428866f3133b4abba2d932ee3faa708db1d.cpp https://raw.githubusercontent.com/unicode-org/icu/0e7b4428866f3133b4abba2d932ee3faa708db1d/icu4c/source/stubdata/stubdata.cpp
+}
+
+build() {
+    SO=`find /usr -name 'libicudata.so.*' -type f`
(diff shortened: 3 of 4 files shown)


01micko:
Probably another compelling reason for #2229

dimkr:
Interesting. Other than reduction in size (still not sure by how much), I don't see any side-effects.

EDIT: 358 MB -> 348 MB (zstd level 19)

EDIT: RAM consumption is very similar

01micko:
Hmm it's 28M in slacko too.. but your 'install' routine doesn't remove it and the 'find' routine (-type f) doesn't find the symlink.

01micko (approved):
Ok, will wave through and test in slacko shortly.

01micko:
Bummer ..
```
/usr/bin/ld: cannot open output file /usr/lib64/libicudata.so.69.1
/usr/lib64/libicudata.so.68.2: No such file or directory
collect2: error: ld returned 1 exit status
ERROR: failed to build libicudata_stub
```
version difference - I'll see what I can do.

01micko:
Slackware bug? Maybe or maybe my bug.

01micko:
Aha! Slackware `aaa_libraries` has an old copy, so it _is_ a slackware bug. I might try sending `aaa_libraries` (formerly 'aaa_elflibs`) to null in woof.

01micko:
Oh man. With due respect to Pat Volkerding, a one man show  isn't quite near perfection. Sooooo many things depend on libicudata.so.68 (aaa_libraries)  whereas the icu package contains libicudata.so.69 .. both _huge_ (~28M each!).

I'm going to try hacking the petbuild to loop over both :crossed_fingers:

01micko:
Works!
```ls
/initrd/mnt/dev_save/woof/woof-out_x86_64_x86_64_slackware64_15.0/petbuild-output/libicudata_stub-31908cd1dfdc72f2731305645e6b4f4a/usr/lib64$ ls
libicudata.so.68.2  libicudata.so.69.1
/initrd/mnt/dev_save/woof/woof-out_x86_64_x86_64_slackware64_15.0/petbuild-output/libicudata_stub-31908cd1dfdc72f2731305645e6b4f4a/usr/lib64$ for i in *; do stat -c %s $i;done
14048
14048
```

My hack:
```sh
download() {
    [ -f stubdata-0e7b4428866f3133b4abba2d932ee3faa708db1d.cpp ] || wget -t 1 -T 15 -O stubdata-0e7b4428866f3133b4abba2d932ee3faa708db1d.cpp https://raw.githubusercontent.com/unicode-org/icu/0e7b4428866f3133b4abba2d932ee3faa708db1d/icu4c/source/stubdata/stubdata.cpp
}

build() {
    SO=`find /usr -name 'libicudata.so.*' -type f`
    for so in $SO
    do
      g++ $CXXFLAGS stubdata-0e7b4428866f3133b4abba2d932ee3faa708db1d.cpp $LDFLAGS -shared -Wl,-soname=libicudata.so.`basename "$so" | cut -f 3 -d .` -o "$so"
    done
}
```

dimkr:
@01micko I assumed there's just one copy of the library, to intentionally break all builds with multiple libicudata. This shouldn't happen. https://github.com/puppylinux-woof-CE/woof-CE/pull/2397 is fine if this assumption is incorrect, or if you want woof-CE to be tolerant in this case.

peabee:
On the subject of unneeded libraries.......
there are still 2 instances of libglib: /usr/lib64/libglib-2.0.so.0.6800.2 & /usr/lib64/libglib-2.0.so.0.6800.3 in puppy_slacko64_8.2.2.sfs

dimkr:
> Oh man. With due respect to Pat Volkerding, a one man show isn't quite near perfection. Sooooo many things depend on libicudata.so.68 (aaa_libraries) whereas the icu package contains libicudata.so.69 .. both _huge_ (~28M each!).

Missed this comment.

Yes, the loop is the right (and only) solution if the are legitimate cases where a distro has two versions of icu (each with a different soname of course).

PR #2397, opened 2021-07-19
Fix icu petbuild for slacko anomaly

01micko:
the loop should work if only 1 file is found.


dimkr (approved):

dimkr:
@01micko What about the other libicudata*? We copy the original library to icu_NLS in the icu template, so we do it only for icu, and assume there's just one library. Maybe this copying needs to happen in aaa_libraries, or wherever the other libicudata* comes from.

EDIT: #2398

01micko:
@dimkr, I broke something and your PR #2398 is gone!

~Can you reissue it against `testing` and merge please.~

PR #2400, opened 2021-07-20
Add a prefix to DISTRO_NAME in scheduled builds

dimkr:


01micko:
Looks simple enough. What can go wrong here? I'm not particularly familiar with work flows.

dimkr:
> Looks simple enough.

Yes, it's supposed to be a super small change. I want to make it easy to tell if a Puppy is a proper release, or came from one of the scheduled builds. All Puppy builds have a version number, and unless somebody changes it, it's hard to make this distinction (for example, Slacko 7.0 is very similar to the "Slacko 7.0" built every two weeks, but the latter may introduce a bug).

Maybe, adding `.${build_number}` at the end of the version number (i.e. 7.0 becomes 7.0.43 in run 43 of build.yml) is a better solution, but it's a bit misleading (7.0 is not 7.0 plus 43 rounds of bug fixing and zero changes beyond that).

01micko:
So, you thinking of a different/better solution?

EDIT:
> Maybe, adding .${build_number}

What about  `-${build_number}` - dash instead?

dimkr:
> What about `-${build_number}` - dash instead?

If I put a dash in DISTRO_VERSION, this will break all pinstall.sh, quickpet, etc' that compare DISTRO_VERSION to some value without using `vercmp`. I don't think this is a real concern.

Another option is to replace the version number with the build number ("Fossapup64 43", "Slacko 43") - this makes it clear that this is not a stable release, but without producing potentially misleading or confusing version numbers like "7.0.43".

PR #2406, opened 2021-07-26
Use petbuilds for Puppy icons - closes #2404

### LINKED ISSUE #2404: icon theme maintenance
These icone themes, `PMaterial`, `Puppy Standard` and `Puppy Flat` have their own git repos so for ease of maintenance these should be removed from `rootfs-packages` and be made ~into petbuilds~ downladable in `3builddistro` or similar.

01micko:
The current Puppy home grown icon themes, `PMaterial`, `Puppy Flat` and `Puppy Standard` are currently in `woof-code/rootfs-packages`. This is cumbersome to maintain as commit e44f552 suggests.

These themes have had their own github repos for some time and similarly to @dimkr 's gtk themes  [gtk3_flat_grey_rounded](https://github.com/puppylinux-woof-CE/gtk3_flat_grey_rounded) and [gtk3_polished_blue](https://github.com/puppylinux-woof-CE/gtk3_polished_blue) maintenance of these icon themes would would be much simpler if they were petbuilds in `woof-code/rootfs-petbuilds`.

This branch addresses that.

**advantages**
- simpler to maintain
- reduce risk of commit mistakes
- get the latest version with bugfixes on every build
 
**disadvantages**
- slows down `3builddistro` a little
- requires `PETBUILDS=yes` in the build config (`_00build.conf`)

I took the liberty to build (from `petbuild_icons` branch) and test the latest slacko64 in `qemu` - screenshots follow -

brightmouse

buntoo

darkgray

As you can see from the diffs below, changing the icons can result in massive diffs and are near impossible to review.

#2404


peabee (commented):
Very difficult to review such ginormous changes (20106 files)..... web browser meltdown!!....
Unsure petbuilds are "appropriate" for noarch pets especially fairly "static ones" like icon packs.....
Don't like "disadvantages" - don't want 3builddistro to be any slower; don't currently use woof-ce petbuilds so switching on just for these noarch pets seems overkill - that said I haven't so far investigated what happens if petbuilds are activated (as I don't need to...)
We seem to have a plethora of ways of providing "common" components which must be very confusing to new woof-ce users.....
If you really want to make the change then I can live with it.....but not a fan!

01micko:
@peabee the icons _can_ go to `no-arch` repo but they may not get the latest bugfixes/features, however you won't need petbuilds. All you need to do is to copy the current icons in `testing/woof-code/rootfs/packages/` to respective folders with version number (see pet.specs) and run dir2pet, and I recommend using `xz` compression, heaps smaller. At the time of this post they are the latest version.

Part of the reason for this is the massive amount of changes in commits, there's thousands of icons, all text based `svg`.

dimkr:
I propose a third solution, that will satisfy the requirements of moving the icon sets outside of this repo, and not requiring @peabee to use petbuilds: git submodules.

Everything will be the same as today, but those of us that run woof-CE locally instead of running it on GitHub Actions will have to `git submodule update --init --recursive` to make sure they're using the correct version of the icons, and always clone the repo with `--recursive`.

**The icons will always be tied to a specific commit of the icons repo, and we'll need to always `git pull` (inside the submodule) to update the submodule so it points to the latest commit in master.** It's both an advantage (can't accidentally update to a commit that breaks the icons) and a disadvantage (must update the icons submodule in woof-CE).

@01micko I can help with this transition - if you push the icons to separate repos, I can set up the submodules, clone recursively when running on GitHub Actions, and make `merge2out` run `git submodule update --init --recursive`, to ease the transition.

EDIT: now I see we don't want the repo in the package as-is, because there's a build script. Yes, something needs to run this build step, and petbuilds or some similar mechanism is the only solution I see.

dimkr:
There are remains of the old packages, for example, `pmaterial_icons="true"` in the various rootfs-packages.conf

```
./arm/raspbian/buster/support/rootfs-packages.conf:pmaterial_icons="true"
./arm/raspbian/buster/support/rootfs-packages.conf:puppy_flat_icons="true"
./arm/raspbian/buster/support/rootfs-packages.conf:puppy_standard_icons="true"
./x86_64/ubuntu/focal64/adrv/rootfs-packages.conf:pmaterial_icons="true"
./x86_64/ubuntu/focal64/adrv/rootfs-packages.conf:puppy_flat_icons="false"
./x86_64/ubuntu/focal64/adrv/rootfs-packages.conf:puppy_standard_icons="false"
./x86_64/ubuntu/focal64/support/big-rootfs-packages.conf:pmaterial_icons="true"
./x86_64/ubuntu/focal64/support/big-rootfs-packages.conf:puppy_flat_icons="false"
./x86_64/ubuntu/focal64/support/big-rootfs-packages.conf:puppy_standard_icons="false"
./x86_64/ubuntu/focal64/support/rootfs-packages.conf:pmaterial_icons="true"
./x86_64/ubuntu/focal64/support/rootfs-packages.conf:puppy_flat_icons="false"
./x86_64/ubuntu/focal64/support/rootfs-packages.conf:puppy_standard_icons="false"
./x86_64/ubuntu/focal64/support/rootfs-packages.conf:puppy_flat_icons="false"
```

01micko:
> EDIT: now I see we don't want the repo in the package as-is, because there's a build script. Yes, something needs to run this build step, and petbuilds or some similar mechanism is the only solution I see.

What about a work flow? Icons could be built weekly and the tar file available for download.

dimkr:
> What about a work flow? Icons could be built weekly and the tar file available for download.

I can set up automated releases of a tarball containing the generated icons (for every push to master), and the petbuild will download and extract it.

01micko:
> > What about a work flow? Icons could be built weekly and the tar file available for download.
> 
> I can set up automated releases of a tarball containing the generated icons (for every push to master), and the petbuild will download and extract it.

Does that help the `PETBUILDS=no` situation though? Probably no different to the PR.

dimkr:
> Does that help the `PETBUILDS=no` situation though? Probably no different to the PR.

It doesn't help, but automated, prebuilt releases will allow some ugly solution like downloading a ready-made package during pinstall.sh.

01micko:
Ok, I've converted this to a draft pending changes to the icon repos and woof.

01micko:
> > Does that help the `PETBUILDS=no` situation though? Probably no different to the PR.
> 
> It doesn't help, but automated, prebuilt releases will allow some ugly solution like downloading a ready-made package during pinstall.sh.

@dimkr take a look at https://github.com/01micko/Pmaterial

I'm trying to learn the github CI / workflow stuff. If I can produce a tarball there I can easily grok out the latest release for direct download to woof in `3builddistro`

A config entry in `_00build.conf` can download all, any or none of the tarballs (all of course will be the default recommendation), totally independant of `petbuilds`

dimkr:
> @dimkr take a look at https://github.com/01micko/Pmaterial
> 
> I'm trying to learn the github CI / workflow stuff. If I can produce a tarball there I can easily grok out the latest release for direct download to woof in `3builddistro`

Uploading an artifact is not the same as downloading a release. I'll try to write a script that downloads an artifact with a specific name from a repo - documentation is at https://docs.github.com/en/rest/reference/actions#artifacts, and it should be fairly similar to https://github.com/dpupos/scripts/blob/master/get-latest-tarball.sh.

dimkr:
```
#!/bin/sh
  
ARTIFACTS=`wget -O- --header "Accept: application/vnd.github.v3+json" https://api.github.com/repos/$1/actions/artifacts`

COUNT=`echo "$ARTIFACTS" | jq .artifacts | jq length`
if [ -z "$COUNT" ] || [ $COUNT -eq 0 ]; then
    echo "Failed to list artifacts"
    exit 1
fi

for i in `seq 0 $(($COUNT - 1))`; do
    ARTIFACT_NAME=`echo "$ARTIFACTS" | jq .artifacts | jq -r ".[$i].name"`

    if [ "$ARTIFACT_NAME" = "$2" ]; then
        echo "Selecting $ARTIFACT_NAME"
        ARTIFACT_ID=`echo "$ARTIFACTS" | jq .artifacts | jq -r ".[$i].id"`
        DOWNLOAD_URL=`wget -O- --header "Accept: application/vnd.github.v3+json" https://api.github.com/repos/$1/actions/artifacts/${ARTIFACT_ID} | jq -r .archive_download_url`
        wget -O "$3" "$DOWNLOAD_URL"
        exit $?
    fi
done

exit 1
```

```
bash -x get-latest-artifact.sh 01micko/Pmaterial PMaterial-tarball /tmp/1.tar.xz
```

It's not ideal, you need a GitHub access token in ~/.netrc that has the right scopes. I'd use releases instead of artifacts, so anyone, including those that authenticate against GitHub using SSH keys, can download the tarball. Also, releases give you a stable URL and you don't have to find the artifact ID.

You can use https://github.com/dpupos/frugalify/blob/master/.github/workflows/release.yml and https://github.com/dpupos/woof-CE/blob/f576e1abdbdc1d6dd2e04cafcb1f68fecba594fe/woof-code/support/cros_image.sh#L125 for reference.

01micko:
> It's not ideal, you need a GitHub access token in ~/.netrc that has the right scopes. I'd use releases instead of artifacts, so anyone, including those that authenticate against GitHub using SSH keys, can download the tarball. Also, releases give you a stable URL and you don't have to find the artifact ID.
> 
> You can use https://github.com/dpupos/frugalify/blob/master/.github/workflows/release.yml and https://github.com/dpupos/woof-CE/blob/f576e1abdbdc1d6dd2e04cafcb1f68fecba594fe/woof-code/support/cros_image.sh#L125 for reference.

My intention is to produce a 'release' (my link is WIP) but there are caveats like the 'tag' which I think I can get right; I want to do most of this from CLI, which I think is doable. Baby steps here :baby: (lol)

dimkr:
> My intention is to produce a 'release'

When I use the word "release", I'm not talking about releasing a stable version with no known bugs or something. I'm merely using "release" in the same sense as in GitHub's API documentation. You can upload an artifact or a GitHub release (which can be marked as a pre-release) using GitHub Actions, and I recommend you to use releases instead of artifacts, because it's much easier to work with them.

01micko:
Yes, making the 'release' is the next step in my learning curve.

01micko:
https://github.com/01micko/Pmaterial/releases :octocat: 

```shell
#!/bin/sh

ICONREPO=$1

[ -z "$ICONREPO" ] && exit 1

SITE=https://github.com
GUSER=01micko
RLSE=$(curl -s ${SITE}/${GUSER}/${ICONREPO}/releases | grep -o "\/${GUSER}.*\.tar\.xz" | head -n1) # gets latest
DLD="${SITE}${RLSE}"
# download
wget -t 1 -T 10 $DLD
```

Thanks for the help @dimkr . I can move forward now.

01micko:
Going to close this and start a new one

01micko:
Ok, deleting (again!) since we have shiny new petbuild mechanism. :smile:

peabee:
Observations:
1. petbuilds make system builds slower because the devx is populated even if the devx is not being built and the petbuilds do not need updating......
2. empty directories are (unnecessarily?) deleted from packages-****

dimkr:
> 1. petbuilds make system builds slower because the devx is populated even if the devx is not being built and the petbuilds do not need updating......

The code that builds the devx is in 3builddistro, and only the petbuilds "know" if they need it, and only petbuilds.sh "knows" if any petbuild needs to run. The only way to solve problem (1) is to integrate petbuilds.sh into 3builddistro and build the devx conditionally, but that's ugly. 3builddistro is super long and messy already.

Do you have a better idea?

(Personally, I don't think that building the devx needlessly is a huge problem, because 1) we have SSDs in 2021, so it's super fast - copying files to sandbox3/devx is super fast and the slow, CPU-bound work of compressing it is skipped, and 2) most builds have the devx enabled anyway. For me, keeping 3builddistro more maintainable > less disk activity in builds without devx.)

peabee:
Quite fast - but not "super fast" - I am running off SSD and 3builddistro is noticeably slower with petbuilds enabled...
It was an observation - I can live with it.
I mainly build with devx=no as I can tell from what has been updated whether a devx rebuild is needed - squashsfs of devx is a very slow process (even with 8-cores) that I avoid unless it is needed.

peabee:
Perhaps..........
if /sandbox3/devx exists with contents and BUILD_DEVX=no
then do not re-populate /sandbox3/devx just use existing contents

line 271 of 3builddistro removes all empty packages-*** directories - not just _DEV ones - would be better if it was confined to _DEV directories as the echo line 270 suggests.

peabee:
@dimkr @01micko 
Why do 2 of my builds behave differently..........???

A slackware based build only builds icons once 
however

An ubuntu based build
- keeps rebuilding icons every build run with multiple copies in petbuild-output (see below)
- has an empty petbuild-cache directory
- does not have ccache in sandbox3

```
# pwd
petbuild-output
# ls
pmaterial_icons-308c647ce29eaf8424375978b9db09f3
pmaterial_icons-6d1fbe5d153313894b7cb3c5304742ba
pmaterial_icons-a6edb1eee2d163e83b3486408ea9b110
pmaterial_icons-b69fb4d433bb5da1cfd42a0177ea8293
pmaterial_icons-db641907e25b40a909dec5d7721f8eff
pmaterial_icons-fecbade149eae76c6d51930daccc8c5e
pmaterial_icons-latest
puppy_flat_icons-335c4232d0a4911571d35d9e822728c3
puppy_flat_icons-39d6bffd8516efeb9f95d7675f3d78fc
puppy_flat_icons-9b7d7a3880cc84e3b5fe698f92964d5f
puppy_flat_icons-9f90c285719224876dd31973fcc76859
puppy_flat_icons-bdca06ca3e8aa5dd9b89df094466d83a
puppy_flat_icons-c8a89d1a2fff96001dfb833fb298abd4
puppy_flat_icons-latest
puppy_standard_icons-1ee78c31dc8a783baebb70ed565b86a9
puppy_standard_icons-1ff9599fe2ebb8589050b6f9ee93518c
puppy_standard_icons-d6011de8f6c54065bc6b5d14480b6c21
puppy_standard_icons-e81af65f43f95c246642065406bde7b3
puppy_standard_icons-f2ef454059b9dac7807d795edc9121b9
puppy_standard_icons-fa8866b842b1aa02de56c2e63beab197
puppy_standard_icons-latest
# 

```

peabee:
Further.........
the slackware devx does not have /usr/bin/ccache

but the ubuntu devx DOES from:
yes|ccache|ccache|exe>dev,dev,doc>dev|

dimkr:
petbuilds.sh builds a native ccache if it's missing (for backward compatibility with bionic, and faster ARM cross-compilation). It's OK not to have ccache in devx.

petbuild-cache will be empty if you only build icon packages, because this directory is where ccache (a caching wrapper for GCC) stores its cache, and you don't compile anything. Download sources are stored under petbuild-sources.

The multiple directories under petbuild-output/ and repeated builds can happen if DISTRO_PKGS_SPECS* changes between runs, because
```
    HASH=`cat ../DISTRO_PKGS_SPECS-${DISTRO_BINARY_COMPAT}-${DISTRO_COMPAT_VERSION} ../DISTRO_COMPAT_REPOS ../DISTRO_COMPAT_REPOS-${DISTRO_BINARY_COMPAT}-${DISTRO_COMPAT_VERSION} ../DISTRO_PET_REPOS ../rootfs-petbuilds/${NAME}/petbuild 2>/dev/null | md5sum | awk '{print $1}'`
```

This is how petbuilds.sh decides whether or not it should rebuild.

peabee:
Still unclear why the 2 build types differ so much......

Slackware build does not have ccache in devx and only builds icons once over many rebuilds.....

Ubuntu build does have a ccache in devx and seems to rebuild icons every system rebuild...... slowing things down and eating up disk space (petbuild-output should be cleaned up??)

The rate of change of DISTRO_PKGS_SPECS* for both build types is about the same.........

dimkr:
> Still unclear why the 2 build types differ so much......

Feel free to calculate the value of HASH before every time you run 3builddistro. If it changes, it means one of the file has changed, and that's what triggers the build (for a good reason). If it hasn't changed and there's a rebuild, this is a bug. If your petbuild-output directory has multiple build outputs per package, this means something has changed.

> Slackware build does not have ccache in devx and only builds icons once over many rebuilds.....

Icons are not related to ccache. ccache is only used during builds that invoke GCC, and the icon packages don't do that.

peabee:
By doing 2 consecutive 3builddistro with no file changes I have confirmed that a petbuild does not then occur.
I have raised pull request #2496 to tidy up older petbuilds when a petbuild does occur.

PR #2409, opened 2021-07-27
For consistency with X.Org, run Xwayland as :0, not :1

dimkr:
Like other compositors, Cage uses "lazy" start of Xwayland: DISPLAY=:0 is set, and Xwayland is started when a legacy X11 application is started.

But we don't need that in our particular use case of Cage. We want to run Xwayland right from the start and want this instance to use :0, because that's the usual value of DISPLAY, and some scripts rely on this.


01micko (commented):
Ok, without testing seems to make sense to me, but it is late at night here. :sleepy:

PR #2410, opened 2021-07-27
Set PETBUILD_GTK=3 even if not building anything

dimkr:
Otherwise, in some cases (i.e. all packages before leafpad are cached), you get both l3afpad and leafpad.


01micko:
Ah yes, was meaning to report this!

Issue #2411, opened 2021-07-27
Battery applet fails if charge_full is missing but energy_full exists


dimkr:
Happens on my Thinkpad X220 with a non-genuine battery and kernel 5.10.x.

PR #2412, opened 2021-07-27
Fix immediate exit of powerapplet_tray on Thinkpad X220 (#2411)

dimkr:


01micko:
Working fine here on my Dell Latitude E7270

01micko:
Working fine here on my Dell Latitude E7270

PR #2416, opened 2021-08-01
bugfix for new line and missing field

01micko:
- formerly looked like this:

:xlockmore:|pet|
xlockmore-5.66|xlockmore|5.66||BuildingBlock|256||xlockmore-5.66.pet||Lock the screen|puppy|||

- now

:xlockmore:|pet|local|xlockmore-5.66|xlockmore|5.66||BuildingBlock|256||xlockmore-5.66.pet||Lock the screen|puppy|||

Since not from a repo I put 'local'

I did similar with 894a62e

PR #2417, opened 2021-08-01
Update petbuilds.sh

01micko:
No need for this. `rootfs-packages` doesn't record stuff here.

The old version was a bug, but was missed because syntax was wrong.

re: #2416


01micko:
@dimkr - I might be offline this week so do as you see fit - moving house :house:

dimkr:
> @dimkr - I might be offline this week so do as you see fit - moving house 🏠

Good luck!

I'm preparing for my wedding, so maybe I won't have much time for woof-CE in the upcoming weeks 🤵🏿

dimkr:
The purpose of this line (and the copying of petbuilds to packages-*) is making woof-code/support/builtin_files.sh capable of listing petbuild files in builtin_files. This doesn't work, but I still haven't found the time to investigate this.

01micko:
> The purpose of this line (and the copying of petbuilds to packages-*) is making woof-code/support/builtin_files.sh capable of listing petbuild files in builtin_files. This doesn't work, but I still haven't found the time to investigate this.

This is from a build a month ago `/root/.packages/builtin_files/gtkdialog` when the syntax was broken

```
/usr/sbin/gtk2dialog
/usr/sbin/gtkdialog
/usr/share/icons/hicolor/32x32/apps/gtkdialog.png
```

So I'd say it would still work if the entries aren't there. EDIT: no

01micko:
Ok, this is weird; there needs to be something in 
`status/findpkgs_FINAL_PKGS-${DISTRO_BINARY_COMPAT}-${DISTRO_COMPAT_VERSION}`
even if it is broken, so it's a woof bug. However `rootfs-packages` aren't recorded there but are there in `/root/.packages/builtin_files/`

I'm off to work soon so if you don't figure it out by then I'll fix it this afternoon.

dimkr:
/root/.packages/builtin_files/jwm looks sane without this change and it's missing with it:

```
/etc/system.jwmrc
/usr/bin/jwm
/usr/share/jwm/jwm-blue.svg
/usr/share/jwm/jwm-gray.svg
/usr/share/jwm/jwm-orange.svg
/usr/share/jwm/jwm.xbm
```

dimkr:
> Ok, this is weird; there needs to be something in

Look at woof-code/support/builtin_files.sh.

This loop handles what's in findpkgs_FINAL_* (2createpackages + petbuilds):
```
cat ../status/findpkgs_FINAL_PKGS-${DISTRO_BINARY_COMPAT}-${DISTRO_COMPAT_VERSION} | sort | \
while IFS="|" read -r F1 F2 F3 F4 F5 ETC
do
```

And this loop handles packages from rootfs-packages:
```
# do the same for rootfs-packages
if [ -f /tmp/rootfs-packages.specs ];then
        while read line
```

01micko:
I drafted this pending an outcome.

01micko:
The solution is probably to run petbuilds and icons through the same routine as rootfs-packages.

01micko:
I think I'm onto this - make the rootfs-packages routines functions in `support/builtin_files` and `support/findwoofinstalledpkgs`, echo the pet.specs files from petbuilds and icons to /tmp like rootfs-packages and process in mentioned files. No need to put stuff in `status/findpkgs_FINAL_PKGS*`

01micko:
Ugh, tried to use git gui and had a foobar.

PR #2419, opened 2021-08-01
Add scheduled builds of bionic64 to protect it against breakage

dimkr:
Looks like people still like it in 2021, so let's do some super minimal effort to keep it working: without petbuilds, without updates from rootfs-packages, and without the cleanup it needs (zz_bionicfix64, I'm looking at you).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,6 +45,10 @@ jobs:
             compat-distro: slackware
             compat-distro-version: "15.0"
             kernel: 5.4.x-x86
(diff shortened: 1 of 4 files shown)


dimkr:
It's weird that bionic64 eats up so much RAM, but doesn't have PulseAudio, Blueman, GTK+ 3 and other goodies in dpup and slacko64 🤔 

bionic
dpup
dpup2
slacko64

01micko:
This is a good idea I think.

Issue #2418, opened 2021-08-01
petbuilds busybox fails


mrfricks:
when running woofce locally petbuilds fail on busybox.  
this is just a default build for fossapup no changes
```
busybox-1.33.1.tar.bz2: OK
busybox-guess_fstype.patch: OK
Building busybox
mount-FULL: /root/woof-out_x86_64_x86_64_ubuntu_focal64/sandbox3/petbuild-rootfs-complete-busybox: wrong fs type, bad option, bad superblock on petbuild, missing codepage or helper program, or other error.
chroot: can't execute 'sh': No such file or directory
umount-FULL: petbuild-rootfs-complete-busybox: not mounted.
rmdir: failed to remove 'petbuild-rootfs-complete-busybox
```

dimkr:
You're probably using an old kernel ~without support for overlayfs~. What does `dmesg` say?

mrfricks:
hi @dimkr 
dmesg error ...
```
unsupported filesystem, ../petbuild-output/busybox-68f0559094b3c3046bb31fa89c773fc2 (aufs)
```
my kernel config 
```
CONFIG_OVERLAY_FS=m
# CONFIG_OVERLAY_FS_REDIRECT_DIR is not set
# CONFIG_OVERLAY_FS_REDIRECT_ALWAYS_FOLLOW is not set
# CONFIG_OVERLAY_FS_INDEX is not set
# CONFIG_OVERLAY_FS_XINO_AUTO is not set
# CONFIG_OVERLAY_FS_METACOPY is not set
```

dimkr:
aufs is unsupported ... what distro are you running woof-CE on?

mrfricks:
just a regular fossapup.  
would it work on the github produced iso?

dimkr:
@mrfricks Yes, it should, but it also should work on Puppy.

You can grab an ISO at https://github.com/puppylinux-woof-CE/woof-CE/actions/runs/1086152541#artifacts.

peabee:
Why does a pet build system depend on an overlay file system.?........?
Aufs is the puppy standard...?
If there are kernel specific requirements they should be checked?

dimkr:
> Why does a pet build system depend on an overlay file system.?........?

Because it's useful. By using aufs to merge devx+rootfs-complete+a temporary directory, we can build packages in a clean environment without having to copy devx onto rootfs-complete for every build.

> Aufs is the puppy standard...?

[...]

PR #2426, opened 2021-08-02
Let the user select the petbuilds to use

dimkr:
This PR makes `PETBUILDS` a list of packages to build, instead of an on/off toggle that selects all packages not included already.

The list can be tuned later (as requested by @mrfricks in https://github.com/puppylinux-woof-CE/woof-CE/issues/2418#issuecomment-890919569); for now, I set PETBUILDS= to the list of packages that previously got auto-selected.

### DIFF
--- a/woof-code/3builddistro
+++ b/woof-code/3builddistro
@@ -254,7 +254,7 @@ case $BUILD_DEVX in #_00build.conf
 	*) BUILD_DEVX=yes ;;
 esac
 
-if [ "$BUILD_DEVX" = "yes" -o "$PETBUILDS" = "yes" ] ; then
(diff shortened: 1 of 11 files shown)


dimkr:
@01micko With this PR, icon themes can become petbuilds, because @peabee can enable just these 3 petbuilds and keep using PET packages for everything else. What do you say?

01micko:
> @01micko With this PR, icon themes can become petbuilds, because @peabee can enable just these 3 petbuilds and keep using PET packages for everything else. What do you say?

Yes. :smiley:

01micko (approved):
I like it!

PR #2435, opened 2021-08-06
RAM saver

dimkr:
This is a shared library that reduces RAM consumption by tuning malloc() to waste less RAM at the cost of inefficiency and slower operation, using mallopt().

The four applications I'd like to make lighter are Xwayland, Cage, Python (Blueman) and Firefox, and I'm not sure if this solution should be more targeted (for example, `prctl(PR_GET_NAME)` and call mallopt() only if this process is named `python3`, `Xwayland`, `cage`, `firefox` or `firefox-esr`).

Before (x86_64, Xwayland, GTK+ 3):

Screenshot 2021-08-06 10 38 40

After (x86_64, Xwayland, GTK+ 3):

Screenshot 2021-08-06 10 27 20

Before (x86, X.Org, GTK+ 2):

before

After (x86, X.Org, GTK+ 2):

after

On real hardware (this is under QEMU), I believe the effect will be stronger, because this VM doesn't have a Bluetooth adapter (hence, Blueman does less), etc'.

EDIT: RAM usage of 450 MB on x86_64 with Xwayland, Cage, Blueman and Firefox, compared to ~490 MB without this PR.

EDIT 2: same test, about 10% reduction in RAM usage on my Thinkpad X220. Overall, it performs very nicely, and Firefox still feels fast.


01micko:
Looks interesting..

I'm still a bit out of the loop with the house move and all but I'll review soon.

01micko:
In your opinion, is this useful for the other woof-distros apart from dpup?

dimkr:
> In your opinion, is this useful for the other woof-distros apart from dpup?

Yes, of course. It's a RAM vs. CPU tradeoff, so it's useful for any Puppy where lower RAM usage is important.

01micko:
Well add Slacko to the list!

dimkr:
> Well add Slacko to the list!

Building feature/ram-saver-slacko (this branch, with ram-saver added to 8.x) now at https://github.com/dimkr/woof-CE/runs/3532896117?check_suite_focus=true.

dimkr:
8.x already has super low RAM usage, and now it's even lower :)

before

slacko

(The savings will be bigger when Firefox is running)

PR #2445, opened 2021-08-15
Symlink directories at / to /usr

dimkr:
This makes Puppy's directory structure more like Debian's or Ubuntu's, possibly increasing compatibility.

/lib is still a directory and not a symlink, because it's hard to do this in a backwards-compatible way. initrd, fdrv and zdrv replace the /lib -> /usr/lib symlink with a directory, which is the same as deleting /lib, together with libc and everything else inside it.

Therefore, I kept the separate /lib directory, at least until I come up with a way to fix this.


peabee:
Where are the problem reports showing that this "lib structure incompatibility" is causing problems?
What are the implications for existing pets and sfs particularly those large ones that assume the current lib structure? e.g. the ones provided by @mrfricks for FossaPup64
Presumably this only applies to Debian/Ubuntu builds - not to Slackware or other lib architectures such as Arch?

dimkr:
> Where are the problem reports showing that this "lib structure incompatibility" is causing problems?

Define 'problems'. Dozens of Ubuntu and Debian laptops, VM, containers and chroot environments work just fine for me, with the unified directory layout, which is the default.

Problems happen only if you try to replace, say, /bin, with a directory. But if you extract files into /bin (say, install a package), it's transparent (the kernel resolves the /bin symlink when trying to open /bin/x for writing).

Debian, Ubuntu and Fedora survived this transition without breakage, so I see no reason why Puppy can't:

https://lists.ubuntu.com/archives/ubuntu-devel-announce/2018-November/001253.html
https://wiki.debian.org/UsrMerge
https://fedoraproject.org/wiki/Features/UsrMove

EDIT: even Arch survived this, https://archlinux.org/news/binaries-move-to-usrbin-requiring-update-intervention

> What are the implications for existing pets and sfs particularly those large ones that assume the current lib structure? e.g. the ones provided by @mrfricks for FossaPup64

The structure is the same. /lib/libc.so.6 is still there.

However, this change does break compatibility with some SFSs, because a SFS that contains a /lib directory will replace the /lib symlink with a directory containing just the files in the SFS (so /lib/libc.so.6 is gone, etc').

If you look closely - there's an exception for bionic and focal, despite the fact Ubuntu 20.04 already uses the new layout.

> Presumably this only applies to Debian/Ubuntu builds - not to Slackware or other lib architectures such as Arch?

Slackware doesn't use this layout, as far as I know, and I haven't checked if Arch does (I assume it does, will be handled in a separate PR).

peabee:
It seems to me to be an inherently dangerous structure for the reasons you have described - i.e. the symlinks can be overwritten if an incompatible sfs is installed.

I was asking what problems Puppy Linux users had reported when using the current "real directory" structure........not what problems you had encountered with the new structure.

dimkr:
Don't forget the potential problem of incompatibility with some package, which assume /lib contains the same files as /usr/lib, etc'. This is the problem I'm trying to solve here.

dimkr:
https://github.com/dimkr/woof-CE/releases/tag/dpup-8.0-usrmerge1 is built using this branch.

01micko:
Slackware has not merged /usr. There is obviously some resistance in the slacker community to this: search "slackware /usr merge" and a few hits pop up on LQ. I can say with some certainty that it won't happen before Slackware 15.0 is released as it is too close now.

Another problem that slackware packaging creates is that historically, slackware packages do not ship with symlinks; they are created in the `doinst.sh`; slackware's post install script. So slackware packagers would have to change all the slackbuild scripts to exclude `lib, bin, sbin` as install paths. Probably would be met with resistance too.

In saying that, slackware is a benevolent dictatorship led by Pat and the community usually accepts big changes that he makes in the end. So perhaps post 15 may see slackware adopt /usr merge. We'll see.

dimkr:
> Slackware has not merged /usr.

Slacko is not affected by this change, but if Slackware adopts the /usr merge, the condition can change to `debian, raspbian, ...  or ubuntu != (bionic, focal) or slackware > 15.0`.

peabee:
Removing the small utility change_kernels does not stop people changing kernels by the manual method (which many prefer anyway).........

dimkr:
> Removing the small utility change_kernels does not stop people changing kernels by the manual method (which many prefer anyway).........

It doesn't stop people from deleting /sbin/init or /etc/rc.d/rc.sysinit, either. But if this tool is guaranteed to break your system, I prefer to remove it until we have a better solution.

peabee:
This is a major structural change with profound implications**
If it is to go ahead then it should be an explicit configuration option in 00build.conf rather than be obscurely buried in the code of 3builddistro
It should be possible to still build all systems with the old "real directories" structure
There should be some way for users to be alerted that they are running a system that cannot use some existing sfs as they are now incompatible - maybe a specific readme in the .iso?

**
it makes all existing fdrv obsolete because they no longer load
it make all existing huge-kernels obsolete because they crash
etc.

dimkr:
Works nicely in https://github.com/dpupos/woof-CE/releases/tag/dpupos-x86_64-1-rolling120

Screenshot 2021-08-18 18 54 25

01micko:
I'm ok with this PR as it is 'opt in'. Shouldn't affect anything really, apart from old sfs files.

BTW, `change_kernels` script is ~hidden~ not well publicised  for a reason :wink:

bigpup1:
If change_kernels script has a problem.
Someone needs to fix it to correctly work.
More people use it than you may know about.
I use it!
It keeps from making naming errors when changing a kernel.
Yes, there are times when changing the kernel fixes problems.
what a 3 series, 4 series, 5 series kernel supports for hardware, is not the same.
I have fixed hardware issues, simply by changing the kernel, to one that better supports the hardware.
Trying to help a new Puppy user, is easy, if you tell them to use change_kernels script, to do the kernel change.

dimkr:
This procedure can be automated with a script, but the freedom to use any kernel means a proper Puppy release needs to be tested against a variety of kernels, and mention what was tested in the release notes. Also, the script must have extra protection measures to ensure it doesn't break a Puppy installation, and that's not trivial.

Personally, I think this tool is mostly useful to tinkerers willing to swap kernels, but if it's so easy to do this manually and tinkerers are ... well, tinkerers, then this script doesn't help much, and it can easily fool inexperienced users into breaking their installation. I liked the old idea of a "retro" edition that uses an older kernel, so people don't have to swap kernels themselves, and don't have to guess what kernels are known to work or what kernel to go for.

I can add protection against kernel SFSs with /lib instead of /usr/lib or the other way around, but I'm not a big fan of the kernel swapping thing in first place, and kernel-kit needs to support building SFSs that use /usr, before I do this.

(And, I think kernel-kit issues unrelated to the /usr merge should be handled on a "one issue per GitHub issue #" basis, and not in this PR.)

Butterfly:
> Don't forget the potential problem of incompatibility with some package, which assume /lib contains the same files as /usr/lib, etc'. This is the problem I'm trying to solve here.

Hey. Just poking my nose in as a thought occurred. It's been a while since I tried to install packages, but I remember a couple times installing dot debs. If that ability continues, that's one place where following the evolving symlink change might help. Everything's compatible for now, but maybe one day they'll decide to eliminate those top tier, e.g. /lib and /lib32, directories entirely. That seems slightly foreseeable in maybe five or ten years because I've witnessed other hierarchy changes that did cause developer distress in years past.

Devil's Advocate on this is that it makes those /usr directories so big that it takes "a while" for some to finally load to where you can actually view them. It hasn't happened in a couple years, but that used to lock up my computer to where I had to hit the hardware button to reboot. For me, that's on a debootstrap'ed copy of Debian which is very minimal in size. I can only begin to imagine how sluggish things might be for an installation via Debian Developer created ISO file. :laughing:

Butterfly:
> EDIT: even Arch survived this, https://archlinux.org/news/binaries-move-to-usrbin-requiring-update-intervention
> 
> > What are the implications for existing pets and sfs particularly those large ones that assume the current lib structure? e.g. the ones provided by @mrfricks for FossaPup64
> 
> The structure is the same. /lib/libc.so.6 is still there.
> 
> However, this change does break compatibility with some SFSs, because a SFS that contains a /lib directory will replace the /lib symlink with a directory containing just the files in the SFS (so /lib/libc.so.6 is gone, etc').

If you can ever find something that follows a "rote" (a predictable), boring, time consuming project that you keep putting off while you do the brain work, I'm here on the side. I'd like to do *something* to help pay back those times that 5.5 and 5.6 were lifesavers for my being able to access the Internet when nothing else worked on very old laptops. :partying_face:

dimkr:
> Devil's Advocate on this is that it makes those /usr directories so big that it takes "a while" for some to finally load to where you can actually view them.

Do you mean, when you open the directory with a graphical file manager?

dimkr:
Closing this PR until bionic is dead. The work needed to stay backward compatible with it is super expensive, and these changes are merged into my fork, so the opposition to this PR is not a problem.

PR #2455, opened 2021-08-20
Add squashfs-tools petbuild

dimkr:
This allows us to upgrade squashfs-tools to a newer version.

I want to use the new "actions" feature in squashfs-tools 4.5 to achieve a better compression ratio for the main SFS, in a separate PR.


dimkr:
Closing this until I find a way to improve the compression ratio. The most "compressible" files in dpup's rootfs-complete are larger than the block size, so they get compressed independently.

PR #2457, opened 2021-08-21
Make 5.10.x kernels very similar to Debian Bullseye kernels

dimkr:
This makes them bigger and increases RAM consumption, but makes Puppy more likely to work where Debian does.

The difference between Debian kernels and these configurations are:
- The kernel is optimized for size, not performance
- File system drivers and block device drivers are built-in
- Support for AppArmor, SELinux, etc' is disabled
- No debugging info, which increases the package size to ~1 GB
- Automatic use of "huge" memory pages is disabled, to reduce RAM usage

TODO:
- [x] Test USB boot
- [x] Test Bluetooth audio
- [x] Make sure the x86 build boots


dimkr:
Tested the x86_64 kernel, boots nicely (both BIOS and UEFI), Bluetooth audio works.

(`neofetch` definitely gets some things wrong)

screeny

dimkr:
> Many of these MMC modules need to be builtin to boot from mmc card, especially that one for my Dell lappy.

I took the Debian list of drivers, and changed =m to =y wherever possible. Does Debian boot on this laptop? 🤔 

EDIT: it probably does, if the initramfs takes care of loading that module

EDIT: changed to =y

dimkr:
@01micko Objections?

01micko:
> @01micko Objections?

Nope.

But @gyrog should have a look at the DOTconfig to make sure certain keyboard drivers are builtins.

01micko:
@gyrog ??

dimkr:
Let's get this one merged, extra keyboard drivers can be added later if needed.

PR #2467, opened 2021-08-28
Attempt to fix shutdown bugs

01micko:
- fix the killzombies() function
- restore missing variable ABSPUPHOME
- kill some processes that may not be dead; eg pulseaudio, dbus, dhcpcd
- use ash instead of bash

Note that are still shutdown bugs especially toward the end of the script when stuff is getting un-mounted


woof-code/rootfs-skeleton/etc/rc.d/rc.shutdown
```diff
@@ -130,10 +136,26 @@ for i in /sys/class/net/* ; do
...
+	DPID=$(cat /var/run/dbus/dbus.pid)
+	kill $DPID >/dev/null 2>&1
+	rm -f /var/run/dbus/dbus.pid
+fi
+killall dbus-daemon >/dev/null 2>&1
+# udev
+udevadm control --exit
+killall udevd >/dev/null 2>&1
+# pulse
+pidof pulseaudio >/dev/null 2>&1 && killall pulseaudio >/dev/null 2>&1
```
dimkr: This `pidof` finds the PID of the process called `pulseaudio`, then `killall` does exactly the same - if errors are silenced anyway, `killall pulseaudio >/dev/null 2>&1` is enough

woof-code/rootfs-skeleton/etc/rc.d/rc.shutdown
```diff
@@ -130,10 +136,26 @@ for i in /sys/class/net/* ; do
 	fi
 done
 
-#terminate dbus daemon
-killall dbus-daemon 2>/dev/null
+#=======================================================================
 
-killall udevd > /dev/null 2>&1
+# terminate stuff
+# dbus daemon
```
dimkr, line 145: There's a second D-Bus running as spot - killing `dbus-launch` should kill it

woof-code/rootfs-skeleton/etc/rc.d/rc.shutdown
01micko: Ok, so no need to check if it's running.

woof-code/rootfs-skeleton/etc/rc.d/rc.shutdown
```diff
@@ -393,13 +414,13 @@ if [ "$SAVE_LAYER" ];then
 	else
 	   
 	   #Unload save folder gracefully
-	
+	   
 	   if [ -L /initrd${SAVE_LAYER} ]; then
    
 	    rm /initrd${SAVE_LAYER}
```
01micko: This fails because we made the fs `ro` a few lines above.

woof-code/rootfs-skeleton/etc/rc.d/rc.shutdown
```diff
@@ -130,10 +136,26 @@ for i in /sys/class/net/* ; do
...
 
-killall udevd > /dev/null 2>&1
+# terminate stuff
+# dbus daemon
+if [ -e /var/run/dbus/dbus.pid ]; then
+	DPID=$(cat /var/run/dbus/dbus.pid)
+	kill $DPID >/dev/null 2>&1
+	rm -f /var/run/dbus/dbus.pid
+fi
+killall dbus-daemon >/dev/null 2>&1
```
01micko, line 155: So after this line..
```
killall dbus-launch
```
@dimkr :question: 
I'm aware there is some redundancy in this block but if we can't kill it with an axe maybe the sledge hammer works.

dimkr: Yes, I think this is good enough.

gyrog:
Don't bother trying to umount anything that contains any branch of the aufs stack, while it is still "/".
Unless "/" is changed to be something other than the aufs stack, the umount will fail because the "device" is "busy", the best you can try for is to remount stuff as "ro".

In the past some folk have tried to "fix" rc.shutdown, by adding lots's of umount's.

The symbolic link is purely for our benefit, so we can find the savefolder directory via '/initrd' in a running Puppy, "aufs" uses only the "real" path of the savefolder.
Removing it is probably an attempt to cleanup before the next boot.
In the past there might have been problems in 'init' when someone tried to change from a savefolder to a savefile.
It should be safe to either remove it, or move it to earlier in 'rc.shutdown'.

01micko:
The way Fatdog does it is probably the way to go, with caveats:
- fatdog uses a 'huge' initrd and does a 'pivot_root' back to initramfs to umount the aufs stack because all the utilities seem to be available
- is it possible to 'pivot_root' and `busybox umount` the aufs stack?

01micko:
@gyrog 
What lead me to chase these bugs is my laptop (Dell Latitude E7270)  that I regularly boot from a frugal pup formatted and installed `f2fs` sdcard.

Certainly I was getting errors on my main box but these hacks in this PR have alleviated that and `fsck` reports the ext4 boot partition as 'clean'.

I hacked `init` to fsck the f2fs partition every boot and that occurs every boot due to shutdown failures. 

`fsck.f2fs` (standard in initrd.gz for a while now) is quite descriptive, and takes at least 90 seconds on a core i5 with 8GB RAM. It outputs the time when done and has been up to 100 seconds.

At first I was getting a 'bad shutdown' error (sorry don't remember exactly) but since I've hacked `rc.shutdown` the errors now are 'failed unmount' (or words to that effect).

woof-code/rootfs-skeleton/etc/rc.d/rc.shutdown
```diff
@@ -130,10 +136,27 @@ for i in /sys/class/net/* ; do
...
 done
 
-#terminate dbus daemon
-killall dbus-daemon 2>/dev/null
+#=======================================================================
 
-killall udevd > /dev/null 2>&1
+# terminate stuff
+# dbus daemon
+if [ -e /var/run/dbus/dbus.pid ]; then
```
dimkr: dpup still uses /var/run/dbus/pid and still suffers from the issue of D-Bus that refuses to start if it exists, despite https://github.com/puppylinux-woof-CE/woof-CE/commit/76a6e9a660b7e624d0a89406fafebab734c1b227#diff-c75f6042f552eb9e4ab0a07074a889c068ba6f3c9f027b429b075758ae456bbfR23 :(

01micko: what about a loop?

``` sh
if [ -d /var/run/dbus/ ]; then
	for p in  /var/run/dbus/*pid 
	do
		DPID=$(cat $p)
		echo killing $DPID
		kill $DPID >/dev/null 2>&1
		rm -f $p
	done
fi
```
## dot removed

dimkr: Maybe `/var/run/dbus/*pid`, without the dot.

I just reverted that commit, because Debian's /etc/init.d/dbus deletes the pidfile if the PID in it doesn't belong to a running process. If the problem persists, a loop like this (or a simple `rm /var/run/dbus/*` in rc.sysinit) can fix the issue.

01micko: Ok, I'll add the loop but for a failsafe, say power failure, it should be added to rc.sysinit as well.

01micko:
Although this PR isn't feature complete, I believe most bugs are fixed. Plan to merge very soon.

01micko:
Merging in 1 hour

PR #2468, opened 2021-08-28
Make 431 less green and remove BlueGreen.svg

dimkr:


dimkr:
Definitely looks better, merged.

PR #2472, opened 2021-08-29
Offer ext4 without a journal, not ext3, as an alternative to ext4

dimkr:
ext4 should be faster, and a better file system overall. There's no reason to choose ext3 over ext4, because Puppy supports ext4 since 4.3. However, there is a case for ext4 without journaling - it may be a more flash memory friendly option.


01micko (approved):
Neat idea.

bigpup1:
> ext4 should be faster, and a better file system overall.

Sure about that?

Bugs ext3
https://bugzilla.kernel.org/buglist.cgi?bug_status=NEW&bug_status=REOPENED&bug_status=ASSIGNED&component=ext3&product=File%20System
Bugs ext4
https://bugzilla.kernel.org/buglist.cgi?bug_status=NEW&bug_status=REOPENED&bug_status=ASSIGNED&component=ext4&product=File%20System

dimkr:
> Sure about that?

This is a bit misleading. The ext4 driver can be used for ext2 and ext3. Where do these bugs come from - the ext3 driver, or the ext4 one (i.e. an ext3 file system mounted by the ext4 driver)? Also, if nobody is using ext3 for a decade, new bugs won't crop up in the ext3 list, so it's not surprising to see a shorter list of bugs for ext3.

PR #2474, opened 2021-08-29
Add a scalable wallpaper reminiscent of in_flight1600x1200.jpg

dimkr:
in_flight1600x1200

in_flight


01micko:
svg is fun!

dimkr:
> svg is fun!

But it can be so ugly! 😊

01micko:
That last effort isn't so bad.

PR #2477, opened 2021-08-30
Scale 412.svg to 4K 16:9

dimkr:


01micko:
I'm merging this but I am also working on an alternate solution for SVG wallpapers that will (hopefully) end silly resolution problems.

PR #2481, opened 2021-09-02
Replace eudev petbuild in dpup with systemd udev 

dimkr:
focal64 used an eudev .pet before I converted it to use PETBUILDS, so dpup is the only user of the eudev petbuild. Because Debian's udev package can be used without systemd, we don't really need eudev in dpup. Let's get rid of it now and deal with potential issues as early as possible, even if eudev gets adopted by a new maintainer.

https://github.com/gentoo/eudev/issues/199

udev


peabee:
FYI....
a collaboration between Alpine, Devuan and Gentoo contributors has announced their adoption of eudev and a new repository has been created where the new project will further develop eudev: https://github.com/eudev-project/eudev/blob/master/README.md .

PR #2483, opened 2021-09-03
Confine deletes to empty _DEV directories only

peabee:


dimkr:
What problem does this fix?

peabee:
Directories should not be deleted unnecessarily..... 
Deleting directories which have been created earlier in the build process just to suppress error messages in 3builddistro seems to be logically unsound particularly when more directories are deleted than is required to suppress the messages.
Specifications such as:
yes|gvfs_lib_check|gcr,libsecret|exe>null,dev>null,doc>null,nls>null
are valid but deleting all empty directories rather than just _DEV means other error messages are generated.
What is wrong with making the delete specific to just empty _DEV directories? (in the absence of better logic to deal with the error messages)?

woof-code/3builddistro
```diff
@@ -268,7 +268,7 @@ if [ "$BUILD_DEVX" = "yes" -o -n "$PETBUILDS" ] ; then
 	##rm -f ./${WOOF_OUTPUT}/${DEVXSFS} 2>/dev/null #100911
 	rm -f ./sandbox3/${DEVXSFS} 2>/dev/null #100911
 	echo "removing empty _DEV directories"
-	find packages-${DISTRO_FILE_PREFIX} -maxdepth 1 -type d -empty -delete -print # suppresses warnings in the following
+	find packages-${DISTRO_FILE_PREFIX} -maxdepth 1 -type d -name _DEV -empty -delete -print # suppresses warnings in the following
```
dimkr: I think you meant `-name "*_DEV"`

dimkr (approved):

PR #2484, opened 2021-09-04
Assorted assistance to newbies.

JamesC1:
Test in /merge2out to resist spinning on bad parameters.
Test-case distribution for the fix in /merge2out.
Documentation enhancement in /README.md.
--------------
Slackware .iso boots EFI on a MacBookPro8,1 with an ASUS external drive, but not CSM, the same as the official distribution.
Starts the GUI.  Does not find the wireless card, so the browser test is limited to the splash page.


README.md
```diff
@@ -11,12 +11,15 @@ Currently supported: [![Build status](https://github.com/puppylinux-woof-CE/woof
 Puppy can be built directly on GitHub, saving the hassle of preparing a suitable build environment on a fast machine with adequate storage.
 
 0. Fork woof-CE.
+    1.  If you want something that can cleanly make a pull request, create a branch.
+    1.  If your changes include automation (.yml, .yaml) files, set the default branch to this new branch.  (NB: this will confuse repositories that fork you.)
```
dimkr, line 15: 2?

merge2out
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash
 #BK nov. 2011
 
+# spun for 6 hours in git actions, asking for an architecture
+# https://github.com/JamesC1/woof-CE/runs/3454321862?check_suite_focus=true
+
```
dimkr, line 6: A much simpler fix would be:

```
if [ ! -d "$1" ]; then
    echo "$1 does not exist"
    exit 1
fi
```

in line 32

woof-distro/broken/README.md
```diff
@@ -0,0 +1,12 @@
+# broken - the broken distribution for testing purposes
```
dimkr, line 1: What's the purpose of this directory?

README.md
dimkr: Why change the default branch? The "release" workflow can run on any branch

JamesC1:
Hello Dima,

Could you please check whether this is a good attempt at the simpler
fix in your email below:
https://github.com/JamesC1/woof-CE/commit/c0adc8e12eab6bf16c25b5e61177ee713689a920

Please find an action run where, with that change and the carefully
crafted broken distribution, I can still make it spin enough to
trigger my other change:
https://github.com/JamesC1/woof-CE/runs/3582354815?check_suite_focus=true

I also dislike the change above because of an interaction with this
other change to merge2out:
https://github.com/JamesC1/woof-CE/commit/95299bb6a964aa41510c89f5733de8af2c26ad0b

With the earlier change only, merge2out can be run with no arguments,
almost certainly in a shell on a local machine, to interactively begin
to create a new architecture or distribution.  With both changes, it
can no longer be run interactively, because $1="" to satisfy the older
one will fail [-d "$1"] in the newer one.

I would like an agreement to revert the newer change, linked above.
How would you like to proceed?

I plan to go on with other branches of the review in the mean time.

Regards,
James.

JamesC1:
Hello Dima,

I'm getting inconsistent results.

This documentation:
https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow
says "
To trigger the workflow_dispatch event, your workflow must be in the
default branch.
"

When I first added the "Log event inputs" step on branch
woof-safety-01 (to keep it together with the other items in my future
pull request), I had a lot of difficulty with getting it to run.

I interpreted the documentation as meaning that, when release.yml is
on multiple branches, the version that runs is the version on the
default branch, not the version on the branch from which the rest of
the files are taken.

Now, when I set the default branch to testing (without "Log event
inputs") and run with files on woof-safety-01 (with "Log event
inputs") I get event inputs logged.  There is an additional change,
from echoing the inputs to dumping the whole environment, that you
might merge or not, as you please, and it shows in my runs against
woof-safety-01:
https://github.com/JamesC1/woof-CE/runs/3582773591?check_suite_focus=true

Are you sure enough that I was wrong to change branches, so that
future new developers should not be warned about it?

Regards,
James.

JamesC1:
Hello Dima,

I am taking the view that a new developer will attempt to test their
changes from github actions, so the behaviour in github actions is the
behaviour that I am trying to modify and test.

In order to test, I have to check in.

That subtree is test code for (my current understanding of the
situation around) the bigger change to avoid spinning.

I've tested the change to merge2out and I'm happy with it.  You could
merge the test subtree or not.  If you don't, then the next person who
wants to test those lines will need to reinvent it.

Regards,
James.

01micko:
@JamesC1 I'm sure there is still interest in this PR

merge2out
dimkr: merge2out already has this check!

JamesC1:
Hello Dima,

Whether there is already sufficient checking depends on how you view
my test architecture, "broken".

With the test architecture, I can make a fairly recent version of
merge2out trigger one of the checks that I added, which means that I
think that I could make a version without the check spin until GitHub
Actions killed it after six hours.

If you think that the test architecture is a logical mistake, and that
no newbie will ever run merge2out against a broken architecture, then
the check is wasted safety and wasted risk around code-change.

Regards,
James.

dimkr:
@JamesC1 I triggered a build that uses a non-existing distro version, and it doesn't spin forever. This problem should be fixed by https://github.com/puppylinux-woof-CE/woof-CE/commit/95299bb6a964aa41510c89f5733de8af2c26ad0b.

JamesC1:
Hello @Dima,

That fix is in the branch on which I trigger the counting change:

https://github.com/JamesC1/woof-CE/commit/c0adc8e12eab6bf16c25b5e61177ee713689a920#diff-14799092a0b516a48efb9784ed77f88faef7e8a7f81f8c3f39de7f8815934862

https://github.com/JamesC1/woof-CE/commits/woof-safety-01/merge2out


It would be relevant if 'broken' was a non-existent distribution,
rather than a broken one.

Are you making a policy decision to not cope gracefully with
distributions which have
a DISTRO_SPECS that does not define
		TARGETARCH=${DISTRO_TARGETARCH}
		COMPATDISTRO=${DISTRO_BINARY_COMPAT}
		COMPATVERSION=${DISTRO_COMPAT_VERSION}
?

If so, I should revert both the checking and 'broken' distribution.

What would remain are:

- the additional output of build parameters and environment in
release.yml, which I added in order to try to understand what I had
done wrong

- the additional guidance in README.md, of which changing the default
branch is controversial, and if it stays then you want it renumbered

Regards,
James.

PR #2487, opened 2021-09-06
Delete /var/run/dbus/pid on boot, otherwise D-Bus does not start on dpup

dimkr:


01micko:
Should be deleted on shutdown, (note not _is_ but _should_)

https://github.com/puppylinux-woof-CE/woof-CE/blob/5822702bad38d3f4832ff16c70890bf4d4e296ff/woof-code/rootfs-skeleton/etc/rc.d/rc.shutdown#L146

01micko:
By the way, I don't mind redundancy in case of a forced poweroff so maybe your PR has merit, but my above comment does too.

dimkr:
@01micko Fixed - the pidfile is now renamed to dbus.pid on dpup

01micko:
Ok, I'll get to work soon on #2467

PR #2496, opened 2021-09-10
remove older petbuilds of $NAME

peabee:


dimkr (approved):
Also, the line `rm -f ../petbuild-output/${NAME}-latest` can be removed (the link is already deleted).

I approve this PR, but the original idea was allowing you to use one petbuild-output directory for multiple builds (say, Slacko and dpup), without rebuilding every time you move between them. I think it's useful, but not super important.

peabee:
I cannot believe that anybody would try to work like that....... I certainly keep different build types very separate. And I would think the petbuild names would need to be enhanced with the target system name for it to work reliably? as I thought the whole point of petbuilds is that they are built to conform to the target system architecture and state?

dimkr:
> I cannot believe that anybody would try to work like that....... I certainly keep different build types very separate. And I would think the petbuild names would need to be enhanced with the target system name for it to work reliably? as I thought the whole point of petbuilds is that they are built to conform to the target system architecture and state?

Yes, but you can use one directory for both Slacko and dpup, for example, and 3builddistro will always pick the correct, cached petbuild output (based on the hash of DISTRO_PKGS_SPECS, etc').

PR #2501, opened 2021-09-11
Remove change_kernels if not using a "huge" kernel

dimkr:


peabee (changes requested):
I don't feel comfortable with build decisions like this being buried in the code of 3builddistro. System builders should be free to put whatever components they want into their build and not have their decisions overturned by code like this in 3builddistro.
Maybe raise a warning if somebody tries to do something silly - "Are you sure...." but don't overturn without warning.

dimkr:
> I don't feel comfortable with build decisions like this being buried in the code of 3builddistro.

Why default to including change_kernels, or even allow the woof-CE user to add it, if it simply won't work when not using a "huge" kernel? It doesn't work, creates confusion, and makes the menu longer.

Do you oppose this change because you think this tool should be included in every build, or because you resist this change in the name of resistance to changes?

peabee:
I think code in 3builddistro is the wrong place to make this change...
We already have /support/rootfs-packages.conf
and can set:
change_kernels="false"

dimkr:
But why make things harder? 3builddistro can behave as if change_kernels=false automatically, if change_kernels won't work. Why force the user to modify a configuration file, if 3builddistro can just do the right thing?

peabee:
Because configuration files should correctly specify configuration and code buried in 3builddistro should not override the configuration without the system builder making a conscious decision ..............
What / how many builds are we talking about anyway that don't use huge-kernels???

PR #2507, opened 2021-09-17
Put libramsaver.so.1 alongside libc.so.6

dimkr:
Without this, it's impossible to take /usr/lib from 32-bit dpup and load it in 64-bit dpup, to run 32-bit applications like wine or Steam.


dimkr:
steam

PR #2513, opened 2021-09-23
run-as-spot: Added pulseaudio support

rizalmart:
Run pulseaudio as spot when executing a program

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/run-as-spot
+++ b/woof-code/rootfs-skeleton/usr/sbin/run-as-spot
@@ -65,11 +65,36 @@ if [ $(id -u) -eq 0 ]; then
 		export DBUS_SESSION_BUS_PID
 	fi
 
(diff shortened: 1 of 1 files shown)


dimkr:
@rizalmart This is not needed!

See this block in .xinitrc:

```
# start a D-Bus session bus for both root and spot, for PulseAudio and applications that don't work without it
run-as-spot dbus-launch --exit-with-x11 > /tmp/.spot-session-bus

# start PulseAudio over the session bus and use it for both root and spot, so they see the same output devices
if [ -e /usr/bin/pulseaudio ]; then
    run-as-spot sh -c "pulseaudio --kill > /dev/null 2>&1; pulseaudio --start --log-target=syslog"
    export PULSE_SERVER=unix:/tmp/runtime-spot/pulse/native
    export PULSE_COOKIE=/home/spot/.config/pulse/cookie
fi
```
In dpup, pavucontrol, pa-applet and the browser run as spot, and they work.

PR #2514, opened 2021-09-23
Replace SNS with ConnMan in dpup

dimkr:
Much simpler than SNS, and all functionality is provided by the tray icon.

tray

x86 stays with the much lighter SNS.

EDIT, after 9002f1b:

icons

EDIT 2, after eae5f85:

left

right

EDIT 3, after aa5f423: now, both the spinner and the "Available networks" label disappear when the scan is finished


01micko (approved):
Looks interesting.

PR #2516, opened 2021-09-25
Keep tray icon right-click menus after the mouse key is released

dimkr:
https://docs.gtk.org/gtk3/method.Menu.popup.html says:

```
The activate_time parameter is used to conflict-resolve initiation of concurrent requests for mouse/keyboard grab requests. To function properly, this needs to be the timestamp of the user event (such as a mouse click or key press) that caused the initiation of the popup. Only if no such event is available, gtk_get_current_event_time() can be used instead.
```

I found this bug in the Puppy tray applets when I checked if the connman-ui bug fixed in https://github.com/puppylinux-woof-CE/woof-CE/pull/2514/commits/e4cb2c2827538d081b27619b20230e537a429a8f (see grab.patch) happens with them too. It does.


dimkr:
Tested on x86_64 (GTK+ 3, Xwayland and X.Org) and x86 (GTK+ 2, X.Org).

In all combinations, the menu stays after releasing the right mouse button:

menu

PR #2518, opened 2021-09-27
Fix NLS support in 5.10 configs

01micko:
From the README
The 'huge' kernel builds require a number of drivers built into the kernel
image (vmlinuz in x86/x86_64; zImage, uImage or kernel.img in ARM). Things
that must be builtin are filesystems (ext2/3/4, f2fs, NTFS, FAT and others
which you may want to support as a boot partition) , AUFS, NLS support,
keyboard drivers and possibly others.


dimkr:
I left CONFIG_NLS_CODEPAGE_850 and CONFIG_NLS_CODEPAGE_852 the way they are in Debian. Why is Puppy special here?

01micko:
NLS support in the init script I believe.

dimkr:
> NLS support in the init script I believe.

Things work just fine here, and I don't have nls_cp850.ko and nls_cp852.ko loaded. So far I haven't seen any issues.

I don't object this change - these modules are tiny. I'd like to stay as close as possible to Debian's kernel configuration, and 2 extra lines of difference are not much.

dimkr (approved):

PR #2531, opened 2021-09-30
Use XDG_RUNTIME_DIR in powerapplet_tray and run as spot in dpup

dimkr:


01micko:
The reason I chose `/tmp` for the icon is that it doesn't write to disk being mounted on a tmpfs. I have no qualms with writing a tiny file to disk periodically but some users, especially if using flash media, don't like extra [drive writes](https://bkhome.org/news/202006/pmcputemp-modified-to-remove-drive-writes.html).

That's why I gave `/tmp/powerapplet` dir 01777 permissions. If user can't write to it then I can fully can support this PR.

dimkr:
> The reason I chose `/tmp` for the icon is that it doesn't write to disk being mounted on a tmpfs.

It's still under /tmp:

```
~$ echo $XDG_RUNTIME_DIR
/tmp/runtime-root
~$ run-as-spot sh -c 'echo $XDG_RUNTIME_DIR'
/tmp/runtime-spot
```

In other distros, XDG_RUNTIME_DIR is a tmpfs too (under /var/run or /run).

01micko:
Ok, I'm going to use the `glib` misc utility functions in [pmcputemp](https://github.com/01micko/pmcputemp) but a query about `g_build_filename()`

From the [docs](https://people.gnome.org/~ryanl/glib-docs/glib-Miscellaneous-Utility-Functions.html#g-build-filename):



Returns : | a newly-allocated string that must be freed with g_free().
-- | --


I see no `g_free()` call in the code for `powerapplet_tray.c`

dimkr:
If it's only one call, it's OK to leak that memory: powerapplet_tray.c doesn't free everything explicitly. It is a problem if you call g_build_filename() periodically, and RAM usage increases over time.

01micko:
> If it's only one call, it's OK to leak that memory: powerapplet_tray.c doesn't free everything explicitly. It is a problem if you call g_build_filename() periodically, and RAM usage increases over time.

Fair enough.

I did already rewrite pmcputemp, not committed yet, but here is my main() - well the relevant part.

```c
        temp_icon = g_build_filename(g_get_user_runtime_dir(), ICON, NULL);
	configdir = g_build_filename(g_get_user_config_dir(), CONFDIR, NULL);
	mkdir(configdir, 0755);
	conf = g_build_filename(g_get_user_config_dir(), CONFDIR, CONF, NULL);
	
	paint_icon(style); /* needed to kick it off */
	gtk_init(&argc, &argv);
	create_tray_icon();
	Update(NULL); /* needed to kick it off */
	g_timeout_add(interval, Update, NULL); /*update after 'interval' secs (default 5)*/
	gtk_main();
	
	g_free(temp_icon);
	g_free(configdir);
	g_free(conf);
	
	return 0;

```

Issue #2186, opened 2021-03-28
puppy package manager


mrfricks:
Now that there's talk of improving the speed of the PPM @01micko and before work on this starts, i would like to suggest we roll back to version 2.1.2 and have some saner defaults. 

PPM version 2.5 issues......
1. It's taken away the ability to install multiple packages at once and only installs 
 .  the last item clicked on. 
2. It  has some graphical corruption, when installing a grey blocky area appears 
    over the PPM. 
3. It's removed the install progress bar
4. It's removed the 'non verbose' option when updating the databases so that you 
    have to click 'enter' multiple times
5. It's removed 'skip package size check when more than 4gb space available'

As for saner defaults (applies to version .2.1.2)...
1. Skip package size check when more than 4gb space available
2. Do not show terminal with PPM actions.  
3. Use 'non verbose' option when updating the databases.

Any speedup for the PPM would be great! particularly uninstalling stuff as that seems to take twice as long as installing.

BTW to compare the PPM versions try fossapup64 or bionicpup64 v Latest slacko or @peabee's latest

rizalmart:
QuickPup 21.03 brings back the PPM 2.1.x you talked about. However classic install mode was dropped due to complexity.

PR #2532, opened 2021-09-30
Move PETBUILD_GTK to _00build.conf

dimkr:
This will be useful if:
1. @mrfricks decides to use GTK+ 2 to make his next Puppy closer to FossaPup 9.5 (https://forum.puppylinux.com/viewtopic.php?f=40&t=4138)
2. @peabee wants to use GTK+ 2 in a 64-bit build not based on an old distro like Ubuntu 18.04; the automatic PETBUILD_GTK choice logic in petbuilds.sh (what we have today) will pick GTK+ 3
3. @01micko decides to stick with GTK+ 2 for Slacko 8.x, and deal with GTK+ 3 related issues later, possibly in a point release
3.  I want to build a "lite" or "barebones" 64-bit flavor of dpup that uses the lighter GTK+ 2

### DIFF
--- a/woof-code/support/petbuilds.sh
+++ b/woof-code/support/petbuilds.sh
@@ -28,6 +28,7 @@ fi
 
 if [ -z "$PETBUILD_GTK" ]; then
     echo "WARNING: PETBUILD_GTK is empty, this may be a hard error in the future"
+    [ -n "$GITHUB_ACTIONS" ] && exit 1
 
     PETBUILD_GTK=2
     if [ "$DISTRO_TARGETARCH" = "x86" ]; then
--- a/woof-code/support/petbuilds.sh
+++ b/woof-code/support/petbuilds.sh
@@ -26,22 +26,26 @@ if [ "$WOOF_HOSTARCH" = "x86_64" -a "$WOOF_TARGETARCH" = "x86" ]; then
     CHROOT_PFIX=linux32
 fi
 
-PETBUILD_GTK=2
-if [ "$DISTRO_TARGETARCH" = "x86" ]; then
-    echo "Using GTK+ 2 for x86 petbuilds"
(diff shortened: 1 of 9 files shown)


01micko (approved):
Looks good from my perspective.

peabee (approved):

mrfricks:
there's only a few gtkdialog apps that i kept as gtk2 as i found gtk3 problematic with them.
they are quicksetup (i found selecting the timezone a chore), samba simple management (doesn't display properly) and the PPM (i use an older version that works well but gtk3 removes various install & info bars)

everything else is gtk3

dimkr:
> there's only a few gtkdialog apps that i kept as gtk2 as i found gtk3 problematic with them. they are quicksetup (i found selecting the timezone a chore), samba simple management (doesn't display properly) and the PPM (i use an older version that works well but gtk3 removes various install & info bars)

Can you open a PR that makes these problematic apps use gtk2dialog?

01micko:
@mrfricks I forced **pMusic** to use gtk2, perhaps you could do something similar for the gtkdialog apps you need to run with gtk2. (see `rootfs-packages/pmusic`).

mrfricks:
i can do a PR for samba in petbuilds but for quicksetup i actually just used fossapups version as just running it with gtk2dialog doesn't solve the selection list issue.

as for the PPM i use an older version so a PR wouldn't be appropriate ... unless it is a complete roll back to the last good working version :) https://github.com/puppylinux-woof-CE/woof-CE/issues/2186

@01micko yes i have done that with a few apps.  most are working well with gtk3 though and look nicer.

mrfricks (approved):

01micko:
Well let's merge then :smiley:

PR #2541, opened 2021-10-01
Stop xlock on mouse motion, when using Xwayland (#2535)

### LINKED ISSUE #2535: Desktop doesn't wake up on mouse activity (cage)
A key press, at least on slacko, is needed to wake the display after it has timed out after 10 minutes.

dimkr:


01micko:
Just testing this now. .. and works fine. :+1:

PR #2543, opened 2021-10-01
If Terminal=true, wrap the menu entry command with rxvt -e

### LINKED ISSUE #2542: Nothing happens when htop is started using the menu entry
The htop menu entry has `Terminal=true`, but our menu generator ignores this, and tries to run htop outside of a terminal emulator.

dimkr:


01micko:
```desktop
[Desktop Entry]
Encoding=UTF-8
Name=Htop
Icon=htop.png
Comment=System Process Viewer
Exec=/usr/bin/urxvt -e htop
Terminal=false
Type=Application
Categories=Monitor
GenericName=System Process Viewer
```

So I can ditch that ^^ when merged.

01micko:
Hmm.. should we use `urxvt`?

dimkr:
> Hmm.. should we use `urxvt`?

PPM and other things use "rxvt" 🤔

01micko:
Doesn't matter I suppose.

Issue #2554, opened 2021-10-02
Missing support for display configuration when using Xwayland


dimkr:
These two are intentionally grayed out:

![gray](https://user-images.githubusercontent.com/1471149/135714944-62ad811c-7065-4fa0-9dea-8995bbcb2c51.png)

Once #2528 and #2550 are fixed, it should be possible to run wlr-randr in startxwayland, and configure things like resolution, refresh rate, rotation and the scaling factor.

PR #2558, opened 2021-10-03
Add --right-off support to wlr-randr (fixes #2555)

### LINKED ISSUE #2555: wlr-randr doesn't support --right-of

dimkr:


01micko:
I was going to issue a PR on `quicksetup` but I might wait and see if we can get this working well.

dimkr:
> I was going to issue a PR on `quicksetup` but I might wait and see if we can get this working well.

I think the most important issue to fix using wlr-randr is https://github.com/puppylinux-woof-CE/woof-CE/issues/2554.

It's not as easy as it sounds: Cage doesn't have the equivalent of xorg.conf, so something needs to call wlr-randr for every monitor, with its own settings (resolution, color depth, etc'), **after the monitor is detected**. I'm not sure what's the best way to do this, and maybe the only solution is to patch Cage to add a configuration file with monitor settings, so it knows what to do with each monitor once it's enabled.

wlr-randr and xrandr are not 100% compatible, but they should be close enough to fix xrandrshell without to many `if xwayland else xorg`, at least, for the use case of one-off changes that don't persist to the next X session:

````
~$ wlr-randr  | grep -E '*[0-9]x[0-9].*[0-9]\.' 
    1366x768 px, 60.103001 Hz (preferred, current)
~$  xrandr --screen 0  | grep -E '*[0-9]x[0-9].*[0-9]\.' 
   1366x768      59.98*+
~$ wlr-randr  | grep -E '*[0-9]x[0-9].*[0-9]\.' | awk '{ print $1 }'
1366x768
~$ xrandr --screen 0  | grep -E '*[0-9]x[0-9].*[0-9]\.' | awk '{ print $1 }'
1366x768
```

PR #2575, opened 2021-10-13
Update defaults.patch for geany so underscore char appears

### LINKED ISSUE #2574: sometimes geany doesn't render underscores
In #2563 running geany doesn't render underscore chars. I've seen this before in `raspup` and have noticed on other occasions. It is documented in several geany issues, notably  [#1387](https://github.com/geany/geany/issues/1387).

Apparently it isn't actually a geany bug but a font bug, so the developers say, but the OP in that issue seems to be using `Noto Sans 9` , we just use Dejavu Sans by default. It probably works fine with Liberation or Ubuntu fonts so mostly goes unnoticed. It does work fine in vanilla slacko too, but not in all pups built from woof.

01micko:
closes #2574


dimkr:
The petbuild should apply the patch for this to work

01micko:
> The petbuild should apply the patch for this to work

This PR patches `rootfs-petbuilds/geany/defaults.patch`

dimkr (approved):

PR #2579, opened 2021-10-14
Remove SNS only if the ConnMan package is enabled (fixes #2578)

### LINKED ISSUE #2578: SNS is missing from dpup on x86

dimkr:
- [x] Test x86 https://github.com/dimkr/woof-CE/actions/runs/1342539546
- [x] Test x86_64 https://github.com/dimkr/woof-CE/runs/3897645718

### DIFF
--- a/woof-code/3builddistro
+++ b/woof-code/3builddistro
@@ -145,7 +145,7 @@ and edit it to include your customised package list.
 			;;
 		frisbee|network_wizard|pupdial|simple_network_setup)
 			state=true
-			grep -q 'yes|connman|' $PKGSSPECS && state=false
+			[ -d packages-${DISTRO_FILE_PREFIX}/connman ] && state=false
 			;;
 		*)
 			state=true


peabee:
I don't think build specific things like this should be hardcoded into 3builddistro.....
What if somebody wants a build with both sns and connman??
I suspect your "sharing" of DISTRO_PKGS_SPECS- between different different builds is causing these problems.....

dimkr:
> What if somebody wants a build with both sns and connman??

Why would anyone want that? They can't coexist, and keep killing each other's network interfaces and processes.

> I suspect your "sharing" of DISTRO_PKGS_SPECS- between different different builds is causing these problems.....

No, it doesn't! The `grep` check is the problem - a line can begin with `yes|connman` even if it's inside an `if`. This is not the right way to check if ConnMan is present.

01micko:
> I don't think build specific things like this should be hardcoded into 3builddistro..... 

They already are.

Maybe then we should change how `rootfs-packages` are installed then.

dimkr:
And let's not forget that the inclusion of SNS by default was hardcoded into 3builddistro, until ac493ee! You can't object this PR on the grounds of bad defaults or bad non-configurable choices in 3builddistro. Harcoding SNS used to be the default before ac493ee, and this PR fixes a bug introduced in that commit; both changes are steps towards making 3builddistro more configurable (and generally, decoupling Puppy from SNS), not the other way around.

peabee:
I missed https://github.com/puppylinux-woof-CE/woof-CE/commit/ac493ee0dec7f78ef9cd41d808fc6a6669bdc9e9 as the title did not properly indicate its wide significance.

The inclusion of frisbee|network_wizard|pupdial|simple_network_setup in a build is already adequately specified by entries in /support/rootfs-packages.conf and there is no need for code in 3builddistro

dimkr:
> The inclusion of frisbee|network_wizard|pupdial|simple_network_setup in a build is already adequately specified by entries in /support/rootfs-packages.conf and there is no need for code in 3builddistro

There **is** need to change 3builddistro.

Before ac493ee, the default behavior was to include SNS if ConnMan is present. This is a bad idea, because neither works when both are present. ac493ee changes that, and turns off SNS by default if ConnMan is present. Enabling SNS by default even if ConnMan is present, and expecting woof-CE users to know that, is nothing short of leading users into a trap.

01micko (approved):

dimkr:
I'm merging this one. The choices made by 3builddistro if no configuration is specified should be the safe and sane default, not the other way around. Replacing rootfs-packages.conf with a better solution is a problem for another time.

PR #2581, opened 2021-10-15
Allow mtPaint to use Wayland (fixes #2580) 

### LINKED ISSUE #2580: mtPaint refuses to start when using Wayland
https://github.com/wjaguar/mtPaint/issues/50

dimkr:


01micko:
Pure wayland I presume?

dimkr:
> Pure wayland I presume?

Yes

dimkr:
Maybe only the `gdk_set_allowed_backends()` call needs to be removed, because all other code is unreachable under Wayland 🤔

01micko:
All good. I'm woofing up a slacko64 right now with your mtpaint bump and patch and will test in my labwc install.

dimkr:
Does `mtpaint -s` work? 🤔

01micko:
> Does `mtpaint -s` work? thinking

Wait a minute, just testing vanilla slacko build in qemu, then transferring binary to my labwc testbed.. stay tuned!

dimkr:
`mtpaint -s` seems to work under X, but black screen under Wayland ... import_pixmap() needs to be replaced with something like https://github.com/emersion/grim.

01micko:
It works! (ish)

- `mtpaint -s` fails to capture, just black
- error message when `mtpaint` is started 

```
~$ mtpaint

(mtpaint:7876): Gtk-CRITICAL **: 21:22:23.788: gtk_box_gadget_distribute: assertion 'size >= 0' failed in GtkScrollbar
```
- at least in `labwc`,  mtpaint starts with a small gui top left at 0,0 and the 'Settings Toolbar' starts in the centre of screen. Probably a `labwc` bug.

screeny (from `grim`)
20211015_21h18m47s_grim

woof-code/rootfs-petbuilds/mtpaint/petbuild
```diff
@@ -5,6 +5,7 @@ download() {
 build() {
     unzip -q mtpaint-3.50.09.zip
     cd mtPaint-199472ad6a4ecee6c8583fb5a504a2e99712b4fc
+    patch -p1 < ../wayland.patch
```
01micko, line 8: Should we not apply this globally and just add it to the _gtk3_ build?

01micko: Actually, since it works ok in X11 just disregard my last comment.

01micko (approved):

PR #2583, opened 2021-10-16
frontend_change: Some fixes

rizalmart:
Exit immediately if ROX pinboard is not running.


01micko (commented):
@dimkr this PR might be relevant to #2654

dimkr (approved):

dimkr:
Sounds like a good idea.

I see many ROX-Filer references in other places too (including jwm_config), so this fix is one among many required changes.

01micko:
Merged. I think it'll be fine.

PR #2584, opened 2021-10-16
connectwizard: Some improvements

rizalmart:
Added support for NetworkManager and Connman

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/connectwizard
+++ b/woof-code/rootfs-skeleton/usr/sbin/connectwizard
@@ -376,4 +376,18 @@ RADIOBUT="`echo "$RETSTR" | grep '^RADIO' | grep '"true"' | cut -f 1 -d '='`"
 [ "$RADIOBUT" = "RADIONM" ] && echo -e '#!/bin/sh\nexec NetworkManager' > /usr/local/bin/defaultconnect #180919
 [ "$RADIOBUT" = "RADIOCONM" ] && echo -e '#!/bin/sh\nexec connmand' > /usr/local/bin/defaultconnect #180919
 
+#Save default network tool as config file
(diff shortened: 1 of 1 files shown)


woof-code/rootfs-skeleton/usr/sbin/connectwizard
```diff
@@ -364,5 +373,7 @@ RADIOBUT="`echo "$RETSTR" | grep '^RADIO' | grep '"true"' | cut -f 1 -d '='`"
 [ "$RADIOBUT" = "RADIONETWIZARD" ] && echo -e '#!/bin/sh\nexec net-setup.sh' > /usr/local/bin/defaultconnect #100227
 [ "$RADIOBUT" = "RADIOSNS" ] && echo -e '#!/bin/sh\nexec sns' > /usr/local/bin/defaultconnect #100227
 [ "$RADIOBUT" = "RADIOPWF" ] && echo -e '#!/bin/sh\nexec peasywifi' > /usr/local/bin/defaultconnect #180919
+[ "$RADIOBUT" = "RADIONM" ] && echo -e '#!/bin/sh\nexec NetworkManager' > /usr/local/bin/defaultconnect #180919
+[ "$RADIOBUT" = "RADIOCONM" ] && echo -e '#!/bin/sh\nexec connmand' > /usr/local/bin/defaultconnect #180919
```
dimkr: I think this PR doesn't make any sense, unless there is some mechanism that prevents connman, nm, etc' from starting, when one of them is selected. They can't co-exist, they kill each other's network interfaces. And don't forget you also need to stop the UI from starting (connman-ui, nm-applet, etc') - this is inadequate.

rizalmart: I tried both connman and NetworkManager coexists. When I selected connman the NetworkManager got disabled and connman do the network management. I think the default network tool settings must be saved as config file not shell script.

PR #2586, opened 2021-10-16
network_default_connect: Some improvements

rizalmart:
* Improved connman support
* Added NetworkManager support


woof-code/rootfs-skeleton/usr/sbin/network_default_connect
```diff
@@ -111,6 +118,49 @@ case $NETCHOICE in
   /etc/rc.d/rc.network_basic #this only sets up interface 'lo'.
   /etc/rc.d/rc.network_eth &   #test for wired network.
  ;;
+ connman)
+  if [ "$(pidof connmand)" == "" ]; then
+   connmand &
```
dimkr: Please don't do this, it will conflict with /etc/init.d/connman

PR #2601, opened 2021-11-01
mountpartition: Added NTFS3 support

rizalmart:
ntfs3  is new native linux driver written by paragon software
It will use ntfs3 as primary for mounting ntfs formatted disk if detected otherwise use ntfs-3g as fallback


bigpup1:
May want to consider the answers to these questions.
https://www.paragon-software.com/us/home/ntfs3-driver-faq/

PR #2602, opened 2021-11-04
Use overlayfs in CI

dimkr:
Looks aufs is gone from hosted GitHub Actions runners, and aufs is becoming less common as overlayfs takes its place.

### DIFF
--- a/woof-code/3builddistro
+++ b/woof-code/3builddistro
@@ -1028,7 +1028,7 @@ if [ "$BUILD_DEVX" = "yes" ] ; then
 	done
 
 	echo -e "\ncleaning out whiteouts"
-	clean_out_whiteouts $sandbox3/devx # _00func
+	clean_out_whiteouts sandbox3/devx # _00func
 
 	sh $MWD/support/files2delete.sh sandbox3/devx
 	for i in usr/share/gtk-doc usr/share/doc/*-dev usr/share/doc/*-common ; do
--- a/woof-code/_00func
+++ b/woof-code/_00func
@@ -175,6 +175,10 @@ EOD
 }
 
 function set_layer_type() { # sets $LAYER_TYPE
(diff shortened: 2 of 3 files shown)


peabee:
This only applies to GitHub builds in the cloud??
petbuilds run natively on real puppy machines will still use aufs? not all puppy kernels include overlayfs.....

dimkr:
> This only applies to GitHub builds in the cloud??

Because I knew you or somebody else would oppose the switch to overlayfs everywhere.

> petbuilds run natively on real puppy machines will still use aufs?

Yes, and I knew you'd say that - that's why set_layer_type still selects aufs, unless running on GitHub Actions, where aufs doesn't work anymore.

01micko:
> not all puppy kernels include overlayfs.....

They should.

gyrog:
In my experience, overlayfs is not a drop-in replacement for aufs.
The most glaring example is the inability to change the layers in a stack after it is mounted.
The obvious result is that 'snapmergepuppy' cannot work (pupmode=13), and neither can 'sfs_load'.

"savefiles" are also a challange because of the requirement of 2 directories for the RW layer instead of just the 1.

I haven't used all Puppy huge kernels, but the ones I have used all had support for overlayfs.
The overlayfs support can be a module, without any problems.
Besides, if an "overlayfs" Puppy were released, then the included kernel would surely contain "overlayfs" support.

I'm writing this on Xenialpup with an overlayfs stack using my "mio" 'init'.

dimkr:
@gyrog After this PR, overlayfs is used **by woof-CE** and only when running on GitHub Actions, and **not in Puppy**. The sfs_load/savefiles/etc' discussion is irrelevant.

gyrog:
@dimkr,
my misunderstanding.

I have found overlayfs is very easy to use to "merge" a few directories, even in a "normal" aufs Puppy.

PR #2634, opened 2021-11-11
Add swaybg and grim to the labwc petbuild

dimkr:
`dbus-launch labwc` from the terminal starts something that starts to look more and more like a desktop environment 🐌


01micko:
Also this `gtkdialog` app [`wwals`](https://gist.github.com/01micko/192701ac9b329d3f737194b6623c08f5) for `swaybg` is useful - needs a .desktop file for the menu though.

dimkr:
> Also this `gtkdialog` app [`wwals`](https://gist.github.com/01micko/192701ac9b329d3f737194b6623c08f5) for `swaybg` is useful - needs a .desktop file for the menu though.

Let's do it in a separate PR then. We need to decide what to do first - add Wayland/labwc support to ptheme, or re-implement theming.

I think I prefer the second option, and do separate the GTK+ theme from the labwc theme and the wallpaper - one configuration tool for each of these things. But I'm still not 100% sure, and I think it's a decision worthy of discussion.

01micko (approved):

01micko:
> Let's do it in a separate PR then. We need to decide what to do first - add Wayland/labwc support to ptheme, or re-implement theming.
> 
> I think I prefer the second option, and do separate the GTK+ theme from the labwc theme and the wallpaper - one configuration tool for each of these things. But I'm still not 100% sure, and I think it's a decision worthy of discussion.

On that note Joe has released JWM-2.4.0 so pTheme is likely to need an overhaul anyway.

I think pTheme is ok and we should keep it the central place for theming. We can get the current window manager/compositor from /etc/windowmanager as long as `labwc` is written there. Still, the wwals script can be  
accessible from the 'More' button in ptheme, without needing a .desktop file.

dimkr:
> windowmanager

We can bypass xwin and .xinitrc, and rid ourselves of many ugly things including /etc/windowmanager.

I like the idea of the 'More' button, sounds like a good solution, at least for now.

01micko:
I'll merge this and we can sort the rest out later, perhaps opening a new issue.

PR #2653, opened 2021-12-03
add 'ppavolume' to pulseaudio_FIXUPHACK

01micko:
This is a simple gtkdialog volume control designed for wlroots compositors


01micko:
Note that this doesn't _depend_ on a wayland compositor and would likely work in any desktop environment and could be adjusted so that position works akin to `gtkdialog-splash`, but there really isn't much point at the moment

dimkr (approved):

PR #2654, opened 2021-12-04
Set the wallpaper using JWM, if ROX-Filer is missing

dimkr:
I want to make GTK+ 2 dependencies optional, and the first step is making ROX-Filer optional. (The other dependencies are Sylpheed, Xdialog and gtk2dialog).

Before and after `echo true > /etc/desktop_app`:

before
after


01micko (approved):
Good idea

PR #2669, opened 2021-12-08
Add an update tool to ext4 images

dimkr:
update

Once this branch is merged into the `images` branch, I'll run two dpup draft releases based on `images`, test update from the first release to the second release, delete the drafts and merge `images` to `testing`. (I tested the updater manually and it works for both UEFI and BIOS.)

### DIFF
--- a/merge2out
+++ b/merge2out
@@ -283,6 +283,7 @@ if which git &>/dev/null && [ "`git log -n 1`" != "" ]; then
 	else
 		echo "BUILD_FROM_WOOF='${GITBRANCH};${GITHEAD}'" >> ${WOOF_OUT}/DISTRO_SPECS
 	fi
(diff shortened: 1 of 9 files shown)


peabee (commented):
I already provide incremental updates by supplying xdelta files to convert a base .iso with a "System Update" menu item in "System"............. I don't want updates to be any more automated than this as I do not like automated updates - the user should be in control. I have not reviewed the changes proposed but am concerned about the quantity / extent of them.

dimkr:
> I already provide incremental updates by supplying xdelta

The next PR after this one will implement delta-based updates, with fallback to regular updates.

> I have not reviewed the changes proposed but am concerned about the quantity / extent of them.

There won't be automated updates: the user has to click the menu entry and approve if an update was found. In the case of Vanilla Dpup, updates will be available on a weekly basis, and they consist of bug and security fixes. They shouldn't break anything, and they should be very small once I introduce delta updates.

01micko (approved):

01micko:
> I don't like the idea of using binary blobs from an unknown source (like FrugalPup's EFI executables)

from #2666

PR #2672, opened 2021-12-11
update quicksetup and 2 close friends

01micko:
- xorgwizard (wayland support)
- pupx (wayland screen blank support)

### DIFF
--- a/woof-code/rootfs-skeleton/usr/bin/pupx
+++ b/woof-code/rootfs-skeleton/usr/bin/pupx
@@ -1,13 +1,13 @@
 #!/bin/bash
 
-SCREENSAVERFLAG='yes'
-SCREENVISIBLE="enabled"
-
 if [ -n "$WAYLAND_DISPLAY" ]; then
     . ~/.Xwaylandrc
+    SCREENSAVERFLAG='yes'
+    SCREENVISIBLE="enabled"
     if [ -n "$XWAYLAND_SCREENSAVER_DELAY" ]; then
         SCREENSAVERDELAY="$XWAYLAND_SCREENSAVER_DELAY"
     else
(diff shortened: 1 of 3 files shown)


woof-code/rootfs-skeleton/usr/bin/pupx
```diff
@@ -130,12 +135,18 @@ RETPARAMS="`gtkdialog --center --program=MAIN_DIALOG`"
...
 [ "$EXIT" != "save settings" ] && exit
 
-if [ -f ~/.Xwaylandrc ]; then
-    if [ "$SCREENSAVERFLAG" = "true" -a "$SCREENSAVERDELAY" -gt 0 ]; then
+if [ -n "$WAYLAND_DISPLAY" ]; then
+	if [ "$SCREENSAVERFLAG" = "true" -a "$SCREENSAVERDELAY" -gt 0 ]; then
         sed -i "s%^XWAYLAND_SCREENSAVER_DELAY=.*%XWAYLAND_SCREENSAVER_DELAY=$SCREENSAVERDELAY%" ~/.Xwaylandrc
+		if pidof labwc >/dev/null 2>&1 ; then # could add sway here
+			set_wl_monsleep
+			pidof swayidle >/dev/null 2>&1 && killall swayidle
```
dimkr, line 143: You don't really need this `pidof` - `killall` does the same lookup under /proc

dimkr:
This PR is tricky, I need to take a second look.

My intention is to gradually make ~/.Xwaylandrc the equivalent of Xorg.conf for Wayland compositors, and I'm not sure if the X.Org/Wayland distinction is still a good idea - maybe I'll need to do some refactoring so we're able to distinguish between X.Org and anything else.

PR #2674, opened 2021-12-11
update imgview for wayland

01micko:
still works fine in X based systems

### DIFF
--- a/woof-code/rootfs-packages/imageview/usr/sbin/imgview
+++ b/woof-code/rootfs-packages/imageview/usr/sbin/imgview
@@ -1,6 +1,6 @@
 #!/bin/bash
 # a simple image viewer using gtkdialog and bash
-# (c) Mick Amadio. 01micko@gmail.com, 2019, GPLv2 (/usr/share/doc/legal)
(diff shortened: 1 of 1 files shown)


woof-code/rootfs-packages/imageview/usr/sbin/imgview
```diff
@@ -30,10 +30,16 @@ MANUAL=true
 [ "$SLIDE" == 'true' ] && MANUAL=false
 
 cd "$CWD"
-
-read DIMS SCRX SCRY ROOTVERT << EOF
-`xwininfo -root | grep ' \-geometry ' | cut -f 1 -d '+' | tr 'x' ' '`
+if [ "$XDG_SESSION_TYPE" = 'wayland' ]; then
```
dimkr, line 33: Why XDG_SESSION_TYPE and not WAYLAND_DISPLAY? Who sets this one?

01micko: I'll change that.

01micko: Done

PR #2679, opened 2021-12-11
Add frugalify support to ext4 images

dimkr:
frugalify is a static executable that replaces initrd.gz, init, syslogd and klogd, while also providing periodic TRIM.

See https://github.com/puppylinux-woof-CE/frugalify/blob/master/README.md for more information about it.

I've used and tested it extensively in dpup OS, for almost a year, and I think it's a great way to simplify Puppy, get rid of some old cruft (like the static binaries in initrd.gz) and make it boot faster. It's mature enough to make it an option in 3builddistro.

(It supports both overlay and aufs; this PR only adds support for the overlay variant.)

As always, this feature is optional, and it's activated only when `BUILD_ISO=no FRUGALIFY=yes`.

### DIFF
--- a/woof-code/3builddistro
+++ b/woof-code/3builddistro
@@ -704,6 +704,9 @@ if [ ! "$ISOFLAG" ]; then
 		      rootfs-skeleton/usr/sbin/pupsave-backup \
 		      rootfs-complete/usr/share/applications/Pupsave-hot-backup.desktop
 	fi
+	if [ "$FRUGALIFY" ]; then
+		rm -f rootfs-complete/usr/share/applications/SFS-Load.desktop
+	fi
 fi
 
 # puppyPDF needs Abiword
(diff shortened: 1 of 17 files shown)


01micko:
*Frugalify* interests me. I haven't time right now to review but will do soon enough.

dimkr:
Just pushed puppylinux-woof-CE/frugalify@e79d9ba. Vanilla Dpup with frugalify is super fast! The second boot takes 12 seconds, in a x86_64 VM, with the hard drive image placed on an old 5400 RPM hard drive, a stopwatch from boot loader to desktop, and that's on my old Thinkpad. A persistent installation on the second drive in this machine, an old and slow mSATA SSD from AliExpress which is partially corrupt and therefore even slower than it was, is even faster than the VM.

dimkr:
Objections?

01micko (approved):

PR #2684, opened 2021-12-18
Add a "rolling" dpup build based on Debian Sid

### LINKED ISSUE #2611: Update wlroots to 0.15.0

dimkr:
Bullseye is simply too old to fix #2611 cleanly. Some wlroots 0.15.0 changes must be reverted to allow older dependency versions (for example, https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/bedfec94bb57cf7dc32f03fce463a30d3370bb20).

However, the situation is getting much worse over time, because Debian Bullseye was already "old" when it was released.

Therefore, all Wayland-related development should be much easier against a Sid base. When the Bookworm freeze starts, it should be possible to copy woof-distro/x86_64/debian/sid64 into woof-distro/x86_64/debian/bookworm64 and start stabilizing a Vanilla Dpup 10.x release.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,6 +49,10 @@ jobs:
             compat-distro: ubuntu
             compat-distro-version: bionic64
             kernel: 4.19.x-x86_64
+          - arch: x86_64
+            compat-distro: debian
+            compat-distro-version: sid64
+            kernel: 5.15.x-x86_64
           - arch: arm
             compat-distro: debian
             compat-distro-version: bullseye-veyron-speedy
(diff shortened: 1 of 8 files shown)


01micko:
I face a similar situation with Slackware. Once `-current` goes to `15.0` I'll probably still have to follow `-current`. Already `libdrm` is too old in Slackware :disappointed:

dimkr:
> I face a similar situation with Slackware. Once `-current` goes to `15.0` I'll probably still have to follow `-current`. Already `libdrm` is too old in Slackware 😞

https://github.com/puppylinux-woof-CE/woof-CE/pull/2685 updates wlroots to 0.15.0 and it builds fine on Debian Bullseye, despite the old libdrm. Had to do some ugly patches 😢 

But labwc is a bigger problem, because it requires wlroots master. The dependency versions are changing too often. Sid and Current based builds are the easiest way to work on labwc integration going forward.

01micko (approved):

PR #2688, opened 2021-12-18
pupx: modify for wayland

01micko:
add killall swayidle and restart swayidle

### DIFF
--- a/woof-code/rootfs-skeleton/usr/bin/pupx
+++ b/woof-code/rootfs-skeleton/usr/bin/pupx
@@ -4,10 +4,11 @@ SCREENSAVERFLAG='yes'
 SCREENVISIBLE="enabled"
 
 if [ -n "$WAYLAND_DISPLAY" ]; then
-    . ~/.Xwaylandrc
+    [ -f .Xwaylandrc ] && . ~/.Xwaylandrc
     if [ -n "$XWAYLAND_SCREENSAVER_DELAY" ]; then
         SCREENSAVERDELAY="$XWAYLAND_SCREENSAVER_DELAY"
     else
+        SCREENSAVERDELAY='600'
         SCREENSAVERFLAG='no'
         SCREENVISIBLE="disabled"
     fi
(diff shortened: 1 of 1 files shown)


woof-code/rootfs-skeleton/usr/bin/pupx
```diff
@@ -133,9 +137,15 @@ eval "$RETPARAMS"
 if [ -f ~/.Xwaylandrc ]; then
     if [ "$SCREENSAVERFLAG" = "true" -a "$SCREENSAVERDELAY" -gt 0 ]; then
         sed -i "s%^XWAYLAND_SCREENSAVER_DELAY=.*%XWAYLAND_SCREENSAVER_DELAY=$SCREENSAVERDELAY%" ~/.Xwaylandrc
+		if pidof labwc >/dev/null 2>&1 ; then # could add sway here
```
dimkr, line 140: In a Puppy with labwc but without Cage, ~/.Xwaylandrc doesn't exist

01micko: Cross that bridge when we come to it. I _was_ running pure wayland but hardly anything worked so call it a transitional phase.

01micko: However, if it's done now (as in #2689 ) at least we are somewhat future proof.

dimkr (approved):

woof-code/rootfs-skeleton/usr/bin/pupx
01micko: Done in fd27bb4

woof-code/rootfs-skeleton/usr/bin/pupx
```diff
@@ -146,6 +146,12 @@ if [ -f ~/.Xwaylandrc ]; then
         sed -i "s%^XWAYLAND_SCREENSAVER_DELAY=.*%XWAYLAND_SCREENSAVER_DELAY=%" ~/.Xwaylandrc
 	fi
     exit
+elif [ -f "$XDG_CONFIG_HOME/swayidle/config" ]; then # created by set_wl_monsleep at startup absent
```
dimkr, line 149: WAYLAND_DISPLAY is set when running labwc, so you'll never reach this `elif` - is that right?

woof-code/rootfs-skeleton/usr/bin/pupx
```diff
@@ -4,7 +4,7 @@ SCREENSAVERFLAG='yes'
 SCREENVISIBLE="enabled"
 
 if [ -n "$WAYLAND_DISPLAY" ]; then
-    . ~/.Xwaylandrc
+    [ -f .Xwaylandrc ] && . ~/.Xwaylandrc
```
01micko, line 7: Depends on existence of `~/.Xwaylandrc`

woof-code/rootfs-skeleton/usr/bin/pupx
01micko: See line 7 of the diff

01micko: BTW< I tested with replacing every instance of `,Xwaylandrc` with a dummy and it works as expected.. you get an unchecked checkbox which you need to check to sensitize and change the spinbutton value.

woof-code/rootfs-skeleton/usr/bin/pupx
```diff
@@ -146,6 +146,12 @@ if [ -f ~/.Xwaylandrc ]; then
         sed -i "s%^XWAYLAND_SCREENSAVER_DELAY=.*%XWAYLAND_SCREENSAVER_DELAY=%" ~/.Xwaylandrc
 	fi
     exit
+elif [ -f "$XDG_CONFIG_HOME/swayidle/config" ]; then # created by set_wl_monsleep at startup absent
+	CURDELAY=$(grep -o 'timeout [0-9]*[0-9]' $XDG_CONFIG_HOME/swayidle/config)
```
01micko, line 150: Only thing I need to do is check for stupid values here, like a single digit.

OR

Perhaps the range should start at something more sane like a min of `30` (which is still silly, but ok for testing).

Issue #2690, opened 2021-12-19
`filemnt` can break savefiles


01micko:
See [puppy forum post](https://forum.puppylinux.com/viewtopic.php?t=4720)

Suggested work around is to replace `rm -rf` with `rmdir`

There could be more of these.

PR #2691, opened 2021-12-19
filemnt: critical bugfix

01micko:
see https://forum.puppylinux.com/viewtopic.php?t=4720

Thanks to @JakeSFR


01micko:
re #2690

01micko:
I think this is critical enough, and a trivial enough change to merge without review.

dimkr:
Thanks for clean fix, and the useful commit message. Cherry-picked into stable.

PR #2692, opened 2021-12-19
change_kernels: fix UI with gtk-3

01micko:
backend is the same for now

## Before:

ch-k-before

## After:

ch-k-after


01micko:
Only UI is touched with space-expand stuff, etc so merging.

PR #2695, opened 2021-12-20
Use DISTRO_NAME in wallpapers

dimkr:


01micko (approved):
I wondered WTF this was about then the penny dropped!

PR #2703, opened 2021-12-21
Start labwc instead of X.Org, if present (fixes #2702)

### LINKED ISSUE #2702: Support for starting labwc instead of xwin in ~/.profile
This issue needs to be fixed before #2684 is merged

dimkr:


01micko:
I suppose we need to move all the envars into $XDG_CONFIG_HOME/labwc/environment to complete this or something.

01micko:
Actually, probably should make something for envars more global for _all_ wayland compositors such as `sway` and `wayfire`, not that we're likely to include them soon but when wayland does catch on in Puppyland, and it will eventually, we are well set up.

01micko (approved):

dimkr:
> Actually, probably should make something for envars more global for _all_ wayland compositors such as `sway` and `wayfire`, not that we're likely to include them soon but when wayland does catch on in Puppyland, and it will eventually, we are well set up.

I agree, this needs to be replaced with a `startwayland` that sets the environment variables, then `exec dbus-launch $(cat /etc/compositor)`.

dimkr:
@01micko I'm merging this one, so I can merge #2684 as well

PR #2725, opened 2021-12-25
Run xdg_autostart.sh when labwc starts

dimkr:
535810c needs to be cherry-picked into stable, it's a bug. xdg_autostart.sh can "run" .desktop files via xdg-open because ROX-Filer, unlike other file managers, just runs the command line in the Exec= line, without prompting the user first or defaulting to opening .desktop files with a text editor.

### DIFF
--- a/woof-code/rootfs-skeleton/bin/xdg_autostart.sh
+++ b/woof-code/rootfs-skeleton/bin/xdg_autostart.sh
@@ -26,6 +26,17 @@ verify_not_running() {
 	return 0
 }
 
+run_desktop() {
+	while read j
+	do
+		case $j in "Exec="*)
+			sh -c "${j#Exec=}" &
+			break
+			;;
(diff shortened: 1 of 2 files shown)


woof-code/rootfs-skeleton/bin/xdg_autostart.sh
```diff
@@ -26,6 +26,17 @@ verify_not_running() {
...
+			sh -c "${j#Exec=}" &
+			break
+			;;
+		esac
+	done < "$1"
+}
+
 #=================================================
 
 for i in /etc/xdg/autostart/*.desktop
```
01micko, line 42: Do you think this is feasible?

`for i in /etc/xdg/autostart/*.desktop $HOME/.config/autostart/*.desktop`

01micko (approved):

Issue #146, opened 2013-12-10
Deprecate cryptoloop in favour of dm-crypt for pup_save


go2null:
Most (Debian, Arch, RedHat) distros are using cryptsetup.

Please see comment from **JamesBond** at [Barry's Blog](http://bkhome.org/blog2/?viewDetailed=00144)

```
Posted on 22 Feb 2013, 20:06 by jamesbond
cryptoloop is going the dinosaurs way too
losetup from klibc (the latest I tried is 2.0.2) also supports encryption (both -e and -p).

That being the case, the cryptoloop module itself will be going soon as soon as this patch is merged with mainline https://lkml.org/lkml/2012/11/1/375.

cryptsetup (aka dm-crypt) has been around for years and getting more mature, as well as more secure than cryptoloop.

It is relatively easy to replace losetup with cryptsetup:

"losetup -e aes /dev/loop5 x.img"

can be replaced with

"cryptsetup -M plain -c aes-cbc-plain -h plain open `pwd`/x.img encrypted_device"

There is however no known methods of replacing the "xor" (aka "light") encryption.

For more info please see https://lkml.org/lkml/2012/11/2/162 (that URL still uses cryptsetup 1.5.x format, "open" is introduced in 1.6.0).

Fatdog already uses cryptsetup by default for new encrypted savefile although it still uses losetup as "compatibility mode" for older, existing encrypted savefile. Once the cryptoloop is gone I will have to migrate to cryptsetup for all encryption modes.

cheers!
```

Some good links:
*Ubuntu*:
https://help.ubuntu.com/community/EncryptedFilesystemOnIntrepid

*Arch*:
https://wiki.archlinux.org/index.php/Dm-crypt_with_LUKS
https://wiki.archlinux.org/index.php/System_Encryption_with_LUKS

Puppy

mavrothal:
Actually changing busybox ps to full ps is going to break a lot of scripts that have `ps` and expect the `ps axu` output. This could be fixed in woof scripts but there are also many wild pets that expect this behaviour. 
But what is the _actual_ problem with busybox ps or losetup?

wdlkmpx:
The same problem I had with truncate when I was writing this script.
http://pastebin.com/6XJuA4gj

Puppy is designed to confuse people. This is my filemnt
http://pastebin.com/KheG7QJc

01micko:
Should be dealt with in #789

PR #2729, opened 2021-12-26
yambar config - run /usr/local/pup_event/drive_all for removeables

01micko:
wasn't doing anything before


dimkr (approved):

01micko:
The good part of using `drive_all` is that clicking the icon opens the media, mounted or playing, dependant on what media it is, however there is a bug in yambar that I reported - [**#146** - Audio CD not recognized in removables](https://codeberg.org/dnkl/yambar/issues/146#issuecomment-294898).

Looks like it may get fixed soon.

Fixed in a PR and looking good.

audio_cd

audio_cd2

EDIT: PR is merged so likely we need to update `yambar` at some point soon :)

PR #2750, opened 2022-01-01
Add swaylock and run it when idle, on lid close, or on Super+l

dimkr:


01micko:
I tried swaylock and the shadow backend failed miserably for me, Does it work for you? ~or are you using `PAM`?~ (edit: I see it's disabled)

dimkr:
> I tried swaylock and the shadow backend failed miserably for me, Does it work for you or are you using `PAM`?

It works without PAM. I had to patch it, so it doesn't refuse to run as root.

01micko:
> It works without PAM. I had to patch it, so it doesn't refuse to run as root.

Yes, I patched too, and ran but the input was messed up and I couldn't unlock.

01micko:
Ok, I'll pass this.

01micko (approved):

PR #2767, opened 2022-01-11
Fix operator order when deciding whether or not to include simple_updater

dimkr:
```
before() {
	ISOFLAG=$1
	GITHUB_REPOSITORY=$2
	state=false
	[ ! "$ISOFLAG" -a "$GITHUB_REPOSITORY" ] && state=true
	echo $state
}

after() {
	ISOFLAG=$1
	GITHUB_REPOSITORY=$2
	state=false
	[ -z "$ISOFLAG" -a -n "$GITHUB_REPOSITORY" ] && state=true
	echo $state
}

before 1 ""
before "" ""
before "" a
before 1 a

after 1 ""
after "" ""
after "" a
after 1 a
```

The output of this truth table is:

```
true         <- before: simple_updater is included if building an ISO locally                        - bug
true         <- before: simple_updater is included if building an ext4 image locally             - bug
true         <- before: simple_updater is included if building an ext4 image on GitHub
false        <- before: simple_updater is included if building an ISO on GitHub                  - bug
false        <- after: simple_updater is not included if building an ISO locally
false        <- after: simple_updater is not included if building an ext4 image locally
true         <- after: simple_updater is included if building an ext4 image on GitHub
false        <- after: simple_updater is not included if building an ISO on GitHub
```

What I don't understand is, why CI builds don't behave according to this truth table. The two 3builddistro in the Vanilla Dpup build flow (`BUILD_ISO=yes` and `no`) work as they should, and simple_updater is **not** included in the latest dpup64 build:

```
2022-01-10T01:50:44.3805744Z copying acpid_busybox
2022-01-10T01:50:44.3806298Z copying backend_modprobe
2022-01-10T01:50:44.3806924Z copying change_kernels
2022-01-10T01:50:44.3807376Z copying check_deps_gui
2022-01-10T01:50:44.3807816Z copying firewall_ng
2022-01-10T01:50:44.3808247Z copying imageview
2022-01-10T01:50:44.3808671Z copying jwm_config
2022-01-10T01:50:44.3809079Z copying pclock
2022-01-10T01:50:44.3809469Z copying pdiag
2022-01-10T01:50:44.3809880Z copying pfilesearch
2022-01-10T01:50:44.3810303Z copying pfind
2022-01-10T01:50:44.3810712Z copying pschedule
2022-01-10T01:50:44.3811134Z copying ptheme
2022-01-10T01:50:44.3811538Z copying ptiming
2022-01-10T01:50:44.3811984Z copying pup_advert_blocker
2022-01-10T01:50:44.3812483Z copying redshiftgui_wrapper
2022-01-10T01:50:44.3812961Z copying wallpaper
2022-01-10T01:50:44.3813380Z copying xlock_gui
2022-01-10T01:50:44.3814074Z Copying to sandbox3/rootfs-complete...
```

### DIFF
--- a/woof-code/3builddistro
+++ b/woof-code/3builddistro
@@ -162,7 +162,7 @@ and edit it to include your customised package list.
 		simple_updater)
 			state=false
 			case $WOOF_TARGETARCH in
-			x86*) [ ! "$ISOFLAG" -a "$GITHUB_REPOSITORY" ] && state=true ;;
+			x86*) [ -z "$ISOFLAG" -a -n "$GITHUB_REPOSITORY" ] && state=true ;;
 			esac
 			;;
 		*)


01micko (approved):

peabee:
Seems to work - however
I wonder if it is "good design" to have items in rootfs-packages activated or deactivated by complex code buried deep in 3builddistro when we have a direct and obvious mechanism in rootfs-packages.conf
Also it maybe not appropriate to have items in rootfs-packages that do not apply to all builds - they should be included as specific .pets in the builds they are aimed at.....

Review was requested but I am unable to find the review button.......

dimkr:
> I wonder if it is "good design" to have items in rootfs-packages activated or deactivated by complex code buried deep in 3builddistro when we have a direct and obvious mechanism in rootfs-packages.conf

IMHO, if the sane default is to enable or disable a package in some build configurations, that should be the default. The user can always customize rootfs-packages.conf to override the default.

In this particular case, the default is to enable the updater when building ext4 images on GitHub, and disable it in other cases. (But the one-liner that made this decision was buggy, and it's fixed now.)

If every rootfs-packages.conf needs to be updated every time a new package is introduced, existing woof-CE configurations (especially, those that aren't part of the repo and don't have CI builds) will break every once in a while.

A compromise where 3builddistro has some sensible defaults, but warns you when it decides for you because your rootfs-package.conf is outdated (and doesn't explicitly enable or disable a certain package) sounds useful to me, and I can make CI fail if any rootfs-packages.conf inside the repo ever becomes outdated.

PR #2773, opened 2022-01-15
Add a Debian chroot environment to dpup ydrv

dimkr:
In a clean installation, there's a menu entry that runs xfce4-terminal as spot, under a Debian chroot environment:

terminal

/home/spot, /root, among other things, are accessible using bind mounts. You can modify your files, download stuff, etc'. The appearance of applications is consistent with applications installed outside of the chroot environment - as you can see, with #2774 applied, xfce4-terminal looks the same as a "native" xfce4-terminal installed via PPM.

You can install stuff using `apt`! For example, here's `sudo apt install pcmanfm`:

pcmanfm

After you install an application, there's a process that runs in the background and creates a menu entry whenever a new .desktop file is added:

chromium

(It also cleans up old menu entries when you `apt remove` a package, and synchronizes the menu with available applications every time it starts, to avoid broken menu entries after unclean shutdown. The creation and deletion of menu entries is nearly 100% seamless).

I tested many applications, including browsers and Steam, and they work just fine. Even Synaptic works!

There are issues with missing icons, applications that don't work because the rely on D-Bus activation (D-Bus runs outside of the chroot environment, but the services they activate are present only inside it). Generally speaking, many applications can't work in this weird setup, but it's super nice to have popular applications like browsers, Wine and Steam. I don't see a clean way to fix all the applications with strong D-Bus and systemd dependencies.

In addition, many applications appear in the menu more than once:

vlc

I remember PPM had some ugly hack for this: a predefined list of applications and a matching category for each, and a `sed` that modifies the .desktop file. I prefer to have no solution for this problem, until I find a better idea.

In the future, it should be relatively easy to add support for more than one chroot environment - for example, an Ubuntu chroot environment for up-to-date desktop applications like browsers and games, and a Debian chroot environment for everything else.

#2771 will make it possible to de-duplicate files, and layer the chroot environment on top of /. That will make it super small, because the chroot environment consists of packages in the main SFS, plus `apt` and its dependencies, and that's not a huge delta.


woof-code/rootfs-petbuilds/compat-chroot/petbuild
```diff
@@ -0,0 +1,51 @@
...
+    tar -xzf ${DISTRO_BINARY_COMPAT}-${DISTRO_COMPAT_VERSION}.tar.gz -C /opt
+    mv /opt/${DISTRO_BINARY_COMPAT}-${DISTRO_COMPAT_VERSION} /opt/compat-chroot
+
+    chroot /opt/compat-chroot /debootstrap/debootstrap --second-stage
+
+    chroot /opt/compat-chroot addgroup --gid `id -g spot` spot
+    yes "" | chroot /opt/compat-chroot adduser --uid `id -u spot` --ingroup spot --disabled-password spot
+    chroot /opt/compat-chroot adduser spot sudo
+
+    chmod 700 /opt/compat-chroot/root
```
dimkr, line 32: TODO: `ls /root` works as spot, and shouldn't

woof-code/rootfs-petbuilds/compat-chroot/usr/sbin/compat-chroot
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh -e
+
+BINDS="/dev /proc /sys /tmp /run /root /home/spot /usr/share/fonts /etc/fonts /usr/share/fontconfig /bin/busybox /bin/ash /usr/sbin/run-as-spot /usr/sbin/setup-spot /etc/hosts /etc/resolv.conf"
```
dimkr, line 3: TODO: in Puppy, /var/run/dbus is a symlink to /run/dbus, but in Debian, /var/run is a symlink to /run, creating a circular dependency that makes the D-Bus socket inaccessible

PR #2777, opened 2022-01-16
Fix the pupmoon petbuild. closes #2776

### LINKED ISSUE #2776: pupmoon petbuild doesn't preserve docs
I've marked this critical because it is a _breach of license_.

I'll just `mv` the docs over to `/etc` in the petbuild that way the `petbuild.sh` script doesn't delete it.

01micko:


woof-code/rootfs-petbuilds/pupmoon-font/petbuild
```diff
@@ -5,5 +5,7 @@ download() {
 build() {
     tar xJf pupmoon-font-1.0.tar.xz
     cd pupmoon-font-1.0
-    cp -arf usr/ /
+    mkdir -p etc/pupmoon
+    mv usr/share/doc/pupmoon/* etc/pupmoon
```
dimkr, line 9: Why not keep all documentation of all petbuilds, to prevent this issue from returning in other packages?

01micko: Your call on that :)

01micko: Can also be a case by case thing (per petbuild) if all docs are kept.

dimkr: I think it's OK to remove the deletion of /usr/share/doc from petbuilds.sh.

I can add explicit `rm -f` of large files later, where needed.

01micko: Fair enough. I'll close this and we'll sort out deletion of docs on 'as needed' basis in each petbuild.

dimkr: Maybe the problem is that most woof-CE builds come without docx, and docx is not auto-loaded (so the result is the same as having no documentation).

How about moving the license to to a file called LICENSE or COPYING? If we do that, we can change petbuilds.sh so it keeps files with these special names in the main package instead of moving them to _DOC.

PR #2778, opened 2022-01-16
Add support for automatic menu refresh when a .desktop is added or deleted

dimkr:
This improves the user experience with #2779.

### DIFF
--- a/woof-code/rootfs-petbuilds/fixmenusd/etc/xdg/autostart/fixmenusd.desktop
+++ b/woof-code/rootfs-petbuilds/fixmenusd/etc/xdg/autostart/fixmenusd.desktop
@@ -1,0 +1,7 @@
+[Desktop Entry]
+Version=1.0
+Name=fixmenusd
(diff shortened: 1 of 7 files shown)


01micko (commented):
Interesting. I would have thought just putting in a watch script to watch .desktop file locations would have been enough. I have thought of doing his before.

dimkr:
> Interesting. I would have thought just putting in a watch script to watch .desktop file locations would have been enough. I have thought of doing his before.

I like this simple inotify-based solution. It's lightweight and works well, especially after #2782.

Maybe I'll add some grace period - ignore events for 2s after each .desktop file creation/modification/deletion, then update the menu. Some packages have more than one .desktop file, so the second fixmenus is redundant. But this is nice-to-have, because most packages have just one, and #2782 makes it safe to run fixmenus again before it's done.

dimkr:
Merged into #2779 after optimization and testing

PR #2779, opened 2022-01-16
Add apt to dpup, as a layer above the main SFS

dimkr:
The wonders a small 19 MB SFS can do 😺 

update

vlc

synaptic

abiword


peabee (commented):
I guess I'm surprised at how complicated and pervasive these apt changes need to be...... 
neither pkg nor xbps (as in VoidPup) needed such extensive Woof-CE changes.....
On a philosophical point I think that initiatives like this that only affect a single build should use other Woof-CE "mechanisms" such as a specific specialized .pet to introduce just this feature and/or use of a pinstall in such a .pet or code in _00build.conf EXTRA_COMMANDS to achieve what is needed to incorporate the feature.

dimkr:
> neither pkg nor xbps (as in VoidPup) needed such extensive Woof-CE changes.....

The changes are not that big - most of the diff is the addition of users and groups. woof-code/support/bdrv.sh, the only new file, is pretty short, and I think it's really clear.

The customization of the Debian installation is needed to reduce its size, and prevent conflicts with Puppy. For example, the user must not `apt upgrade coreutils`, because that can easily replace /bin/cp, a symlink to /bin/busybox, which will replace busybox with cp, and the system will no longer boot.

VoidPup and any other Puppy that uses a foreign package manager will need the same kind of protection.

After this explanation, do you still find this PR "complicated and pervasive" (quote)?

EDIT: forgot to address this:

>  I think that initiatives like this that only affect a single build should use other Woof-CE "mechanisms" such as a specific specialized .pet

It's impossible. `apt` and its dependencies must sit below the main SFS, and be separate. And this "bdrv" must be built after the contents of rootfs-complete are known. Existing mechanisms, like pinstall.sh, package templates, ydrv, EXTRA_COMMANDS, you name it - won't work here.

mavrothal:
Interesting!
Do you know if changes in /etc/{group,passwd,shadow} affect spot user and the run-as-spot utility?
Also, there is no change in init to load bdrv. Is that intentional?

woof-code/3builddistro
```diff
@@ -962,20 +963,28 @@ fi
 
 if [ "$BUILD_SFS" = 'yes' ]; then
 	sh $MWD/support/files2delete.sh rootfs-complete
+
+	# frugalify puts bdrv below the main SFS
```
01micko: @mavrothal

01micko: Doesn't touch `init`

dimkr:
> Interesting! Do you know if changes in /etc/{group,passwd,shadow} affect spot user and the run-as-spot utility?

No, spot works. (dpup has many applications running as spot and they all work; Firefox, PulseAudio and Blueman, among others, run as spot. But this is unsurprising, because the things that make spot work are all in the main SFS and always override the Debian configuration files).

I added all users and groups defined in a clean Debian `debootstrap`. I fixed every collision between Puppy and Debian by giving the Puppy user/group a different ID. For example, `sshd` had UID 33: I changed it to 1003 so the `www-data` user can have the same ID it has on Debian.

Packages with files that must be owned by a specific user or group can break, but only if these packages contain executables that run as non-root users (for example, the `disk` group is no longer 100).

(apt needs the `_apt` user, but I decided to sync all users and groups in the name of compatibility. It's safe to add more users and groups, but a missing user can break an application.)

> Also, there is no change in init to load bdrv. Is that intentional?

Yes - for now, this feature only works with https://github.com/puppylinux-woof-CE/frugalify. https://github.com/puppylinux-woof-CE/frugalify/commit/62e6efaabcb42ffd8dbd79af1aab2d7113d14088 changes the SFS sorting algorithm, so bdrv is below the main SFS, then everything else.

For this "bdrv" concept to work with initrd.gz too, `init` must place bdrv below the main SFS.

woof-code/3builddistro
dimkr: Testing apt when booting with initrd.gz - so far, all basic tests passed. I'll add initrd.gz support to this PR soon.

dimkr:
> For this "bdrv" concept to work with initrd.gz too, init must place bdrv below the main SFS.

Done! Please review.

I had to move some apt dependencies from devx to the main SFS, because the ISO images don't come with devx.

~~I'll publish a test ISO soon,~~ so you can try it out. https://github.com/dimkr/woof-CE/releases/download/dpup-8.0-apt6/dpup-8.0.iso

dimkr:
https://github.com/dimkr/woof-CE/releases/download/vanilladpup-9.9.9-apt/vanilladpup-9.9.9.iso

Together with #2771:

wine32

links

steam

As you can see, it's possible to install 32-bit libraries and applications that rely on them, like Wine and Steam.

initrd-progs/0initrd/init
```diff
@@ -76,13 +76,15 @@ export KERNELVER="$(uname -r)"
 export DISTRODESC="${DISTRO_NAME} ${DISTRO_VERSION} - Linux ${KERNELVER} - `uname -m`"
 
 #precaution - if DISTRO_SPECS was not processed by 3builddistro...
+[ ! "$DISTRO_BDRVSFS" ] && DISTRO_BDRVSFS="bdrv_${DISTRO_FILE_PREFIX}_${DISTRO_VERSION}.sfs"
```
dimkr: @mavrothal Did I mess up initrd in any way? This code is messy and I want to make sure I didn't break anything.

dimkr:
https://github.com/dimkr/woof-CE/releases/download/vanilladpup-9.9.9-apt4/vanilladpup-9.9.9.iso

Missing icons are back:

icons

Installation of Xfce after first boot of the ISO, without devx, works now:

start

Xfce even works **(and works under Xwayland!)**:

xfce

After setting the GTK+ theme, icon theme and wallpaper it looks pretty usable:
 
xfce2

Same with KDE:

kde

And, the menu JWM is updated automatically and silently after every `apt install`.

IMHO this PR is useful and complete enough to merge, and everything is optional and disabled by default.

dimkr:
With 8a2d829 and 6c5f14e, all builds except dpup are unaffected by this PR.

72cc171 makes it safe to load bdrv above the main SFS, so it's unnecessary to change initrd to load bdrv first.

All basic tests passed (VLC, GIMP, Xfce, KDE, gdm3, and now also LibreOffice):

writer

@peabee @mavrothal @01micko Objections?

01micko (approved):

PR #2791, opened 2022-01-21
remasterpup2: Include flatpak packages

rizalmart:
Include installed flatpak packages upon remastering puppy


01micko (approved):
I'll pass this but soon all calls to [u]rxvt will be removed, including in `xterm`

PR #2793, opened 2022-01-22
z_update_system_cache.sh: Added udev rule reload

rizalmart:
Reload udev rule if the packages contains udev rules


woof-code/rootfs-skeleton/usr/local/petget/z_update_system_cache.sh
```diff
@@ -85,6 +85,8 @@ if [ "$(grep -q -m 1 '/udev/hwdb.d/' $PKGFILES)" != "" ]; then
   [ "$(udevadm --help 2>&1 | grep hwdb)" != "" ] && udevadm hwdb --update
 fi
 
+[ "$(grep -q -m 1 '/udev/rules.d/' $PKGFILES)" != "" ] && udevadm control --reload-rules
```
01micko: Wrong, see the line below.
`grep -q` only returns exit code, `0` for success and any thing else is a fail.

should be 
```sh
grep -q -m 1 '/udev/rules.d/' $PKGFILES && udevadm control --reload-rules
```

01micko (changes requested):

01micko (approved):

PR #2797, opened 2022-01-23
Make setup-spot look for executables under all directories in PATH

dimkr:


woof-code/rootfs-skeleton/usr/sbin/setup-spot
```diff
@@ -58,7 +58,11 @@ do
 
  ONEAPPname="`echo -n "$ONEAPP" | cut -f 1 -d '='`"  #ex: seamonkey
  ONEAPPvalue="`echo -n "$ONEAPP" | cut -f 2 -d '='`" #ex: true
- ONEAPPspec="`find ${PREFIXDIR}/bin ${PREFIXDIR}/sbin ${PREFIXDIR}/usr/bin ${PREFIXDIR}/usr/sbin ${PREFIXDIR}/usr/local/bin -mindepth 1 -maxdepth 1 -name ${ONEAPPname} | head -n 1`"
+ if [ $RUNNINGPUP -eq 1 ]; then
```
dimkr, line 61: The meaning of RUNNINGPUP is not what you expect. RUNNINGPUP=0 means "running under Puppy" and RUNNINGPUP=1 means "running during 3builddistro".

01micko: Yes that was wrong from start. I think some people get confused with evaluating the output of a program (success = 0, fail = non-zero) and setting a 'truth' variable (1=true, 0=false)

01micko: For better or worse should probably reverse that. 

<OT>lol, for better or worse my 23 year wedding anniversary today</OT>

01micko (approved):

Issue #2803, opened 2022-01-24
Frugality does not allow partition resizing (?) and other little things


mavrothal:
Tried 9.1.2 briefly on a PC (BTW the included bootx86.efi does not boot intel MACs - I use rEFInd that is good for both). 
Worked OK. 
One issue is that you can not resize the OS partition from within as it is always mounted. 
Given that also holds the savefile, 2GB is too small.
(Yes you can do it in a different system but does not look good…)
Maybe pfix=ram should move everything in RAM to allow unmount and resize.

Also if this is to be run from a USB stick, ext4 without journal will be better.
Or maybe (if not always in RAM - not sure about that) upper layer should be in RAM and only saved at shoutdown to avoid continuous rw.

BTW having everything in RAM is really important if a(n old 2GB) USB2 stick is used that can slow thinks down a lot.
Also since power management is not quite there, running from a laptop that goes to suspension sometimes results in USB unmounting and hanging the system.
 
Another little annoyance is the behaviour of /mnt/home desktop icon that instead of opening the drive when clicking, brings up pmount unless you (try to) unmount it first clicking the eject arrow. 

A grievance about distribution as IMG is that is making hard to test in a VM (VirtualBox, VMWare, Parallels). At least in early sages you should make testing as easy and safe as possible.
Maybe a script should be included  to the build an ISO out of the IMG. If isolinux and friends is too much additional stuff, they could be downloaded by the script to be included in the ISO image.

More if I mange to get a hold of that PC or the next version boots Macs

dimkr:
> Also if this is to be run from a USB stick, ext4 without journal will be better.

https://github.com/puppylinux-woof-CE/woof-CE/blob/24a04e5db270a35a1a37e2521eb889352a3674b5/woof-code/support/pc_image.sh#L24

> Another little annoyance is the behaviour of /mnt/home desktop icon that instead of opening the drive when clicking, brings up pmount unless you (try to) unmount it first clicking the eject arrow.

#2804

PR #2804, opened 2022-01-24
Fix detection of mounted partition in pmount if mounted at /

dimkr:
This is one of the issues in #2803:

```
Another little annoyance is the behaviour of /mnt/home desktop icon that instead of opening the drive when clicking, brings up pmount unless you (try to) unmount it first clicking the eject arrow.
```

The user should not be able to unmount /.

```
~/woof-CE$ mount
...
/dev/root on /initrd type ext4 (rw,noatime)               <- this is /dev/sdb2
...
~/woof-CE$ grep 8:18 /proc/self/mountinfo   
56 55 8:18 / /initrd rw,noatime - ext4 /dev/root rw                   <- it's mounted at /initrd (originally at /, it was moved); no "sdb2" in this line
57 55 8:18 / /mnt/home rw,noatime - ext4 /dev/root rw                <- and at /mnt/home; no "sdb2" in this line either
~/woof-CE$ probepart -k -extra-info
...
/dev/sdb2|ext4|62255104||not_mounted                                 <- bug
~/woof-CE$ ./woof-code/rootfs-skeleton/sbin/probepart -k -extra-info
...
/dev/sdb2|ext4|62255104||mounted
```

### DIFF
--- a/woof-code/rootfs-skeleton/sbin/probepart
+++ b/woof-code/rootfs-skeleton/sbin/probepart
@@ -144,6 +144,12 @@ probepart_func() {
 		while read dev etc ; do
 			[ "$dev" = "$blockdev" ] && MOUNT_STATUS="mounted" && break
 		done < /proc/mounts
+		if [ "$MOUNT_STATUS" != "mounted" ]; then
+			devmajorminor=`cat /sys/class/block/${ONEDEV}/dev`
+			while read mountid parentid majorminor etc; do
+				[ "$majorminor" = "$devmajorminor" ] && MOUNT_STATUS="mounted" && break
+			done < /proc/self/mountinfo
+		fi
 		EXTRA_STUFF="|${PARTITION_LABEL}|${MOUNT_STATUS}"
 	fi
 
--- a/woof-code/rootfs-skeleton/usr/sbin/pmount
+++ b/woof-code/rootfs-skeleton/usr/sbin/pmount
@@ -473,7 +473,10 @@ do
 	if [ "$ONEFS" = "crypto_LUKS" ]; then
 		ONEMNTPT="$(grep " /mnt/$DEVNAME " /proc/mounts | cut -f 2 -d ' ')"
 	else
-		ONEMNTPT="`grep "${ONEDEV} " /proc/self/mountinfo | grep '/ ' | cut -f 5 -d ' '`"
+		MAJORMINOR=`cat /sys/class/block/${DEVNAME}/dev`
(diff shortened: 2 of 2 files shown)


mavrothal:
After applying f19046b behaviour did not change

dimkr:
> After applying [f19046b](https://github.com/puppylinux-woof-CE/woof-CE/commit/f19046bee040e3871ccb1518437fc1c2330fad2c) behaviour did not change

Make sure you replaced /usr/sbin/pmount, otherwise the fix doesn't work. pmount re-runs itself from /usr/sbin, in https://github.com/puppylinux-woof-CE/woof-CE/blob/f19046bee040e3871ccb1518437fc1c2330fad2c/woof-code/rootfs-skeleton/usr/sbin/pmount#L44 and https://github.com/puppylinux-woof-CE/woof-CE/blob/f19046bee040e3871ccb1518437fc1c2330fad2c/woof-code/rootfs-skeleton/usr/sbin/pmount#L745.

The expected behavior is inability to unmount the partition, and the "Show location" menu entry should work.

mavrothal:
I did replace the files. Even rebooted just in case.
It could never unmount home, was just indicating that it could and still does.
Also it keeps bringing pMount up when you click on it after you click to unmount it.
So was never a real harm. Just strange behaviour. Which persists

dimkr:
> It could never unmount home

You can't unmount it, it's still in use, because it's a layer of the overlay at /.

pmount offers you to unmount every mounted partition, but it doesn't know if mounting can succeed. This is a design flaw, beyond the scope of this PR. EDIT: forgot to mention, pmount has another design flaw: it restarts silently when it fails to unmount a partition.

This PR only prevents pmount from mounting the same partition again, under a different mount point.

01micko (approved):

mavrothal:
The problem I say/see is that pup_event/drive_all can not figure out which partition is /mnt/home
It uses mounts to identify home but mount shows /mnt/home mounted on /dev/root and the the frugality partition  /dev/sdXN mounted on /mnt/sdXN. This is what messes up pup_event and pmount
Should  add code to identify the drive of /mnt/home (by blkid?) in frugality.

dimkr:
> Should add code to identify the drive of /mnt/home (by blkid?) in frugality.

Does this mean you want /mnt/sdb1 in addition to the /mnt/home bind mount? Do I understand you correctly?

This can be done quite easily.

mavrothal:
No, I just want pup_event and pmount to know which  partition is the /mnt/home partition and act accordingly.
I can post a bunch of pictures, but try clicking on the home desktop drive icon and see for yourself.

PR #2805, opened 2022-01-25
Add a urxvt and xterm replacement that runs lxterminal (fixes #2716)

### LINKED ISSUE #2716: Drop the rxvt-unicode dependency
All these calls make labwc start Xwayland, which consumes >50 MB of RAM:

```
fgrep -l 'rxvt -' `find rootfs-skeleton -type f | grep -v /etc`
rootfs-skeleton/usr/local/petget/removemodes.sh
rootfs-skeleton/usr/local/petget/pkg_chooser.sh
rootfs-skeleton/usr/local/petget/testurls.sh
rootfs-skeleton/usr/local/petget/installmodes.sh
rootfs-skeleton/usr/local/petget/configure.sh
rootfs-skeleton/usr/share/applications/Top-view-running-processes.desktop
rootfs-skeleton/usr/bin/xterm
rootfs-skeleton/usr/bin/wvdialshell
rootfs-skeleton/usr/sbin/quicksetup
rootfs-skeleton/usr/sbin/dir2iso
rootfs-skeleton/usr/sbin/puppyapps
rootfs-skeleton/usr/sbin/remasterpup2
rootfs-skeleton/usr/sbin/download_file
rootfs-skeleton/usr/sbin/momanager
rootfs-skeleton/usr/sbin/pmount
rootfs-skeleton/usr/sbin/Pudd
rootfs-skeleton/usr/sbin/burniso2cd
rootfs-skeleton/usr/sbin/connectwizard
```

Probably, most `rxvt -e` can be replaced with `defaultterminal -e`.

dimkr:
When PPM runs `xterm`, it reaches `woof-code/rootfs-skeleton/usr/bin/xterm`, a shell script (!!!) that writes a shell script (`$command; read`) to simulate `--hold`, if `--hold` was specified. Then, it runs `urxvt` with the same arguments, minus `--hold`.

Ugly! But wait, it gets better.

With this PR, `urxvt` is a tiny C wrapper that ~~runs `xterm` with the same arguments if `--hold` was specified. Why? Because it's inconsistent: woof-CE has some direct `urxvt --hold` calls.~~ runs `lxterminal -e sh -c "command; read"`, to simulate what this `xterm` hack does.

Otherwise, this wrapper just runs ~~`defaultterminal`~~ `lxterminal` with `-e` and everything that came after `-e`. All other arguments (for example, `-title`) are ignored, ~~because some terminal emulators don't support them.~~

If `-e` wasn't specified, the wrapper just calls ~~`defaultterminal`~~ `lxterminal` with all arguments. For example, `urxvt --help` runs `lxterminal --help`.

This ~~two or four stage chain (`urxvt->defaultterminal` or `urxvt->xterm->urxvt->defaultterminal`)~~ makes it possible to ~~use any terminal emulator that supports `-e` as `defaultterminal`,~~ migrate all existing users of `rxvt` or `xterm` to ~~`defaultterminal`~~ `lxterminal` ~~(often, a nicer terminal emulator with anti-aliased fonts and even some degree of hardware acceleration)~~ and completely drop the rxvt-unicode dependency.

~~Yes, there is some risk here - `defaultterminal` must not run `urxvt`.~~

If all new code starts using `defaultterminal -e` directly, this stub and the `xterm` wrapper (years and years of technical debt) can be deleted in the future.

Screenshots of PPM and pmount under labwc, with lxterminal as a replacemnt for rxvt-unicode:

ppm

pmount

### DIFF
--- a/woof-code/rootfs-petbuilds/lxterminal/petbuild
+++ b/woof-code/rootfs-petbuilds/lxterminal/petbuild
@@ -15,6 +15,9 @@ build() {
         ./configure --prefix=/usr --enable-gtk3
     fi
     make install
+    cd ..
     rm -rf /usr/share/icons
     sed -e 's/^Categories=.*/Categories=TerminalEmulator;/' -e 's~^Icon=.*~Icon=/usr/share/pixmaps/puppy/terminal.svg~' -i /usr/share/applications/lxterminal.desktop
+
+    $CC $CFLAGS `pkg-config --cflags glib-2.0` -D_GNU_SOURCE urxvt.c $LDFLAGS `pkg-config --libs glib-2.0` -o /usr/bin/lxterminal-urxvt
 }
--- a/woof-code/rootfs-petbuilds/lxterminal/pinstall.sh
+++ b/woof-code/rootfs-petbuilds/lxterminal/pinstall.sh
@@ -1,3 +1,12 @@
 echo '#!/bin/sh
 exec lxterminal "$@"' > usr/local/bin/defaultterminal
 chmod 755 usr/local/bin/defaultterminal
+
+if [ -e usr/bin/urxvt ]; then
+	rm -f usr/bin/lxterminal-urxvt
+else
+	ln -s lxterminal-urxvt usr/bin/urxvt
+	ln -s lxterminal-urxvt usr/bin/rxvt
+	rm -f usr/bin/xterm
+	ln -s lxterminal-urxvt usr/bin/xterm
+fi
--- a/woof-code/rootfs-petbuilds/lxterminal/urxvt.c
+++ b/woof-code/rootfs-petbuilds/lxterminal/urxvt.c
@@ -1,0 +1,53 @@
+#include <string.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <glib.h>
+#include <glib/gprintf.h>
+
+int main(int argc, char **argv)
+{
+	static char *new_argv[32] = {
+		"lxterminal",
+		"--no-remote",
+		"-e"
+	};
+	gchar *cmd;
+	int i;
+	gboolean hold = FALSE;
+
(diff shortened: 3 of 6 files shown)


01micko:
This seems like a complex solution to a simple problem.

This what I did:
```sh
#!/bin/sh
#rxvt does not understand the '-hold' option.
#urxvt 20120214
#defaultterminal 20220119
case "$*" in *"-hold "*)
 EXECCOMMAND="`echo -n ${*} | grep -o ' \-e .*' | sed -e 's/ \-e //'`"
 if [ "$EXECCOMMAND" != "" ];then
   (
    echo '#!/bin/sh'
    echo "$EXECCOMMAND"
    echo 'echo'
    echo 'echo -n "FINISHED. PRESS ENTER KEY TO CLOSE THIS WINDOW: "'
    echo 'read simuldone'
   ) > /tmp/xterm_simulate_hold.sh
   exec defaultterminal -e sh /tmp/xterm_simulate_hold.sh
 fi
 ;;
esac

exec defaultterminal "${@}"

### END ###
```

01micko:
I suppose though all calls to `urxvt` can stay for now with your solution.. I'll think about it.

dimkr:
> This seems like a complex solution to a simple problem.

It's basically the same, but in C instead, and lxterminal-specific because of `--no-remote`. Without this, `lxterminal -e` doesn't block until the console application exits, and this is why it must be lxterminal-specific.

(Originally it was a separate petbuild, but then I found out this non-blocking behavior of -e is inconsistent with other terminals.)

dimkr:
@01micko What do you think?

01micko (approved):

01micko:
> @01micko What do you think?

:+1:

dimkr:
I'm merging this one then: it should affect only builds that use the lxterminal petbuild and don't have rxvt-unicode (= only dpup).

PR #2813, opened 2022-01-28
Partial: add support for .xbps files

peabee:
PARTIAL Add support for .xbps files
Cannot:
upload:  /root/.config/rox.sourceforge.net/OpenWith/.application_x-voidlinux-package

or add: 
/usr/share/icons/Puppy Flat/scalable/mimetypes/application-x-archlinux-package-zst.svg
/usr/share/icons/Puppy Flat/scalable/mimetypes/application-x-voidlinux-package.svg
/usr/share/icons/Puppy Standard/scalable/mimetypes/application-x-archlinux-package-zst.svg
/usr/share/icons/Puppy Standard/scalable/mimetypes/application-x-voidlinux-package.svg

all 4 as links to application-x-compressed.svg

n.b. Can no longer do cli push:
remote: Support for password authentication was removed on August 13, 2021. Please use a personal access token instead.
remote: Please see https://github.blog/2020-12-15-token-authentication-requirements-for-git-operations/ for more information.
fatal: Authentication failed for 'https://github.com/peabee/woof-CE.git/'

PR #2819, opened 2022-01-29
partial fix for #2813

01micko:
@peabee 

#2813 still needs more cleanup. You need to fix up those `common[32|64]` files you had in a terrible location.


peabee (approved):
Yes - sorry again - they must have been lurking in my fork from 4 years ago - can't remember how or why!!!

PR #2820, opened 2022-01-30
Add 431-dark.svg wallpaper

01micko:
This is based on 431.svg with 'moon' rays instead of sun rays and few strtegically placed stars.
Also added to rootfs-complete/pinstall.sh

#### before

431-b

#### after

431-a

### DIFF
--- a/woof-code/rootfs-skeleton/pinstall.sh
+++ b/woof-code/rootfs-skeleton/pinstall.sh
@@ -7,16 +7,18 @@
 
 sed -i "s/Puppy Linux/${DISTRO_NAME}/g" usr/share/backgrounds/*.svg
 # take some pixels off for better alignment for 431.svg
-if [ -e usr/share/backgrounds/431.svg ];then
-	XVAL=872.666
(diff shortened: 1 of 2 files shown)


dimkr (approved):
Looks good. Does it work well with the traditional desktop icon layout? The bright spots are dangerous, they might make white labels unreadable.

Also, does it work with Polished-Blue or needs a more purplish/darker Polished-Blue?

01micko:
> Looks good. Does it work well with the traditional desktop icon layout? The bright spots are dangerous, they might make white labels unreadable.

Haven't tried, but labels have dark shading anyway. Should be legible.

> Also, does it work with Polished-Blue or needs a more purplish/darker Polished-Blue?

I've been using it with Stardust_dark. Which is grey/dark orange, seems to complement well. But a purpler theme would be a good addition.

dimkr:
> But a purpler theme would be a good addition.

Great. Let's merge this one, and I'll try to create a Polished-Blue-Dark variant.

PR #2822, opened 2022-01-30
add numlockt.c to labwc petbuild

01micko:
This is a very simple program to turn on numlock. That's it.


01micko:
There will be big changes to `quicksetup` as soon as I get time to refine them.

dimkr (approved):

PR #2825, opened 2022-01-31
Allow .xbps packages to be used in builds

peabee:


woof-code/rootfs-skeleton/usr/local/petget/verifypkg.sh
```diff
@@ -17,7 +17,10 @@ case $DLPKG in
 		tar --force-local -tf "$DLPKG" >/dev/null 2>&1
 		RETVAL=$?
 		;;
+	*.xbps)
+		RETVAL=0
```
01micko: I'm not getting this. 

Why in _all_ other cases is there a decompression call and success or failure recorded for the next step but this just returns success?

01micko (changes requested):

woof-code/rootfs-skeleton/usr/local/petget/verifypkg.sh
peabee: Chose not to do checks but I guess something could be added - I've NEVER seen the checks for other package types triggered.

peabee: check added

01micko (approved):

PR #2831, opened 2022-01-31
Make it possible to use only the icon petbuilds, without devx (fixes #2827)

### LINKED ISSUE #2827: Philosophy - petbuilds - debate
Making system builds rely on petbuilds means that a working devx is needed before a build can be undertaken.
When starting a new build, where even the main system may not work at first attempt having to have a working devx in order to do petbuilds in order to build the new system is a tall (impossible) ask.
Therefore it must be possible to do builds with only pre-built components and get that working before using the petbuild system.
For debate............

Note: 3builddistro says:
"WARNING: PETBUILDS is empty, this may be a hard error in the future"
and forces:
PETBUILDS="pmaterial_icons puppy_flat_icons puppy_standard_icons"
which means the build fails if the devx is not yet available and/or working

dimkr:


peabee (commented):
My comments in #2827 were not just about icon petbuilds - I will leave some further comments there. However being able to petbuild icons without having to build devx (which slows down a system build considerably) seems like a good idea.

PR #2833, opened 2022-02-01
Add special handling for the /mnt/home desktop icon under PUPMODE 6

dimkr:


mavrothal:
Works as expected in pupmod 6 and 12.

PR #2834, opened 2022-02-01
If PETBUILDS is unspecified and GTK+ 3 is new enough, auto-enable themes

dimkr:


peabee (commented):
Not clear to me what this is trying to do.....

Why do you have to auto-enable anything? Surely the _00build.conf specifies how the build should proceed? Why is automation needed?

dimkr:
You missed the "if". If PETBUILDS is not specified, which can happen if someone took an old _00build.conf, 3builddistro warns the user and uses sensible defaults.

peabee:
_00build.conf should be updated in such a case
I do not like build decisions which are hidden away in code - wherever possible they should be visible and recorded in .conf data files that can be easily read and altered.

dimkr:
> I do not like build decisions which are hidden away in code

I don't think it's right to modify configuration files from code. When you use a configuration file, you assume the behavior and output match what you put in the configuration file, and not some modified version of it.

And an "upgrade tool" that fixes bad configuration files is technically infeasible: for example, to auto-enable the GTK+ 3 theme petbuilds in an old configuration file, you need to know the GTK+ 3 minor version. How do you do that from outside of woof-CE, without a prior 0setup run and before you have the version number?

You can call this a "philosophical debate", but the reality is that configuration files rot quickly, as woof-CE and .pet package repos change. Take each and every Puppy built in the last couple of years, and you'll see it can't be rebuilt using the latest woof-CE.

I don't get this universal resistance to change - why do you care so much about not changing woof-CE, if this fallback behavior doesn't affect your own builds in any way?

peabee:
> I don't think it's right to modify configuration files from code. ......
> 
> I don't get this universal resistance to change - why do you care so much about not changing woof-CE, if this fallback behavior doesn't affect your own builds in any way?

I definitely wasn't suggesting this - just manual modification / updating.

Sorry - untrue. Many many changes have occurred that I have not commented on.

dimkr:
> just manual modification / updating

But that's exactly the problem ... if the configuration file is old or misses fields introduced in a newer woof-CE, you get broken/bad output. This PR is trying to help the users with bad configuration files, what's so wrong about that?

peabee:
Then the bad configuration file should be updated to be a good configuration file.

peabee:
A check at the start of 3builddistro that the config files were complete and sane and a refusal to proceed if they weren't would be acceptable.

01micko:
Should be able to check the `mtime` on the config file and reject it if it's older than a month or so.

dimkr:
> Should be able to check the `mtime` on the config file and reject it if it's older than a month or so.

That won't work. If you extract _00build.conf from an ISO, the modification time won't be preserved unless you use `cp -p` or something.

PR #2835, opened 2022-02-01
Set CONFIG_NLS_{ASCII,UTF8}=y in 5.1{0,5}.x kernels

dimkr:


peabee (commented):
Need a build so Marv can test - he has the problem with his particular partition setup - I don't.

dimkr:
> Need a build so Marv can test - he has the problem with his particular partition setup - I don't.

Give it 1.5-2 hours and you'll have it.

https://github.com/puppylinux-woof-CE/woof-CE/actions/runs/1779730532

Issue #2561, opened 2021-10-05
The Midnight Commander menu entry doesn't work


dimkr:
To reproduce on dpup:
1. Install mc
2. Try to run it via the menu

The .desktop file has `Exec=TERM=xterm mc`, so ~/.jwmrc runs `rxvt -title 'Midnight Commander' -e 'TERM=xterm mc'`.  Instead, it should run `rxvt -title 'Midnight Commander' -e sh -c 'TERM=xterm mc'`.

PR #2846, opened 2022-02-06
Auto-configure applications to run as spot when executable is created

dimkr:
This is much stronger than PPM's hack, because this can also handle manually installed packages and packages installed via apt.

And it doesn't rely on parsing .desktop files, which can be tricky (for example, stuff like `Exec=TERM=xterm mc`, a la #2561), and sometimes there's no .desktop file.

### DIFF
--- a/woof-code/rootfs-petbuilds/fixmenusd/fixmenusd.c
+++ b/woof-code/rootfs-petbuilds/fixmenusd/fixmenusd.c
@@ -9,6 +9,10 @@
 #include <signal.h>
 #include <sys/stat.h>
 #include <fcntl.h>
+#include <regex.h>
+#include <stdio.h>
+
+#define SPOTLIST "^(firefox|firefox-[a-z]+|google-chrome-[a-z]+|chromium|vivaldi-[a-z]+|brave-browser|microsoft-edge-[a-z]+|transmission-[a-z]+|seamonkey|sylpheed|claws-mail|thunderbird|vlc|steam)$"
 
 static int
 sh(const char *cmd, const sigset_t *set)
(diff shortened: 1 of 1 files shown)


woof-code/rootfs-petbuilds/fixmenusd/fixmenusd.c
```diff
@@ -61,6 +110,9 @@ handle_events(const int fd)
 			    (strcmp(&event->name[len - (sizeof(".desktop") - 1)], ".desktop") != 0))
 				continue;
 
+			if (!(event->mask & IN_DELETE))
+				setup_spot(re, event->name, set);
```
dimkr: Bug here: we don't know if the executable is already there when the .desktop file is created

woof-code/rootfs-petbuilds/fixmenusd/fixmenusd.c
```diff
@@ -9,6 +9,10 @@
 #include <signal.h>
 #include <sys/stat.h>
 #include <fcntl.h>
+#include <regex.h>
+#include <stdio.h>
+
+#define SPOTLIST "^(firefox|firefox-[a-z]+|google-chrome-[a-z]+|chromium|vivaldi-[a-z]+|brave-browser|transmission-[a-z]+|seamonkey|sylpheed|claws-mail|thunderbird|vlc|steam)$"
```
dimkr: Tested Chrome (the official .deb, manually extracted to /), Vivaldi (the official .deb, manually extracted to /), Chromium (apt), Thunderbird (apt) and VLC (apt).

I also tested installation of Chrome and Vivaldi using petget, by clicking the .deb file. PPM alters the .desktop file and the application is still successfully configured to run as spot, as expected.

woof-code/rootfs-petbuilds/fixmenusd/fixmenusd.c
```diff
@@ -33,8 +37,40 @@ sh(const char *cmd, const sigset_t *set)
...
+	size_t len;
+	char *cmd;
+
+	if (regexec(re, event->name, 0, NULL, 0) != 0)
+		return;
+
+	if (fstatat(usrbin, event->name, &stbuf, AT_SYMLINK_NOFOLLOW) < 0)
+		return;
+
+	if (((event->mask & IN_CREATE) && !S_ISLNK(stbuf.st_mode)) ||
```
dimkr, line 54: Even links are handled correctly - for example, `ln -s /usr/bin/lxtask /usr/bin/microsoft-edge-stable` triggers `setup_spot microsoft-edge-stable=true` and it works

01micko (approved):

PR #2848, opened 2022-02-07
Connectwizard regression fix and related fixes and refinements

rserwin1:
This commit corrects a recent change to connectwizard and makes some minor changes for probably-rare cases.  It removes a dependency on 'ethtool', allows control of ethernet connections when no network managers have been used, and may reduce the incidence of invalid IP addresses if MAC spoofing is in effect.  Details:
 - Connectwizard windows are centered on screen (as in FossaPup), instead of cascaded from upper left.
 - Connectwizard no longer sets current manager name for pgprs because pgprs now changes it only when connecting, allowing viewing of its setup without making it "current".
 - Connectwizard and network_default_connect accommodate connman alternate network manager.
 - The default ethernet interface (used when no network manager has been set up) can now be disconnected and reconnected with the network tray icon.
 - The default ethernet setup no longer uses ethtool to check for "plugged in", replaced by ip-show and a test for LOWER_UP.
 - The default ethernet setup waits 10 seconds longer for an IP address, which may reduce the incidence of failure (indicated by an address of 169.254...), possibly more likely if MAC address spoofing is used on a LAN.
 - All components have been checked with 'shellcheck' and have had its warnings resolved, making some of the code more efficient, simpler and less vulnerable to coding mistakes.


01micko (approved):

woof-code/rootfs-skeleton/etc/rc.d/rc.network_eth
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
```
dimkr: @rserwin1 Can you reopen this PR after removing all the whitespace/indentation changes? This code is very messy, and it's hard to review it when most changes are cosmetic.

rserwin1:
Instead of undoing some of my effort, I would prefer to submit another PR with only the indentation changes and, if helpful, the replacement of "`...`" with "$(...)".  You could merge that and then examine the remaining changes in the current PR.  OK?

dimkr:
That would make it easier to review the changes. But @01micko should decide whether or not to merge, I'm just an extra pair of eyes looking at the code.

rserwin1:
I sent dimkr a difference listing versus a similarly indented version of the current versions of the changed files.  If necessary, I can send the file sets for him to compare with, say, kdiff3.
In the meantime, shall I add a commit to make a minor change to network_default_connect?  The listing revealed a mistake in the indenting, which could be ignored if a new commit would delay the merging.  The change may not matter if/when we move these files to a new rootfs-package, or can be made at that time (if the connman code is retained).
Please tell me if you need anything more from me to allow the merge.

dimkr:
Even after hiding whitespace (https://github.com/puppylinux-woof-CE/woof-CE/pull/2848/files?diff=unified&w=1), I find this PR hard to review. I don't understand some changes, like the DHCP release thing.

rserwin1:
I assume you refer to the dhcpcd-related changes in rc.network_eth.
It is actually only one change, to increase the dhcpcd timeout value to wait for a DNS to respond with a valid IP address, but only if a timeout setting is not in /etc/dhcpcd.conf.

Those changes would be unnecessary if we simply add a line, "timeout 40", to dhcpcd.conf.  That would be the best over-all fix for the problem, which I will explain.

While testing my changes for SNS-3.3, I found that the MAC address spoofing implementation caused my router to get slower in responding to dhcpcd after rebooting the PC a few times.  My other PC on the same LAN/router also failed to connect due to the "169.254" responses.  Each reboot generates a new spoofed MAC address, possibly necessitating the router to check a longer MAC list.  If dhcpcd timed out it would return an invalid IP address, 169.254...  A day after the spoofing, the invalid IP addresses did not occur, probably because the leases had expired.

After implementing retrying dhcpcd once, which usually then succeeded, it struck me that the better solution is to increase the timeout value.  But I wanted to confine that to SNS, even though the slowdown impacts any device on the same LAN no matter which network manager is in control.

I now believe that this PR should include the "timeout 40" addition to dhcpcd.conf.  BTW, the "40" is arbitrary, one third longer than the default of 30 seconds.  That way the "--timeout" logic can be removed from rc.network_eth (and SNS 3.3).  I request permission to make those (non-SNS) changes under this PR.

Edit:
I see that changing dhcpcd.conf is more complicated than I anticpated.  It is in the dhcpcd package.  Since I think we should be changing the content of that package, is there a standard way to update a configuration file during a build or at installation or boot-up time?

rserwin1:
The "Update hostname-set" commit should resolve the conflict.  Please tell me if there is anything else I need to do to allow the merge.

dimkr:
> The "Update hostname-set" commit should resolve the conflict. Please tell me if there is anything else I need to do to allow the merge.

The conflict is still there

rserwin1:
I merged the conflicting line, "echo "::1 localhost $NEW_HOSTNAME" >> /tmp/hostname-set-hosts" into my changes.  The remaining "conflicts" are my updates to hostname-set, the indents and quotes around $HOSTNAME in two places.

I no longer see the message regarding a conflict, but forget how I actually found the original conflict message.

How do I notify github that the resolution is what I intend?

[edit] Now that I added a comment, I see the message.

[edit] I think I figured it out.  I edited the conflict listing and clicked on the "resolved" button.

dimkr:
You need to do a `git pull` (or `git fetch` + `git merge`), "merge" doesn't mean "take the same lines as in the other branch and putting them in my branch".

rserwin1:
I am afraid that I do not understand how to perform the "git pull (or git fetch + git merge)".  Please assist me with that, now that you know my intent.

[edit] I need to be away for a while, so can resume this conversaion later.  BTW, thank you for your prompt replies.

dimkr:
You'll need to do some basic git tutorial and learn how to use git, which is not the same as using the GitHub web UI to edit files.

rserwin1:
I believe I have pushed the conflict resolution and do not now see any evidence of a current conflict.

Is there anything else I need to do for you to merge this PR?

dimkr:
Now this branch has 234 commits and 158 changed files.

I don't use SNS, so I have no objections other than git-related ones.

rserwin1:
Only three of the commits are mine.  How did the others get added to my branch?

I did refresh my local repo testing branch several times while it held my branch.  Would the updates somehow get added to the branch instead of the testing branch?

What do I do now?  The only solution I can think of is to start over with a new branch and PR.  Is there any other way?

Also, I had to retrieve the remote copy of the branch to rebuild my local repo, which I had inadvertently deleted.  Maybe that pulled along the other commits.

rserwin1:
Thanks for your patience with my effort, here.

Since I am unable to correct the "additional commits" problem with the network_connect_20220202 branch, I would like to start again from the beginning.  Would my deleting that branch (in github) and supplying a new branch delete my commits without impacting the added ones?  Maybe the PR needs to be rejected and closed, as well.  I am reluctant to guess on my own how to proceed, so please instruct me.

I am considering a replacement PR with a commit for the updated connectwizard and "cosmetized" versions of the other components.  Once they are merged I would submit a new PR -- or an addition to the replacement PR, if possible after the merge -- with the full updates of the remaining components.

That way, the critical fix to connectwizard gets done and the other components will reveal (in github) the substantive changes in a second commit, without the annoying indentation and command indicator (`...` -> $(...)) differences.

Please direct me so that I can get this done.

dimkr:
> I did refresh my local repo testing branch several times while it held my branch. Would the updates somehow get added to the branch instead of the testing branch?

It doesn't work like that, you need to `git pull`.

dimkr:
@rserwin1 I did that for you, see https://github.com/puppylinux-woof-CE/woof-CE/pull/3024.

If you want to get stared with git quickly and stop using GitHub's web UI,

```
git clone http://github.com/puppylinux-woof-CE/woof-CE -b network_connect_20220202
cd woof-CE
git remote add fork http://github.com/rserwin1/woof-CE
git branch network_connect_20220202-more-changes
git checkout network_connect_20220202-more-changes
# do your changes
git add file1 file2
git commit -m "commit message goes here"
git pus fork network_connect_20220202-more-changes
# open a pull request from rserwin1:network_connect_20220202-more-changes to puppylinux-woof-CE:network_connect_20220202
```

rserwin1:
Thank you very much for cleaning up my original branch and giving me those instructions.

I have completed them except for:
open a pull request from rserwin1:network_connect_20220202-more-changes to puppylinux-woof-CE:network_connect_20220202

I assume that is to be a normal PR but with commentary that it supplements the original branch.  Before I do that, I want to be sure you don't mean more than that.

dimkr:
> I want to be sure you don't mean more than that.

Nope. Your changes are safe, they're in puppylinux-woof-CE/woof-CE:network_connect_20220202. Just open a PR with additional changes you want to add (don't forget to target that branch and not testing), and everything can be merged into testing together.

rserwin1:
I may have over-responded by committing all but hostname-set.  It appeared to me that connectwizard had not been changed in the original branch, so concluded that only hostname-set is there, with the resolved conflict.  I think I have changed only connectwizard and rc.network_eth from the original branch.

Will it be OK to submit the 9 and assume the unchanged files are handled or ignored?

dimkr:
> Will it be OK to submit the 9 and assume the unchanged files are handled or ignored?

I'm not sure I understand. The changes in the "Files changed" view are the ones that make it into the destination branch, after a pull request is merged. If you don't want some of the changes to be merged, you'll need to revert them.

https://github.com/puppylinux-woof-CE/woof-CE/pull/3024/files

rserwin1:
I double-checked my original branch in rserwin1 and see only hostname-set as having my changes and the conflict resolved.  The other do not have my changes.

So I will go  ahead and submit the commit of the 9 changed files.

Issue #2852, opened 2022-02-10
frontend_change is killed if desktop not provided by ROX-Filer - why? line 58


peabee:
@rizalmart this was your change #2583 

Not entirely sure why this has emerged as an issue.........but in my new build of LxPupSc using Slackware-15.0
line 58 of
https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-code/rootfs-skeleton/usr/local/pup_event/frontend_change

kills pup_event 
frontend_change 
if ROX-Filer is not providing the desktop (e.g. for LXDE PCManFM provides the desktop)

Can line 58 be deleted?? Why is it there??

01micko:
Why do you need it?

As far as I can tell, all the code in there is rox specific.

peabee:
Because I want frontend_change to run to completion when the desktop is LXDE..... so that the auto-launch handlers are run when new devices are detected.

01micko:
Ok, but I think the whole script needs a rewrite.

I'll approve #2852 but raise another issue.

PR #2856, opened 2022-02-11
All front_change to finish even if desktop is not ROX-Filer

peabee:
see #2852


01micko (approved):

woof-code/rootfs-skeleton/usr/local/pup_event/frontend_change
```diff
@@ -55,9 +55,6 @@ echo "$BPS" | grep -q 'ROX-Filer -p' && {
 	ROX_DESKTOP=yes
```
dimkr: This one is unused, no need to run `ps` anymore

PR #2854, opened 2022-02-11
Add GTK+ 3 support to Buntoo

dimkr:

PR #2853, opened 2022-02-10
Add the 605 global theme

dimkr:
@mrfricks Do you have the original wallpaper from 6.0.5 in SVG format?

PR #2857, opened 2022-02-11
Add a jammy64 template with 5.15.x, GTK+ 3 and PipeWire

dimkr:
Looks and feels like a normal Puppy, and has most of the goodies in Vanilla Dpup. Almost all under-the-hood stuff, like the urxvt->lxterminal wrapper, all the fixes that went into jwm-xdgmenu, fixmenusd (automatic menu reload + auto-spot for problematic applications), and jamesbond's Xdialog fork, are included; ~~the most notable omission is Xwayland~~ it's off by default.

If the switch to PipeWire doesn't happen in Ubuntu 22.04, PulseAudio works too and it's easy to switch back to it. Blueman, pavucontrol and pa-applet work just fine with PipeWire.

~~Test ISO with devx and bdrv is available at https://github.com/dimkr/woof-CE/releases/tag/upupj64-10.0-prepreprealpha3.~~

jammy4
jammy2
jammy3

apt works (bdrv is ~~20~~ 10? MB and #2846 works great!), but only if `USR_SYMLINKS=yes`:

jammy1

(`USR_SYMLINKS=yes` will break some old packages, but that's the cost of increased Ubuntu compatibility, and Debian 11 is the last to support the traditional directory layout, so it's time to embrace this change)

~~It would be nice to have a unique theme for this build, hence #2853 and #2854.~~ Now that ~~Buntoo~~ stark-blueish supports GTK+ 3 (#2854), it's the default theme.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,6 +35,10 @@ jobs:
             kernel: 5.10.x-x86_64
           - arch: x86_64
             compat-distro: ubuntu
+            compat-distro-version: jammy64
+            kernel: 5.15.x-x86_64
+          - arch: x86_64
+            compat-distro: ubuntu
             compat-distro-version: focal64
             kernel: 5.4.x-x86_64
           - arch: x86_64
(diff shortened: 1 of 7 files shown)


dimkr:
bdrv support is moved to #2886.

PR #2861, opened 2022-02-12
Modify buntoo wallpaper

01micko:


01micko (commented):
Tested working. Will fix `ptheme` soon to support `labwc`

PR #2862, opened 2022-02-12
Add Stardust* openbox themes

01micko:


01micko (commented):
this is fairly innocuous so mergeing

PR #2863, opened 2022-02-12
Move kernel-kit output to /usr

dimkr:
This PR addresses the concerns raised in https://github.com/puppylinux-woof-CE/woof-CE/pull/2445#issuecomment-904604701 and https://github.com/puppylinux-woof-CE/woof-CE/pull/2771#pullrequestreview-853536828.

Today, woof-CE can build a Puppy with the traditional directory layout, or with /bin, /sbin and /lib* symlinks to /usr/. But for this to work, 3builddistro needs to 1) rebuild fdrv+zdrv, after moving everything to /usr and 2) remove change_kernels, because it won't work unless somebody performed (1) on the new kernel.

The old directory layout is creating compatibility issues with Ubuntu and Debian packages (for example, see #2857), and Debian 11 is the last to officially support it. It's time to have proper support for it in Puppy, and clean up the mess.

This PR 1) makes kernel-kit output compatible with both the traditional and new directory layouts 2) removes the ugly unpacking+packing of fdrv+zdrv and 3) restores change_kernels, when using the new directory layout.

**However, this comes at a price: this PR makes kernel-kit output incompatible with any old Puppy** that doesn't have the rc.sysinit change.

I don't see a cleaner way to make kernel-kit output work for both directory layouts, using a single package.

### DIFF
--- a/kernel-kit/build.sh
+++ b/kernel-kit/build.sh
@@ -835,6 +835,7 @@ make DESTDIR=$CWD/output/${AUFS_UTIL_DIR} install
 		mv $CWD/output/${AUFS_UTIL_DIR}/usr/lib \
 			$CWD/output/${AUFS_UTIL_DIR}/usr/lib64
 	fi
+	mv $CWD/output/${AUFS_UTIL_DIR}/sbin $CWD/output/${AUFS_UTIL_DIR}/usr/
 	log_msg "aufs-util-${kernel_version} is in output"
 	#---
 	[ -z "$OLDPATH" ] || export PATH=$OLDPATH
@@ -855,7 +856,7 @@ else
 	export MAKE_TARGETS="bzImage modules"
 fi
 echo "$MAKE ${JOBS} ${MAKE_TARGETS}
-$MAKE INSTALL_MOD_PATH=${linux_kernel_dir} modules_install" > compile ## debug
+$MAKE INSTALL_MOD_PATH=${linux_kernel_dir}/usr modules_install" > compile ## debug
 
 log_msg "Compiling the kernel"
 $MAKE ${JOBS} ${MAKE_TARGETS} >> ${BUILD_LOG} 2>&1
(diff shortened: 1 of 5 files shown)


peabee (changes requested):
1. Not all Linux versions have moved to this (horrid) lib structure....?
2. should be an option in /kernel-kit/build.conf which structure to use
3. all LTS Woof-CE kernels should be built and released in both structures with a name to match

kernel-kit/build.sh
```diff
@@ -855,7 +856,7 @@ else
 	export MAKE_TARGETS="bzImage modules"
 fi
 echo "$MAKE ${JOBS} ${MAKE_TARGETS}
-$MAKE INSTALL_MOD_PATH=${linux_kernel_dir} modules_install" > compile ## debug
+$MAKE INSTALL_MOD_PATH=${linux_kernel_dir}/usr modules_install" > compile ## debug
```
01micko, line 859: You know I'm not against change but Slackware will likely never adopt the /usr merge.

What about this in `build.conf`

```cfg
## for best compatibility with debian, ubuntu, arch; set the UM variable
UM=/usr
## slackware, salix etc
# UM=''
```

Then all the lines in `build.sh` where `/usr` is required can be thus:

```
$MAKE INSTALL_MOD_PATH=${linux_kernel_dir}${UM} modules_install" > compile ## debug
```

dimkr:
I think you missed the entire point of this PR.

> 1. Not all Linux versions have moved to this (horrid) lib structure....?

True, but this PR makes it easy to support **both** the old and the new structure. Currently, kernel-kit supports only the old one, so users can't use change_kernels when they have the new one.

>     2. should be an option in /kernel-kit/build.conf which structure to use

IMHO that's a bad idea, because that means every kernel needs to be packaged twice. This can be confusing for users, who don't know which one to take. I prefer a single package that can be used with both layouts.

---

Here's an example from bionic64, built from this PR:

bionic

It uses the old layout (no symlinks) and kernel-kit output where everything is under /usr. It works fine, thanks to /lib/{modules,firmware} bind mounts, and this kernel package can be used unmodified, with both layouts.

kernel-kit/build.sh
dimkr: It's OK, see the change in rc.sysinit. It's possible to use the kernel-kit output with a Puppy that didn't adopt this change, using considerably less code than the ugliness I removed from 3builddistro.

dimkr:
Closed in favor of #2866.

PR #2866, opened 2022-02-12
Add /usr merge compatible kernel packages

dimkr:
This is an alternative to #2863 that doesn't upset anyone, but also doesn't address all concerns (the user must use the right package, otherwise the system won't boot).

This PR adds a second build step, which re-packages some kernels after moving everything to /usr:

~~Still testing.~~ Works great! All builds passed and every build uses the right kernel package.

matrix
artifacts

Then, the other jobs check `USR_SYMLINKS` to decide which build artifact to use: `kernel-kit-output-x-y` or `kernel-kit-output-usrmerge-x-y`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -102,6 +102,16 @@ jobs:
       run: |
         cd ../woof-out_*
         echo | sudo -E ./2createpackages
+    - name: Choose kernel variant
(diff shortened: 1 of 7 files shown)


dimkr:
#2870 doesn't work without this one, because the kernel sources SFS is incompatible. Merging.

PR #2867, opened 2022-02-13
Allow users to load modules built against an older longterm kernel

dimkr:
When building against kernel sources from an older 5.10.x:

```
~$ apt install broadcom-sta-dkms
...
Setting up dkms (2.8.4-3) ...
Setting up broadcom-sta-dkms (6.30.223.271-17) ...
Loading new broadcom-sta-6.30.223.271 DKMS files...
It is likely that 5.10.0 belongs to a chroot's host
Building for 5.10.0
Building initial module for 5.10.0
Done.

wl.ko:
Running module version sanity check.
 - Original module
   - No original module exists within this kernel
 - Installation
   - Installing to /lib/modules/5.10.0/updates/dkms/

depmod.....

DKMS: install completed.
```

```
~$ modprobe wl
modprobe: ERROR: could not insert 'wl': Exec format error
~$ dmesg | tail -n 1
[50315.182888] wl: disagrees about version of symbol module_layout
```

With a kernel built from this PR, `modprobe wl` works although the kernel sources SFS is older.


mavrothal:
If this implies that any module can be installed or force-installed in a puppy long-term kernel, looks like a very poor (and unsafe) idea.
Why would you want to do that?

dimkr:
This makes it possible to build wl.ko against 5.10.99, and load it with 5.10.100. This is safe.

What isn't safe, is trying to load a module built against 5.10.99 when running 5.16.9. But that wont' happen, because the module path is different, and the user who tries to force it is knowingly doing something stupid and dangerous.

mavrothal:
I still do not see what is the problem you try to solve 
Why would someone want to load a kernel module from another kernel, instead of recompiling.
Different kernel - different modules.
Omitting this check opens a can of warms to stupidity or worse.

dimkr:
> Why would someone want to load a kernel module from another kernel, instead of recompiling.

To avoid recompiling, that's the answer. Puppy doesn't support dkms like Debian, so you have to recompile after each kernel update.

mavrothal:
I would think that fixing support for dkms or recompiling is a way better solution than allowing a "free for all" in module loading.

PR #2875, opened 2022-02-15
Rename "605" to "Tahrpup" and give it a nicer wallpaper

dimkr:
The current one is an ugly placeholder, until we have something that 1) resembles the 6.0.5 one and 2) has better contrast, unless we make desktop icon labels black somehow and 3) works with today's flat icons.

Old:

New:

Together with the desktop icon labels, GTK+ 2 (ROX-Filer) and GTK+ 3 (Geany and Transmission):


mrfricks:
i do prefer your other version (with the filled in vincents) as a picture but this one is less 'in your face' and maybe better suited for a wallpaper

dimkr:
@01micko @mrfricks Objections to this PR?

01micko:
I'm ok with it but really @mrfricks holds the trump card.

mrfricks:
no objections from me

PR #2878, opened 2022-02-16
Backport puppylinux-woof-CE/gtkdialog#126

dimkr:
This is temporary, until the next gtkdialog is out. Vanilla Dpup already has this fix backported, but it's always good to have consistency across Puppy releases.


01micko (approved):
Makes sense since it is a bugfix

PR #2880, opened 2022-02-16
Update the firewall tray icon when /etc/init.d/rc.firewall runs

dimkr:
It runs `iptables` every 2s! I wonder how this affects battery life. Instead, we can poll very infrequently (say, once in 10m), just in case the user flushes the iptables rules manually, and rely on rc.firewall to ask the tray icon to update.

### DIFF
--- a/woof-code/rootfs-petbuilds/firewallstatus/firewallstatus-0.7/firewallstatus.c
+++ b/woof-code/rootfs-petbuilds/firewallstatus/firewallstatus-0.7/firewallstatus.c
@@ -6,6 +6,7 @@
 #include <gdk/gdk.h>
 #include <gdk/gdkkeysyms.h>
 #include <glib/gstdio.h>
+#include <glib-unix.h>
 #define THIS_VERSION "0.7"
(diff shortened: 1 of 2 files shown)


woof-code/rootfs-petbuilds/firewallstatus/firewallstatus-0.7/firewallstatus.c
```diff
@@ -21,8 +22,11 @@ void quit(GtkWidget *w, gpointer dummy);
 void showme_window(GtkWidget *w, gpointer dummy);
 
 unsigned int interval = 2000; /*update interval in milliseconds */
+unsigned int long_interval = 600*1000;
+unsigned int max_tries = 30;
```
dimkr, line 26: After polling the current status 30 times (once every 2s), we start polling every 10m. We assume that 1) rc.firewall will notify us if something has changed 2) it takes up to 1 minute for the firewall to start on boot and 3) users start and stop the firewall via rc.firewall.

01micko (approved):

PR #2884, opened 2022-02-18
Add command-not-found support to bdrv

dimkr:
vlc

This works nicely with shell completion:

abiword

And it starts working immediately after the first `apt update`, so users who don't use apt won't be bothered by it:

update

### DIFF
--- a/woof-code/rootfs-skeleton/root/.bashrc
+++ b/woof-code/rootfs-skeleton/root/.bashrc
@@ -1,2 +1,12 @@
 . /etc/profile
 [ -f /etc/bash_completion ] && . /etc/bash_completion
+if [ -x /usr/lib/command-not-found ]; then
+ command_not_found_handle() {
+  if [ -f /var/lib/command-not-found/commands.db ]; then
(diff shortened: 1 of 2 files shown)


woof-code/rootfs-skeleton/root/.bashrc
```diff
@@ -1,2 +1,12 @@
 . /etc/profile
 [ -f /etc/bash_completion ] && . /etc/bash_completion
+if [ -x /usr/lib/command-not-found ]; then
+ command_not_found_handle() {
+  if [ -f /var/lib/command-not-found/commands.db ]; then
```
dimkr, line 5: This file is created after `update-command-not-found`, and the first `apt update` triggers it. If `command-not-found` is used before `update-command-not-found` runs for the first time, it prints an error message. We don't want this error, because some users don't use apt at all and won't ever `apt update`. And we want to check if /var/lib/command-not-found/commands.db exists every time the user types an unknown command, because the use may `apt update` for the first time, then install the wrong package.

01micko (approved):

PR #2893, opened 2022-02-20
Make pupswap work under PUPMODE 6

dimkr:
It works, but not if the swap file is under /, and that's the default.


mavrothal (approved):
Now correctly writes swap file in /initrd instead of /

woof-code/3builddistro
```diff
@@ -897,6 +897,7 @@ fi
 if [ "$FRUGALIFY" ]; then
 	echo "PUPMODE='6'" > rootfs-complete/etc/rc.d/PUPSTATE
 	echo "PUNIONFS='overlay'" >> rootfs-complete/etc/rc.d/PUPSTATE
+	echo "PUP_HOME=''" >> rootfs-complete/etc/rc.d/PUPSTATE
```
mavrothal, line 900: If/when frlugalify runs with aufs instead of layerfs sfs_load may have problem with this

PR #2896, opened 2022-02-23
Add support for extra processing defined in _00build.conf

peabee:
e.g. for devx manipulation

### DIFF
--- a/woof-code/3builddistro
+++ b/woof-code/3builddistro
@@ -332,6 +332,9 @@ if [ -n "$PETBUILDS" ]; then
 		echo "Building without qemu-user-static"
 	fi
 
+	# set EXTRA_PROCESSING in _00build.conf to perform any processing required - e.g. for devx
+	[ $EXTRA_PROCESSING ] && . $EXTRA_PROCESSING
+
 	cd sandbox3
 	. ../support/petbuilds.sh
 	cd ..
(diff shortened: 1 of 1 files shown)


woof-code/3builddistro
```diff
@@ -332,6 +332,9 @@ if [ -n "$PETBUILDS" ]; then
 		echo "Building without qemu-user-static"
 	fi
 
+	# set EXTRA_PROCESSING in _00build.conf to perform any processing required - e.g. for devx
+	[ $EXTRA_PROCESSING ] && . $EXTRA_PROCESSING
```
01micko: This test _can_ (not saying it is) be ambiguous without a string comparison operator, `-n` for non-zero string length, `-z` for zero length.

01micko (changes requested):

woof-code/3builddistro
dimkr: And there's a syntax error if it's undefined, you need to wrap `$EXTRA_PROCESSING` with quotes

woof-code/3builddistro
```diff
@@ -332,6 +332,9 @@ if [ -n "$PETBUILDS" ]; then
 		echo "Building without qemu-user-static"
 	fi
 
+	# set EXTRA_PROCESSING in _00build.conf to perform any processing required - e.g. for devx
```
dimkr: Why is this manipulation of devx done only if PETBUILDS is non-empty? I'd do it earlier, at line 306, https://github.com/puppylinux-woof-CE/woof-CE/blob/d5b9484598f2cfa28328ce841eb4d8be6e1d43c2/woof-code/3builddistro#L306

peabee:
devx was only an e.g. so requested changes done but at line 312.

01micko (approved):

PR #2899, opened 2022-02-24
Show only one menu entry for each .desktop file

dimkr:
`apt install -y gimp vlc`

Before:

gimp

After:

gimp2

Before:

vlc

After:

vlc2

Some applications appear in more than one category, and that's fine IMHO. For example, Wireshark appears under the monitoring submenu of System, and under Network.


01micko (changes requested):
Similar needs to be done for labwc.
menu

dimkr:
> Similar needs to be done for labwc. menu

I'll do this in a separate PR, ~~the labwc menu needs other fixes first.~~ done, see #2902.

01micko:
Ok, merge then :)

Issue #2903, opened 2022-02-24
Different scenarios for X and Wayland - fix Puppy scripts


01micko:
There are several files in Puppy that need fixing for `wayland`, both pure wayland and xwayland.

For a start:

- [x]  quicksetup (and associated links) #2924
- [x] xkbconfigmanager #2904
- [ ] ptheme
- [x] xorgwizard #2905 
- [x] box_splash #2911 
- [x] report-video #2906 

plus likely many more.

There are several environment variables that exist that are relevant to these scripts and several different scenarios

Relevant environment vars:
- DISPLAY
- WAYLAND_DISPLAY
- XDG_SESSION_TYPE

Scenarios:
- Pure X; traditional Pups using X.org display server
- Xwayland under `cage`; eg: @dimkr 's `vanilla dpup`
- Xwayland with `labwc` (or other wayland compositor such as `sway` or `wayfire`)
- pure Wayland with `labwc` or other wayland compositor

The problem is that in each of the scenarios above, the values of the above env vars are either different or deliberately unset.

Each Scenario:

#### Pure X

The main variable that is always set when `X` is running is `DISPLAY`. It's value can vary but that doesn't matter in the context of the relevant scripts. It is possible that `XDG_SESSION_TYPE` could be set but in Puppy isn't. `WAYLAND_DISPLAY` is certainly not set.

#### X under Cage

Since `cage` is a wayland display server `WAYLAND_DISPLAY` is set, and as `X` is running under `cage` `DISPLAY` is set (this is also true for Xwayland) and `XDG_SESSION_TYPE` is set to `x11`

#### labwc with xwayland support

`labwc` is a wayland display server so `WAYLAND_DISPLAY` is set, since there is `Xwayland` support `DISPLAY` is set and naturally the `XDG_SESSION_TYPE` var is set to `wayland`

#### labwc with pure wayland support

`labwc` is a wayland display server so `WAYLAND_DISPLAY` is set, since there is __no__ `Xwayland` support `DISPLAY` is __unset__ and naturally the `XDG_SESSION_TYPE` var is set to `wayland`

#### Side note

[...]

PR #2904, opened 2022-02-25
labwc: update keyboard configuration re: #2903

01micko:
This adds labwc support to /usr/sbin/xkbconfigmanager
Also, it fixes a bug where if a variant doesn't exist for a layout
the Xdialog gui crashes because of no input. Instead a message pops saying
that no variant is available for the layout. An example is au (Australia).

@dimkr can you just check that it doesn't mess with the `cage` settings, I think it should be fine.

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/xkbconfigmanager
+++ b/woof-code/rootfs-skeleton/usr/sbin/xkbconfigmanager
@@ -7,6 +7,7 @@ export OUTPUT_CHARSET=UTF-8
 
 export XORG_CONF=/etc/X11/xorg.conf
 export XWAYLANDRC=~/.Xwaylandrc
+export LABWCRC=$XDG_CONFIG_HOME/labwc/environment
 
 if [ -f /etc/X11/xkb/rules/xorg.lst ] ; then
 	XORG_RULES_LST='/etc/X11/xkb/rules/xorg.lst'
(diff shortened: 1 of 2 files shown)


woof-code/rootfs-skeleton/usr/sbin/xkbconfigmanager
```diff
@@ -492,53 +529,60 @@ NOTE: if you select a variant for a layout that already has one,
...
-		# add to file
-		sed -i "s%$ORIGINAL%$MODIFIED%" $XORG_CONF
-		if grep -q '"XkbModel"' $XORG_CONF ; then
-			XkbApplyNow
-		else
-			eval $ERRMSG
+	
+		if [ -f $XWAYLANDRC ]; then
+			sed -i "s%^XKB_DEFAULT_VARIANT=.*%XKB_DEFAULT_VARIANT=$NEW%" $XWAYLANDRC
+			[ -n "$WAYLAND_DISPLAY" ] && XkbApplyLater
```
dimkr, line 578: XkbApplyLater should be called only once. I don't think the use case of having both labwc and Cage is useful, so it's OK not to fix this.

01micko: On condition it's called either if $XWAYLANDRC or $LABWCRC exists, and as you allude to, it's unlikely that cage and labwc will coexist.

PR #2915, opened 2022-02-26
Add the -updates repo only for $DISTRO_BINARY_COMPAT:$DISTRO_COMPAT_VERSION

dimkr:
@peabee This will allow you to keep doing builds that mix Ubuntu and Debian repos


peabee (approved):
Thanks.

PR #2917, opened 2022-02-26
Ignore /etc/profile.d/*.csh

dimkr:
Ubuntu has these, and they're not meant to be sourced by bash. They cause syntax errors in every new terminal tab or window.

### DIFF
--- a/woof-code/rootfs-skeleton/etc/profile
+++ b/woof-code/rootfs-skeleton/etc/profile
@@ -90,7 +90,7 @@ export PREFIX='/usr' #convenient to set this i think...
 # Append any additional sh scripts found in /etc/profile.d/:
 #for profile_script in /etc/profile.d/*.sh ; do
 for profile_script in /etc/profile.d/* ; do #w482 any files.
-	case "$profile_script" in *txt) continue ;; esac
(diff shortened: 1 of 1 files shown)


01micko:
Yes, saw that when I tried vanillaupup in Qemu.

01micko (approved):

rizalmart:
> Ubuntu has these, and they're not meant to be sourced by bash. They cause syntax errors in every new terminal tab or window.

Much better if check csh/ksh/zsh executable first before ignoring it

01micko:
> > Ubuntu has these, and they're not meant to be sourced by bash. They cause syntax errors in every new terminal tab or window.
> 
> Much better if check csh/ksh/zsh executable first before ignoring it

In the context of mainstream puppy we can leave that up to the 3rd party developer. We normally only include `/bin/ash`, `/bin/bash` and `/bin/sh` linked to `/bin/bash` and falls back to `/bin/ash` if bash is not installed.

dimkr:
> Much better if check csh/ksh/zsh executable first before ignoring it

/etc/profile is bash-centric, it needs to be refactored to support other shells.

PR #2922, opened 2022-02-27
quicksetup: make it work in pure wayland and improve gtk3 performance

01micko:
re: #2903

This is a big commit so I will provide the file for testing in a zip

Changes:
- fix some actions on 'OK' press
- suppress 'xset' under wayland
- test for DISPLAY and WAYLAND_DISPLAY for GUI
- add test for numlockt (alternative to numlockx)
- fix LANG=C bug asking for translation
- add 'change root password' checkbox
- fix numerous logic bugs
- remove comboboxtext in favour of a button to launch a 'tree' widget
for locale, timezone, keyboard and crdw choices. It is much better under
gtk3, fixes quicksetup under wayland, and works great in gtk2
- remove redundant, useless stuff

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/quicksetup
+++ b/woof-code/rootfs-skeleton/usr/sbin/quicksetup
@@ -70,10 +70,9 @@ PARAM1=""
 [ $1 ] && PARAM1="$1"
 PARAM2=""
 [ $2 ] && PARAM2="$2"
-TABCHAR="`echo -n -e '\t'`"
 
-[ "$PARAM1" = "composeonly" ] && unset DISPLAY
-[ "$PARAM1" = "cli" ] && unset DISPLAY
+[ "$PARAM1" = "composeonly" ] && unset DISPLAY && unset WAYLAND_DISPLAY
+[ "$PARAM1" = "cli" ] && unset DISPLAY && unset WAYLAND_DISPLAY
 
 MYBASE=$(basename $0)
 WINICON="gtk-preferences"; HELPICON="gtk-index" #gtk-info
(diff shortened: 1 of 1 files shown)


01micko:
As promised, attached is a zip file

[qs.zip](https://github.com/puppylinux-woof-CE/woof-CE/files/8148593/qs.zip)

Try it in a running Pup, or use it to build. Easier than reviewing, but **by all means** review this if yo're up for it.

peabee (changes requested):
There is no way to select the chosen value in the multi-choice drop down lists - no "Select" or "OK" button........

dimkr:
One of the things I have about quicksetup is the huge number of features in one thing. It will take time to test every single feature in each scenario (X, Xwayland and labwc) 😠

But it's great to finally fix all these issues! It's a mess.

01micko:
> There is no way to select the chosen value in the multi-choice drop down lists - no "Select" or "OK" button........

There is an OK button at the bottom! Or does it go off screen for you? I can reduce the height.

EDIT:

~I can add the `button-release-event` action, that way there is no need of an OK button. Just clicking an item will trigger the change. I can add blurb in the dialog at the top. Would you prefer that?~

Seems not possible using `<action function="launch">`

peabee:
No sign of any button here......
And "Enter" does not work so any hidden button must not have focus.....
Tried double click of entry as well - that also didn't work

01micko:
This is in gtk2. There are _no_ hidden buttons. In fact this isn't the version you have because I made the OK button huge

Try it. (attached)

[qs2.zip](https://github.com/puppylinux-woof-CE/woof-CE/files/8149822/qs2.zip)



qs2-3
.

01micko:
I messed something up in my repo, fixed in new PR

PR #2924, opened 2022-02-28
quicksetup:bugfix and improve gtk3, wayland performance

01micko:
re #2903

see #2922

The latest file
[qs3.zip](https://github.com/puppylinux-woof-CE/woof-CE/files/8149921/qs3.zip)

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/quicksetup
+++ b/woof-code/rootfs-skeleton/usr/sbin/quicksetup
@@ -70,10 +70,9 @@ PARAM1=""
 [ $1 ] && PARAM1="$1"
 PARAM2=""
 [ $2 ] && PARAM2="$2"
-TABCHAR="`echo -n -e '\t'`"
 
-[ "$PARAM1" = "composeonly" ] && unset DISPLAY
-[ "$PARAM1" = "cli" ] && unset DISPLAY
+[ "$PARAM1" = "composeonly" ] && unset DISPLAY && unset WAYLAND_DISPLAY
+[ "$PARAM1" = "cli" ] && unset DISPLAY && unset WAYLAND_DISPLAY
 
 MYBASE=$(basename $0)
 WINICON="gtk-preferences"; HELPICON="gtk-index" #gtk-info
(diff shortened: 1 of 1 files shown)


peabee:
As you can see ............ no button ............. from qs3.zip
Screenshot

01micko:
That _is_ strange. What is your setup? An iso download to test?


Edit: that is not qs3 - downloaded a and checked.

This is what it looks like 

20220228_19h50m55s_grim

dimkr:
For some reason, the font is ugly and there's no background (theme bugs in stark-bluish and buntoo-ambience?):

white

So far, everything I tested under X.Org seems to work.

dimkr:
There are issues with Xwayland: 1) ~/.Xwaylandrc is not changed if xorg.conf exists, and this can happen if the user switches away from X.Org and 2) the `sed` line is malformed.

xway

woof-code/rootfs-skeleton/usr/sbin/quicksetup
```diff
@@ -1181,33 +1315,45 @@ FLAG_CHANGED=""
...
- FLAG_CHANGED="${FLAG_CHANGED}$(gettext 'Keyboard Layout'):yes|" #111020 no need to restart X.
  if [ -f /etc/X11/xorg.conf ]; then
   sed -i "s%.*Option.*XkbLayout.*%	Option      \"XkbLayout\" \"${XKB_LAYOUT}\"%" /etc/X11/xorg.conf
- fi
- if [ -f ~/.Xwaylandrc ]; then
-   sed -i "s%^XKB_DEFAULT_LAYOUT=.*%XKB_DEFAULT_LAYOUT=$XKB_LAYOUT%" ~/.Xwaylandrc
- fi
- if [ -f "$XDG_CONFIG_HOME/labwc/environment" ]; then
+  FLAG_CHANGED="${FLAG_CHANGED}$(gettext 'Keyboard Layout'):yes|" #111020 no need to restart X.
+ elif [ -f ~/.Xwaylandrc ]; then
```
dimkr, line 1324: This one should be `if`, not `elif`. For each configuration file, quicksetup should check if it exists, and modify it. This will keep the settings synced, if the user moved from X.Org to Xwayland, or from X.Org to labwc. And some users keep old save files, so xorg.conf can be there while it's a labwc-based Puppy.

01micko:
I've had a long hard think about this and `quicksetup` should go. It's 10+ years old and had crap bolted on all over it. It wasn't that good to begin with. Many of the bugs I have exposed are hidden by the `comboboxtext` widget because it doesn't matter if the `<default>` tag is empty or wrong. The default has always been wrong all these years!

Anyway, I'm abandoning this PR and letting it rot. A script should do one thing and do it well; I'm sure Ken and Dennis would agree, no?. This does several poorly. It's slow (sed, this cut, grep sed that, FFS! `onPATTERN` [WTF?]) and cumbersome. Most of that can be replaced with bashisms, or even POSIX. Much faster and clean, easier to read. 

I think each separate part should be it's own script, just like how it _was_ in 4.1.2 with of course a contemporary approach.

Sorry to have wasted your time (and mine :roll_eyes: ) but it's time to throw away 1000+ lines of (er) code.

dimkr:
> just like how it _was_ in 4.1.2

4.1.2 ❤️

PR #2935, opened 2022-03-01
If not using ROX-Filer, replaces the Drives menu with pmount

dimkr:
There's no easy to unmount a drive without the ROX-Filer desktop icons. Therefore, it makes more sense to put a menu entry that leads to some tool that allows one to mount, unmount and browse partitions.

pmount

@01micko In Vanilla Dpup 9.2.x, I want to get rid of ROX-Filer. The combination of JWM and pcmanfm should be a good stepping stone towards a pure Wayland desktop with pcmanfm. Without the drives menu, we have a pretty static menu layout and it should be much easier to build a 1:1 equivalent of this jwm+pcmanfm combo, using labwc+sfwbar+pcmanfm.


peabee (commented):
This change only applies to VanillaDPup so I have no views on it.
However, it sounds like an inferior implementation compared to retaining Rox-Filer in the build. Rox can happily co-exist with PcManFM and indeed has some functionality that is superior to PcManFM - particuarly in the creation and monitoring of symlinks.

01micko (approved):

01micko:
I have gtkdialog code that solves this - later though :wink:

dimkr:
> This change only applies to VanillaDPup so I have no views on it.

It will affect any user that uses JWM but disables the ROX-Filer pinboard, too.

And keeping ROX-Filer is not an option, because the entire point of this PR (and others, like #2934), is to make it possible to use Puppy comfortably, without ROX-Filer.

PR #2948, opened 2022-03-05
If only Xwayland is present, start Xwayland and hide the X.Org button

dimkr:
X.Org is much bigger than Cage+Xwayland, and I'd like to try to build a Puppy without X.Org.

### DIFF
--- a/woof-code/rootfs-skeleton/usr/bin/xwin
+++ b/woof-code/rootfs-skeleton/usr/bin/xwin
@@ -31,11 +31,14 @@ fi
 
 if [ -f /var/local/xwin_enable_xwayland ] ; then
 	XWAYLAND=1
+elif [ -e /usr/bin/startxwayland -a ! -e /usr/bin/Xorg ]; then
+	XWAYLAND=1
+	touch /var/local/xwin_enable_xwayland
 else
 	XWAYLAND=0
 fi
 
-if [ $XWAYLAND -eq 0 -a `id -u` -eq 0 -a -h /usr/bin/X ] ; then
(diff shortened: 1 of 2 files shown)


woof-code/rootfs-skeleton/usr/bin/xwin
```diff
@@ -31,11 +31,14 @@ fi
 
 if [ -f /var/local/xwin_enable_xwayland ] ; then
 	XWAYLAND=1
+elif [ -e /usr/bin/startxwayland -a ! -e /usr/bin/Xorg ]; then
+	XWAYLAND=1
+	touch /var/local/xwin_enable_xwayland
```
dimkr, line 36: If the user runs `apt install xserver-xorg`, we want to show the "switch to X.Org" button, but don't want to automatically switch to X.Org and cause disruption, because xserver-xorg could be a dependency

woof-code/rootfs-skeleton/usr/sbin/xorgwizard
```diff
@@ -699,11 +699,11 @@ fi
 
 
 xXSERVER=""
-if [ -f /usr/bin/startxwayland -a ! -f /var/local/xwin_enable_xwayland ]; then
+if [ -e /usr/bin/Xorg -a -f /usr/bin/startxwayland -a ! -f /var/local/xwin_enable_xwayland ]; then
```
dimkr, line 702: The button should appear only if both are installed

01micko (approved):

CollaboratorGCM:
+1

PR #2949, opened 2022-03-05
When building an Ubuntu bdrv, use the default apt settings

dimkr:
As a more desktop-oriented or newbie-oriented distro, we want `apt install` to act the same way as it does on Ubuntu. Some packages work better when the suggested or recommended packages are installed: for example, some packages recommend VAAPI drivers, and without them, some applications are slower and more power hungry.

The advanced user can always pass `--no-install-recommends` to avoid the installation of optional dependencies.


01micko (approved):
Makes sense to me.

PR #2950, opened 2022-03-05
If sudo is not installed, provide a dummy sudo

dimkr:
The internet is full of instructions that ask users to `sudo apt install`, but Puppy has no sudo. So let's provide a dummy sudo that works only through an interactive shell and doesn't break anything!

sudo


01micko (approved):

dimkr:
@01micko Just pushed a fix - have you seen it?

01micko:
Yes

dimkr:
@01micko This guy's `sudo apt update` worked :)

https://www.youtube.com/watch?v=wnL5ujlpVfA

image

PR #2954, opened 2022-03-06
installpkg.sh: package bugfixes + usrmerge

### LINKED ISSUE #3006: older pets and SFS with usrmerge puppies
Here is a couple of diffs that would allow old pets to be safely installed in usr-merged puppies and block dangerous SFSs to be installed.
I'm sure you can get better solutions but till then...
```
--- a/usr/local/petget/installpkg.sh	2022-03-27 03:17:23.000000000 +0800
+++ b/usr/local/petget/installpkg.sh	2022-03-31 21:15:53.000000000 +0800
@@ -277,13 +277,41 @@
    pPATTERN="s%^\\./${DLPKG_NAME}%%"
    echo "$PETFILES" | sed -e "$pPATTERN" -e "s#^\.\/#\/#g" -e "s#^#\/#g" -e "s#^\/\/#\/#g" -e 's#^\/$##g' -e 's#^\/\.$##g' > /var/packages/${DLPKG_NAME}.files
    install_path_check
-   tar -x --force-local --strip=2 --directory=${DIRECTSAVEPATH}/ -f ${tarball} #120102. 120107 remove --unlink-first
+   if [ -L /bin -o -L /sbin -o -L /lib ]; then
+    mkdir -p /root/${DLPKG_NAME}/
+    tar -x --force-local --strip=2 --directory=/root/${DLPKG_NAME}/ -f ${tarball}
+    cd /root/${DLPKG_NAME} 
+    [ -d bin ] && mkdir -p usr && mv bin usr/
+    [ -d sbin ] && mkdir -p usr && mv sbin usr/
+    [ -d lib ] && mkdir -p usr && mv lib usr/
+    cp -aR * ${DIRECTSAVEPATH}/
+    cd ~ ; rm -rf /root/${DLPKG_NAME}
+    sed -i 's/^\/bin/\/usr\/bin/g' /var/packages/${DLPKG_NAME}.files
+    sed -i 's/^\/sbin/\/usr\/sbin/g' /var/packages/${DLPKG_NAME}.files
+    sed -i 's/^\/lib/\/usr\/lib/g' /var/packages/${DLPKG_NAME}.files
+   else
+    tar -x --force-local --strip=2 --directory=${DIRECTSAVEPATH}/ -f ${tarball} #120102. 120107 remove --unlink-first
+   fi
   else
    #new2dir and tgz2pet creates them this way...
    pPATTERN="s%^${DLPKG_NAME}%%"
    echo "$PETFILES" | sed -e "$pPATTERN" -e "s#^\.\/#\/#g" -e "s#^#\/#g" -e "s#^\/\/#\/#g" -e 's#^\/$##g' -e 's#^\/\.$##g' > /var/packages/${DLPKG_NAME}.files
    install_path_check
-   tar -x --force-local --strip=1 --directory=${DIRECTSAVEPATH}/ -f ${tarball} #120102. 120107. 131122
+   if [ -L /bin -o -L /sbin -o -L /lib ]; then
+    mkdir -p /root/${DLPKG_NAME}/
+    tar -x --force-local --strip=1 --directory=/root/${DLPKG_NAME}/ -f ${tarball}
+    cd /root/${DLPKG_NAME} 
+    [ -d bin ] && mkdir -p usr && mv bin usr/
+    [ -d sbin ] && mkdir -p usr && mv sbin usr/
+    [ -d lib ] && mkdir -p usr && mv lib usr/
+    cp -aR * ${DIRECTSAVEPATH}/
+    cd ~ ; rm -rf /root/${DLPKG_NAME}
+    sed -i 's/^\/bin/\/usr\/bin/g' /var/packages/${DLPKG_NAME}.files
+    sed -i 's/^\/sbin/\/usr\/sbin/g' /var/packages/${DLPKG_NAME}.files
+    sed -i 's/^\/lib/\/usr\/lib/g' /var/packages/${DLPKG_NAME}.files
+   else
+    tar -x --force-local --strip=1 --directory=${DIRECTSAVEPATH}/ -f ${tarball} #120102. 120107 remove --unlink-first
+   fi
   fi
   rm -f "${tarball}"
   [ $? -ne 0 ] && clean_and_die
```

```
--- a/usr/sbin/sfs_load	2022-03-27 03:17:23.000000000 +0800
+++ b/usr/sbin/sfs_load	2022-03-31 21:16:14.000000000 +0800
@@ -47,6 +47,30 @@
 
 #exec &>/tmp/sfsl.log ; set -x
 
+# check if SFS is compatible with "usrmerge" filesystem
+
+inform_and_exit() {
+	/usr/lib/gtkdialog/box_ok "sfs_load" error "$(gettext "this sfs should not be loaded because it has the wrong file structure")"
+	unount /mnt/"$SFSFILE"
+	rm -rf /mnt/"$SFSFILE"
+	exit 1
+}
+
+if [ -L /bin -o -L /sbin -o -L /lib ]; then
+ if [ "$1" ]; then
+  SFSFILE="$(basename "$1")"
+  mkdir -p /mnt/"$SFSFILE"
+  DEVLOOP="$(losetup -f)"
+  losetup "$DEVLOOP" "$1"
+  mount -t squashfs -o loop,ro "$DEVLOOP" /mnt/"$SFSFILE"
+  [ -d /mnt/"$SFSFILE"/bin -a ! -L /mnt/"$SFSFILE"/bin ] && inform_and_exit || continue
+  [ -d /mnt/"$SFSFILE"/sbin -a ! -L /mnt/"$SFSFILE"/sbin ] && inform_and_exit || continue
+  [ -d /mnt/"$SFSFILE"/lib -a ! -L /mnt/"$SFSFILE"/lib ] && inform_and_exit || continue
+  unount /mnt/"$SFSFILE"
+  rm -rf /mnt/"$SFSFILE"
+ fi
+fi 
+
 . /etc/rc.d/PUPSTATE
 if [ "$PUNIONFS" = "overlay" ] ; then
 	exec sfs_load.overlay "$@"
```

rizalmart:
Fix a critical bug where the package file list got deleted if the new packages file and existing package file having the same filename.

### DIFF
--- a/woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
+++ b/woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
@@ -813,7 +813,9 @@ if [ "$xpkgname" != "" ] && [ "$installed_pkg" != "" ]; then
 	 fi
 	done < /var/packages/${installed_files}.files
 	
-	rm -f /var/packages/${installed_files}.files
+	if [ "${DLPKG_NAME}.files" != "${installed_files}.files" ]; then
+	 rm -f /var/packages/${installed_files}.files
+	fi
 	
 	grep -v "$installed_pkg" /var/packages/user-installed-packages > /var/packages/user-installed-packages.tmp
 	echo "$DB_ENTRY" >> /var/packages/user-installed-packages.tmp
--- a/woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
+++ b/woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
@@ -737,6 +737,7 @@ fi
 
 xpkgname="$(echo "$DB_ENTRY" | cut -f 2 -d '|')"
 installed_pkg="$(grep -m 1 "|$xpkgname|" /var/packages/user-installed-packages)"
+installed_files="$(echo "$installed_pkg" | cut -f 1 -d '|')"
 
 PKGDEP="$(echo "$DB_ENTRY" | cut -f 9 -d '|')"
 PKGDESC="$(echo "$DB_ENTRY" | cut -f 10 -d '|')"
@@ -787,19 +788,31 @@ fi
 
 #If there is an already installed package, just update the package files list and its database entry
 
+if [ "$xpkgname" != "" ]; then
+ if [ "$installed_pkg" != "" ] || [ -e "/var/packages/builtin_files/${xpkgname}" ]; then
+  #Build file database first excluding packages involved in upgrade/downgrade
+  ls /var/packages/builtin_files | grep -v '^'$xpkgname'$' | xargs -i cat '/var/packages/builtin_files/{}' | uniq > /tmp/pkg-files.list.tmp
+  ls /var/packages/*.files | grep -v '/'${DLPKG_NAME}.files'$' | grep -v '/'${installed_files}.files'$' | xargs -i cat '{}' | uniq >> /tmp/pkg-files.list.tmp
+  cat /tmp/pkg-files.list.tmp | uniq > /tmp/pkg-files.list
+  rm -f /tmp/pkg-files.list.tmp
+ fi
+fi
+
+
 PKGUPDOWN=""
 
 if [ "$xpkgname" != "" ] && [ "$installed_pkg" != "" ]; then
   #There is an already installed package. Just update the package file list
-  installed_files="$(echo "$installed_pkg" | cut -f 1 -d '|')"
+  
   if [ "$installed_files" != "" ]; then
    #Check if the old package file list exists
     if [ -e /var/packages/${installed_files}.files ]; then
     
 	while IFS= read -r xline
 	do
-	 #Check if the file was a part of the newly installed package
-  	 if [ "$(grep -m 1 "$xline" /var/packages/${DLPKG_NAME}.files)" == "" ]; then
+	
+	 #Check if the file was a part of the newly installed package and also if the file was not part of other installed packages
+  	 if [ "$(grep -m 1 "$xline" /var/packages/${DLPKG_NAME}.files)" == "" ] && [ "$(grep -m 1 "$xline" /tmp/pkg-files.list)" == "" ]; then
 	  #Not a part of newly installed package. Do action
 
 	   #Delete the file which is not a part of upgrade
@@ -840,6 +853,7 @@ else
  echo "$DB_ENTRY" >> /var/packages/user-installed-packages
 fi
 
+
 if [ $PUPMODE -eq 2 ]; then
 	if [ "$xpkgname" != "" ] && [ "$installed_pkg" == "" ]; then
 	  if [ "$PKGUPDOWN" == "" ]; then
@@ -849,7 +863,7 @@ if [ $PUPMODE -eq 2 ]; then
 		do
 		
 		 #Check if the file was a part of the newly installed package
-		 if [ "$(grep -m 1 "$xline" /var/packages/${DLPKG_NAME}.files)" == "" ]; then
+		 if [ "$(grep -m 1 "$xline" /var/packages/${DLPKG_NAME}.files)" == "" ] && [ "$(grep -m 1 "$xline" /tmp/pkg-files.list)" == "" ]; then
 		  #Not a part of newly installed package. Do action
 
 		   #Delete the file which is not a part of upgrade
@@ -865,7 +879,7 @@ if [ $PUPMODE -eq 2 ]; then
 	fi
 fi
 
-
+rm -f /tmp/pkg-files.list 2>/dev/null
 
 #120907 post-install hacks...
 /usr/local/petget/hacks-postinstall.sh $DLPKG_MAIN
--- a/woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
+++ b/woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
@@ -73,6 +73,7 @@ DL_SAVE_FLAG=$(cat /var/local/petget/nd_category 2>/dev/null)
 
 clean_and_die () {
   rm -f /var/packages/${DLPKG_NAME}.files
+  rm -rf ${DIRECTSAVEPATH}/${DLPKG_BASE}-extracted 2>/dev/null
   [ "$PUPMODE" != "2" ] && busybox mount -t aufs -o remount,udba=reval unionfs / #remount with faster evaluation mode.
   exit 1
 }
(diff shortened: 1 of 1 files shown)


rizalmart (commented):
Fix Issue #3006

mavrothal:
I think the main reason that PPM has become such a beast is that is allowing for any stupid package to be installed and at the same time tries to foresee the stupidity and protect the user from it!

woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
```diff
@@ -813,7 +813,9 @@ if [ "$xpkgname" != "" ] && [ "$installed_pkg" != "" ]; then
 	 fi
 	done < /var/packages/${installed_files}.files
 	
-	rm -f /var/packages/${installed_files}.files
+	if [ "${DLPKG_NAME}.files" != "${installed_files}.files" ]; then
```
mavrothal, line 879: But how and why a new packages file and an existing package file will have the same filename???

woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
```diff
@@ -244,6 +245,15 @@ elif [ $PUPMODE -eq 13 ];then
 	fi
 fi
 
+
+if [ -L /bin ] || [ -L /lib ] || [ -L /lib64 ] || [ -L /lib32 ] || [ -L /libx32 ] || [ -L /sbin ]; then
```
mavrothal, line 249: I though 3 tests were too many to identify a usr-merged distro. 6 is way an overkill
/bin and or /lib should be enough

woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
```diff
@@ -335,6 +345,40 @@ case $DLPKG_BASE in
...
+   #Fix file position for usrmerge FHS
+       
+   mkdir -p usr 2>/dev/null
+
+   [ -d bin ] && mv -f bin usr/
+   [ -d sbin ] && mv -f sbin usr/
+   [ -d lib ] && mv -f lib usr/
+   [ -d lib64 ] && mv -f lib64 usr/
+   [ -d lib32 ] && mv -f lib32 usr/
+   [ -d libx32 ] && mv -f libx32 usr/
```
mavrothal, line 364: Is there any distro or pet that has /lib64, /lib32 and /libx32? 
All cases I have seen with these these dirs have them in /usr. No?

woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
rizalmart: Because some users renamed the packages for some reasons instead of leaving the package filename alone

woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
rizalmart: It was patterned on Ubuntu usrmerge. Its better to anticipate the packages' structure rather than creating trouble in the future or simply called future proofing

woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
rizalmart: Its just future proofing feature. I modified PPM with future proofing and improved package management reliability in mind.

CollaboratorGCM:
Does the forum package upgrade **[seen here](https://forum.puppylinux.com/viewtopic.php?p=53938#p53938)** provide useful resolve for PPM? 

Something that should be considered?

mavrothal:
Is said many times that PPM should be scrapped (and I think also `pkg` because has a similar philosophy).
Instead is becoming an uglier and uglier rabbit hole (that I had a fair contribution into digging it) trying to handle every distributed binary out there.
The result is to keep becoming slower and more unreliable, I believe.

However, a distro package manager should do just that. Manage correctly the **distro** packages. 
For PPM, this should be to manage pets and refer distro-compatibility packages to the mother-distro package manager (apt, rpm, pacman, xbps etc). Mind you **only** one mother distro per pup. i.e. debian-based pups should not have rpm, slackware, etc, support.
Coming to pets, only **distro qualified** pets should be installable by PPM. i.e. pets that are signed or added to a list maintained by the puppy developer.

I know this is not “the puppy way” but if a user wants to install anything else there is uextract or other utilities that can help to manually install anything that flies around. Or maybe a woof-independent "foreign package handler".  But in all cases it should be clear that is not the puppy-builder's responsibility to accommodate you there. 

I think that integration of apt and xpbs in recent puppies is a step in the right direction and if the next slacko follow suit then you can really make a reliable and fast puppy package manager.

Of course the other problem is that PPM infrastructure is used to build the puppies. 
PPM should be uncoupled from puppy building! 
There is already an 8year-old [proposal](https://github.com/puppylinux-woof-CE/woof-CE/tree/woof-next) on this.

Bottom line, I think that patching up PPM goes in the wrong direction and serious though should be given on what the Puppy Package Manager should be doing,  so I have no wish to pull this request. However, the next puppy builder that wants to use PPM **as is**, it should consider it to decrease questions about broken  systems.

dimkr:
> Is said many times that PPM should be scrapped (and I think also `pkg` because has a similar philosophy).

I tend to agree with most of what @mavrothal said.

This package mix n' match party needs to stop, and people who use PPM should understand the risk they're taking. For example, unlike proper package managers, it doesn't check if `x-dev` is the same version as `x`, and it doesn't understand the concept of an `or` relationship between two packages that satisfy a dependency. PPM is not just slow and buggy, it's even unsafe for users who do play by the rules and don't try to force an old .pet package on a Puppy with /usr symlinks.

This issue with PPM is a drop in the ocean, but maybe it's worth fixing. My dpup and upup builds are proudly built with 0% .pet packages and have warnings all over. I don't mind any changes to PPM, but I'd love to see it disappear.

And I also think that a Puppy built from a combination of packages from multiple distros, is a huge step in the wrong direction. A Puppy built from a mix of multiple distros (like Debian+Ubuntu) is incompatible with either distro. And a Puppy that contains old .pet packages built on some older Puppy, is even worse in that regard: very often, it can't be used rebuild these .pet packages, so it's stuck with old magic binaries that may contain bugs (and who knows, may even malware). The JWM menu generator is probably the best example - some Puppy builds use a 4.x era .pet package with horrible bugs.

A Puppy version that is unreproducible is the worst thing that can happen to Puppy's evolution. Whatever can't be rebuilt, this time with an extra fix or two, cannot be improved upon, and users who know that come up with workarounds instead of reporting bugs (and getting them fixed for good).

Is it a good decision to allow the use of old and potentially incompatible packages, unlike other distros? I believe not, but users seem to ignore all warnings, and just want to rinse and repeat their .pet/SFS thing forever.

01micko:
Let's kill it with a :dagger: to the :heart:  , but I'll need some help with `slaptget` and `gslapt` for slacko family.

Issue #2957, opened 2022-03-07
Port GTK+ 3 themes to GTK+ 4


dimkr:
This shouldn't be too hard, because GTK+ 4 uses CSS. The selectors are different and the default settings are different.

The GTK+ 3 themes are a good starting point: there are some errors, some widgets have weird backgrounds, some images are stretched, some widgets use the Adwaita style ... but it looks like 80% of the CSS can be reused:

![gtk4](https://user-images.githubusercontent.com/1471149/156984588-9456399a-9c8c-42ad-98dc-43dee11b737f.png)

dimkr:
Slow but good progress:

![gtk4](https://user-images.githubusercontent.com/1471149/157731719-836821ff-dc0b-4b15-baa0-c7f6fedd734b.png)

PR #2958, opened 2022-03-07
Add GTK+ 4 theme support to ptheme

dimkr:
This will become useful once #2957 is fixed. For now, there are no GTK+ 4 themes, so this doesn't do anything.

Ubuntu 22.04 sticks with GTK+ 3, but some applications have moved on. It won't be long until the horrible visual inconsistency we had between GTK+ 2 and 3 returns, but this time, between GTK+ 3 and GTK+ 4.


01micko (approved):
I think it should be ok

CollaboratorGCM:
+1

PR #2963, opened 2022-03-08
Add overlay support to initrd.gz

dimkr:
With this PR, a Puppy built with `UNIONFS=overlay` uses overlay instead of aufs. sfs_load only manages the list of SFSs to load at boot time, and SFSs cannot be loaded on-the-fly.

Still testing this. So far, tested live boot and use of a savefolder. Seems to boot faster in a VM, compared to aufs, but it could be an illusion.

sfs_load has many levels of indentation, making it impossible to minimize the changes. It's easier to review this PR with whitespace changes hidden: https://github.com/puppylinux-woof-CE/woof-CE/pull/2963/files?diff=unified&w=1.


mavrothal:
A brief test of the init and sfs_load on bionic64 installed in a HD shows no problem (using aufs).
Tried it with the vDpup kernel but is not functional on bionic64 (no input devices)
So, not sure how the init will behave with layerfs. Specially with the options to have the SFS/save files all over the place as the current init options allow.

Using the sfs_load with fugalified vDpup not much is happening.
The SFS-to-be-installed is queued and stays queued (ie not loaded or moved at the root of the partition)

Bottom line, more testing is needed.

mavrothal:
Forgot to mention that adding a `punionfs` as kernel parameter in the new init and setting it to `overlay`  results in kernel panic using the vDpup 5.10.102 kernel and zdrv (un-usrmerged).

dimkr:
This won't work with Vanilla Dpup, it relies on the changes in 3builddistro.

mavrothal:
I managed to get bionic64 (in a subdir) going with overlayfs. Looks OK.
Savefolder created and loaded.  However, is pretty messy .
The savefolder has the usual files and also "upper" that also has the overlayfs save directories.
Trying to sfs_load (with the new sfs_load) an SFS at the root of the partition is queued  and never loads. 
Trying to laod an SFS at the root of the subdir results in kernel panic.

mavrothal:
> This won't work with Vanilla Dpup, it relies on the changes in 3builddistro.

I just added in the init a pfix option `overlay) PUNIONFS="overlay";;` and booted with pfix=overlay

dimkr:
> Trying to sfs_load (with the new sfs_load) an SFS at the root of the partition is queued and never loads.

As I mentioned earlier - sfs_load is used only to select the SFSs to load at boot time.

mavrothal:
> As I mentioned earlier - sfs_load is used only to select the SFSs to load at boot time.

I do not expect to SFS_load to load the SFS in the overlayfs.
What I see and mentioned above  is  the selection of the SFS by sfs_load is not working as expected, because  
a) when the SFS is at the root of the partition the SFS is queued but not loaded  in the layerfs at reboot,
and b) when the SFS is at the root of the subdir that the puppy is booting from, leads to kernel panic at reboot.

mavrothal:
The [f997a89](https://github.com/puppylinux-woof-CE/woof-CE/pull/2963/commits/f997a89a464fff03d30e68984bdcc69c03b977cd) init still gives kernel panic trying to load an extrasfslist sfs file

dimkr:
https://github.com/dimkr/woof-CE/releases/tag/vanillaupup-22.03-beta4-overlay is a rebuild of Vanilla Upup 22.03 beta4, with this branch merged.

Tested saving into a savefile under the partition root and saving to a folder (with psubdir) seem to work just fine. I tested loading of SFSs from psubdir and from the root, and made sure init continues if a SFS is missing or invalid.

mavrothal:
The init in vanillaupup-22.03-beta4-overlay works as advertised with overlayfs AND in with aufs (removed PUNIONFS from the puppy sfs PUPSTATE file that init looks at to deside for the unionfs)
Testing was brief of course but still

mavrothal:
Not sure if this is a design choice buy [acfc73d](https://github.com/puppylinux-woof-CE/woof-CE/pull/2963/commits/acfc73de7190d29d07cc7a3708097b0fad255d81) finds extraSFSs only in then PSUBDIR and not at the root of the partition

mavrothal:
[6ae9d21](https://github.com/puppylinux-woof-CE/woof-CE/pull/2963/commits/6ae9d210234ebc26f0ad506712c6c3786c63d2ba) still fails for me to  load SFSs from the root of the partition in overlayfs.

initrd-progs/0initrd/init
```diff
@@ -1282,7 +1285,7 @@ if [ "$UNIONFS" = 'overlay' -a -f /pup_new/etc/rc.d/BOOTCONFIG ]; then
  . /pup_new/etc/rc.d/BOOTCONFIG
  for ONE_SFS in $EXTRASFSLIST; do
   LOOP=`losetup -f`
-  find_onepupdrv "" "" "$ONE_SFS" "p"
+  find_onepupdrv "" "" "$ONE_SFS" "p" "1"
```
mavrothal: `"1"` or` ""` ?

dimkr: 1

dimkr: This is the only case where we should check both psubdir and the root

mavrothal:
Looks like that what is happening is the `grep -m1 /dev/sda1` shows as mounted in /initrd/mnt/pdrv that does not exist and not in /initrd/mnt/dev_save which is actually the root of the partition.

mavrothal:
[7047bd0](https://github.com/puppylinux-woof-CE/woof-CE/pull/2963/commits/7047bd0b4adebe90ecaf63af5aa653bff8624070) loads extra SFSs from the root of the partition fine. However it fails to load them from PSUBDIR.
Interestingly enough, if you reintroduce the syntax error of  [20eed6f](https://github.com/puppylinux-woof-CE/woof-CE/pull/2963/commits/20eed6f64286d749bb34676653f675e8b38fe02a) it now recognises the extra SFSs in the PSUBDIR but not the root.
You might want to go back to [4a9bdfc](https://github.com/puppylinux-woof-CE/woof-CE/pull/2963/commits/4a9bdfc3a12eb3699912c38416c1876e05ac597c)

mavrothal:
[dfde6df](https://github.com/puppylinux-woof-CE/woof-CE/pull/2963/commits/dfde6df52575df972377000f2b9fcfa7bb9d9cd3) works as expected both under overlayfs and aufs.

One issue I noticed under AUFS is that sfs_load fails to unload extra SFSs indicating that they are in use. Does not appear to be an sfs_load issue as the woof-ce version also fails.  Looks like is an effect of the way extra SFSs are mounted in the union now.

mavrothal:
Not sure if [a68b4ff](https://github.com/puppylinux-woof-CE/woof-CE/pull/2963/commits/a68b4ffca6c4859b73b13ac3270a19f2d2bed7f7) aims to address the unloading of extraSFSs loaded in the union at boot time under AUFS, but it does not. 
(No other testing was done with this init)

BTW I find this addition handy (for me now and) in case unionfs becomes user-configurable

```
--- a/init	2022-03-14
+++ b/init	2022-03-14
@@ -1083,6 +1083,9 @@
 mount
 echo --------------------
 
+# Inform the user for the layer system we use in case of multiple savefiles
+echo -en "$(printf "The layer file system is \\033[1;35m${UNIONFS}\\033[0;39m")" > /dev/console
+
 PUPSAVE=""
 if [ "${SAVE_MP}" != "" -a "$PRAMONLY" != "yes" ];then #have mounted save? partition
  #check if save partition is linux
```

mavrothal:
[6e72953](https://github.com/puppylinux-woof-CE/woof-CE/pull/2963/commits/6e72953a38ffff39c55bcbc327dcb41d180b2535) still fails to unload extrasfs. 
sfs_load fails to `-o remount,del:/initrd/pup_ro3 unionfs /` at L1507 with the generic `mount-FULL: /: wrong fs type, bad...` error

BTW the addition of "filesystem used" at the switch_root stage is impossible to catch up at the console and fairly  late in the game to be of any use to the user.

dimkr:
> [6e72953](https://github.com/puppylinux-woof-CE/woof-CE/pull/2963/commits/6e72953a38ffff39c55bcbc327dcb41d180b2535) still fails to unload extrasfs. sfs_load fails to `-o remount,del:/initrd/pup_ro3 unionfs /` at L1507 with the generic `mount-FULL: /: wrong fs type, bad...` error

Is it with aufs? If yes and it's the same behavior as in testing before this PR, I'm not going to fix this.

If it's with overlay, then sfs_load shouldn't try to unload, and it's a bug I need to fix.

mavrothal:
> Is it with aufs? If yes and it's the same behavior as in testing before this PR, I'm not going to fix this.
>
> If it's with overlay, then sfs_load shouldn't try to unload, and it's a bug I need to fix.

This with AUFS and this behaviour appears to be triggered by the specific PR. 
Unmounting SFSs from the extrasfs list in  9.0.35 works fine. 
However, in  vanillaupup-22.03-beta4-overlay (with modified PUPSTATE to run aufs) and the [6e72953](https://github.com/puppylinux-woof-CE/woof-CE/pull/2963/commits/6e72953a38ffff39c55bcbc327dcb41d180b2535) init, it fails. Both with the modified sfs_load of this PR or the sfs_load from 9.0.35.

Could be an issue with busybox mount and mount-FULL or something else in vanillaupup-22.03-beta4-overlay that triggers that. If not, this PR breaks the sfs_load function for aufs-running puppies

dimkr:
> Unmounting SFSs from the extrasfs list in 9.0.35 works fine.

9.0.35 is not built from testing (very far behind), so this is irrelevant.

I'll test everything aufs-related in a dpup built from this PR.

mavrothal:
Not sure how relevant 9.0.35 is, but using the the [6e72953](https://github.com/puppylinux-woof-CE/woof-CE/pull/2963/commits/6e72953a38ffff39c55bcbc327dcb41d180b2535) init with it gives NO problem in unloading sfs files from the extrasfs list.
This is good I guess.

Latter I may test the sfs_load in that environment too.

mavrothal:
> Latter I may test the sfs_load in that environment too.

The sfs_load from this PR also works  in 9.0.35 with or without the new init.
So looks like the sxtrasfs unloading was some strange  vanillaupup-22.03-beta4-overlay issue.

dimkr:
PUPMODE 13 is broken,

```
+ yaf-splash -bg orange -placement top -close never -text 'Saving RAM to '\''pup_save'\''...'
cat: '/sys/fs/aufs/si_*/br0': No such file or directory
cat: '/sys/fs/aufs/si_*/br1': No such file or directory
stat: cannot statx '': No such file or directory
 is not mounted!
```

mavrothal:
Not sure if it is another issue specific to vanillaupup-22.03-beta4-overlay but with the [dba232d](https://github.com/puppylinux-woof-CE/woof-CE/pull/2963/commits/dba232de55cf92bc4ad3cb4ee8fc2e1169ae6f49) init, under AUFS everything is fine. 
Under overlay, first boot (pupmode 5) is fine. Generate savefolder and reboot (to pupmode 12) is freezing at `Performing a 'switch root' to....` 
Does it need an `else` for the new condition to know what to do if the union IS overlay?

dimkr:
> Does it need an else for the new condition to know what to do if the union IS overlay?

Where? Which condition? 🤔

mavrothal:
Was only one condition  changed in [dba232d](https://github.com/puppylinux-woof-CE/woof-CE/pull/2963/commits/dba232de55cf92bc4ad3cb4ee8fc2e1169ae6f49)  !
`if [ $PUPMODE -eq 12 -a "$UNIONFS" != 'overlay' ];then`
In any case does it work for you in pupmode 12 with overlayfs?

dimkr:
PUPMODE 13 relies on aufs' behavior when a read-only layer that isn't the topmost is modified, and the entire whiteouts thing. It won't work with overlay.

mavrothal:
Let me repeat what I said above (with bold points to hopefully make it clear) 

> with the [dba232d](https://github.com/puppylinux-woof-CE/woof-CE/pull/2963/commits/dba232de55cf92bc4ad3cb4ee8fc2e1169ae6f49) init
> ...
> Under **overlay**, first boot (**pupmode 5) is fine**. Generate savefolder and **reboot (to pupmode 12) is freezing** at Performing a 'switch root' to....

So again, the problem I see is with **overlayfs in pupmode 12**.

mavrothal:
A new problem with the 9cdde65 init and OVERLAYFS in PUPMODE 12 (just to be clear...)
Steps to recreate
1) Boot from a HD PSUBDIR (pupmode 5) and save into a folder.
         Savefolder is populated with the usual folders (etc, home, root, usr etc)
2) Reboot. It boots AGAIN in pupmode 5. i.e. you get the quicksetup dialog  
         On the second reboot however, it does not ask if you want to save (as expected in pupmode 5) but saves the  user data in the `upper` folder within the previously generated savefolder, and populates it with the usual files (etc, home, root, usr etc).  
        A  `work` folder is also generated within the savefolder
From that point on user data are saved in /PUBDIR/savefile/upper/
In other words it generates a setup similar to frugalify but within the savefolder.

dimkr:
https://github.com/dimkr/woof-CE/releases/tag/vanillaupup-22.03-beta6-overlay-kbuild successfully booted from a flash drive and a savefolder, with PUPMODE 12.

> In other words it generates a setup similar to frugalify but within the savefolder.

Yes, overlayfs needs a "working directory" that cannot overlap with the writable layer.

mavrothal:
> Yes, overlayfs needs a "working directory" that cannot overlap with the writable layer

I may not be clear as you would like  but please try to see what I say.
~~Up to 3037ca3 in **overlayfs** pupmode 12 the `vanilapupsave` folder had the expected structure ie `/vanilapupsave/{etc,home,initrd,root,run.usr,var}`~~
~~After  9cdde65~~ it has `/vanilapupsave/{etc,home,initrd,root,run,upper,usr,var,work}`

mavrothal:
Try once more
On a hard disk install of vanillaupup-22.03-beta6-overlay-kbuild
On first boot and shutdown you are asked if you want to save the changes and if yes, the vanillaupupsave folder is generated and populated. 
However ,`vanillaupupsave/etc/rc.d/PUPSTATE` has `PUPMODE=5`
On **first reboot** you are going again through the quicksetup, since PUPMODE=5 ,and on the second shutdown **changes are saved without asking** on `vanillaupupsave/uper/`. 
There, the `vanillaupupsave/upper/etc/rc.d/PUPSTATE` has the correct `PUPMODE=12`
So you end up with a vanillaupupsave folder that has this structure : 
`etc, home, initrd, root, run, upper{etc, home, initrd, root, run, usr, var,}, usr, var, work{work}`

mavrothal:
PUPMODE 13 with al the latest changes on a VFAT ATAFLASH drive
With overlayfs: 
savefile is generated and populated on first boot. 
This file is loaded OK on reboot, and save icon appears on Desktop. 
However, quicksetup also runs...
On reboot a save prompt appears in the console. Prompt to save and then reboot.
The same scenario is run all over in all consecutive reboots.
upper and work dirs are never generated in the savefile.

With aufs,
savefile is generated and populated on first boot. 
This file is loaded OK on reboot, and save icon appears on Desktop. 
Everything looks OK. 
Manual snapmergepuppy call works as expected.
On reboot a save prompt appears in the console. 
Prompt to save (or not) and goes on with computer freeze requiring hard reboot.

dimkr:
https://github.com/dimkr/woof-CE/releases/tag/vanillaupup-22.03-beta7-overlay-kbuild contains all commits until b2fd72c and the fix for aufs segfaults. I tested PUPMODE 12 with a folder, and 13 with both a folder and a file. Moving on to 66.

mavrothal:
with c4d3cc42 and
PUPMODE 13 with al the latest changes on a VFAT ATAFLASH drive
With AUFS: 
the issues mentioned above remain

With OVERLAYFS:
savefile is now generated ok with upper and work
upper is populated ok and on reboot manual save is on the desktop.
On first manual call of snapmergepuppy(.overlay) it fails with errors (files not found) and nothing is saved.
On second call procceeds presumably OK but again files are not saved and can not be found in /initrd/pup_loop/upper/*   
On rebood asks again to generate a new savefile!

Is beta7-overlay up to c4d3cc42 or it has additional changes?

dimkr:
> Is beta7-overlay up to https://github.com/puppylinux-woof-CE/woof-CE/commit/c4d3cc42d227218034d28c1243da35263e1bb178 or it has additional changes?

It has #2975, #2974, #2973 and #2961.

If you're still testing things with beta<7, please don't - aufs segfaults, and beta7 includes the fix.

EDIT: PUPMODE 13 from an ext4-formatted partition on a flash drive (with beta7) seems to work well: changes done in the first boot got saved, and snapmergepuppy works fine:

boot
pm13

Confusingly, when booting from an ISO but with `pmedia=usbflash` (to activate PUPMODE 13), looks like initrd-progs/0initrd/init has this magic block, which causes it to ignore the save file on the flash drive:

```
#establish PDRV
P_PART=""; LOOK_PUP=""; LOOK_SAVE=""
if [ "$P_BP_ID" ];then #specified as parameter...              <- false
 ...
elif [ "$PDRV" = "" ];then #not specified anywhere            <- true
 #determine what to search for
 [ "$PMEDIA" = "cd" ] && LOOK_SAVE="yes"                  <- false
...
```

mavrothal:
Fast check with beta7 plus 834763c but not 3f95ece
In PUPMODE 13 on a VFAT ATAFLASH drive
AUFS keeps panicking at shutdown but files are saved fine both mid session and at the end.
OVERLAYFS Looks good now! 
savefile generation and use, saving mid session, saving at the end,  panic at shutdown, all good
One thing that needs fixing is resizing savefile in pupmode 13 (is OK in AUFS)

dimkr:
> AUFS keeps panicking at shutdown but files are saved fine both mid session and at the end.

Which kernel do you use - the one that came with beta7?

mavrothal:
> Which kernel do you use - the one that came with beta7?

Yes. The test system is fresh beta7 plus 834763c  and repackaging the sfs with the `-comp zstd -Xcompression-level 19 -b 256K -no-exports -no-xattrs` options which I believe is what you use.

And of course `PUNIONFS='overlay'` has been removed from the PUPSTATE to run with AUFS.

dimkr:
> `-comp zstd -Xcompression-level 19 -b 256K -no-exports -no-xattrs` options which I believe is what you use.

`-comp xz -Xbcj x86 -b 256K -no-exports -no-xattrs` 😸

mavrothal:
The kernel panic during aufs shutdown in pupmode 13 (VFAT ATAFLASH drive) hapens at the final
`umount-FULL -in -a -r -d >/dev/null 2>&1` step
during shutdown umount-FULL is complaining during `merging of /initrd/mnt/tmpfs/pup_rw onto /initrd/pup_ro1` for "no such branch" (though mount shows that loop5 is mounted /initrd/pup_ro1)
and then `mount unionfs on / failed` with device busy leading to the panic

I guess the kernel does not have the latest aufs patch as reported [here](https://forum.puppylinux.com/viewtopic.php?p=51730#p51730)

dimkr:
> I guess the kernel does not have the latest aufs patch as reported [here](https://forum.puppylinux.com/viewtopic.php?p=51730#p51730)

It includes https://github.com/sfjro/aufs5-standalone/commit/fbd4f0919618f70d4edc26724a5a27564821d5d7#diff-94ef3d38157205b0e6270b12d5adad47cc24d9b492d6c0fe36e52094b90ac1b4R1107

mavrothal:
> It includes sfjro/aufs5-standalone@fbd4f09

Maybe so, however PeeBee's [5.16.15-lxpup64](https://sourceforge.net/projects/lxpup/files/Other/LxPupSc/interims/) kernel (usrmerged) shuts down without any panicking.

dimkr:
> > [](https://github.com/sfjro/aufs5-standalone/commit/fbd4f09)
> 
> Maybe so, however PeeBee's [5.16.15-lxpup64](https://sourceforge.net/projects/lxpup/files/Other/LxPupSc/interims/) kernel (usrmerged) shuts down without any panicking.

And everything else, except the kernel, is exactly the same? Same set of boot/save/shutdown steps, etc'. I want to know if it's yet another aufs bug, or if the aufs fix works only in 5.16.x. If it's aufs' fault, I can move on to other things, like the save file resize thingy.

mavrothal:
> And everything else, except the kernel, is exactly the same?

In the exact drive (with savefiles and all) just replaced vmlinux and zdrv with the 5.16.15 version. Nothing else.
(Even left  kbuild in though from a different version.)
So, I would say is a kernel issue

peabee:
The name of the aufs branch has not changed so you need to delete the old aufs sources from kernel-kit to ensure that the patched version is downloaded and used....

dimkr:
> The name of the aufs branch has not changed so you need to delete the old aufs sources from kernel-kit to ensure that the patched version is downloaded and used....

The build job runs without any kind of cache, and pulls the latest commit of the 5.15.x branch every time.

mavrothal:
7e62425 works to resize the save file at reboot in overlayfs.
However, at reboot quicksetup is coming up although PUPMODE is 13.
This is an overelayfs-only issue.
I guess the trigger tries to read save file too early while is resizing so does not see the savefile at all?

dimkr:
That's weird, I'll take a look tomorrow. I tested save files on a flash drive with PUPMODE 13, and everything worked 100%.

mavrothal:
> That's weird, I'll take a look tomorrow.

Maybe not. I started with another fresh savefile and did not have this in a couple of tries.

I think that, that one time was because before I had booted the overlay savefolder with PeeBee's kernel (that has overlay as a module) and  failed. This might had the savefile messed up enough to trigger the effect.

mavrothal:
I think is good enough for a test release to wider audiences to catch any other issues. 

Tho an AUFS version should also be tested  with these changes
Does `bbe` works with SFSs or maybe `hexdump / xxd` ???...

dimkr:
> Tho an AUFS version should also be tested with these changes

https://github.com/dimkr/woof-CE/releases/tag/vanillaupup-22.03-beta8-overlay-kbuild
https://github.com/dimkr/woof-CE/releases/tag/vanillaupup-22.03-beta8-aufs-kbuild

mavrothal:
You are looking for trouble !
Is clear enough when you download the ISOs, but after that there is **nothing** to tell which is which.
People will mix them up and all kinds of "problems" will appear.
You must improve your naming scheme!

01micko (commented):
I need to test these changes. Not much time lately so hopefully soon.

dimkr:
> I need to test these changes. Not much time lately so hopefully soon.

You're more than welcome!

The main issue I see right now is with PUPMODE 13: if I install a browser, it works just fine until I save. After saving, the browser executable is missing from /usr/bin, although I can see it under /initrd/pup_ro1/usr/bin. `echo 3 > /proc/sys/vm/drop_caches` doesn't make the file appear under /usr/bin, but the problem is gone after I reboot.

dimkr:
https://github.com/dimkr/woof-CE/releases/tag/vanillaupup-22.03-beta9-overlay-kbuild works very very very nicely here, with PUPMODE 13 from a flash drive. Changes are saved quickly, and under normal use, most changes come from the constantly changing browser cache. This 11 year old laptop is limited to USB 2 speeds, and generally, it's not super healthy. But everything is super snappy and RAM consumption remains below the 2 GB mark even with many tabs open and after I installed several large and heavy applications.

EDIT: 20dee90 should improve things even further. Because most of the contents of pup_rw are unchanged between two consequent snapmergepuppy runs, we can sort the files by modification time. The files that haven't changed recently are far more likely to be unmodified, we're less likely to lose data (i.e. power loss) if we start with the recently modified files, then check what not-recently-modified files have changed.

initrd-progs/0initrd/init
```diff
@@ -785,7 +785,7 @@ remount_overlay() {
   umount -l /pup_new
  else
   mkdir /mnt/tmpfs/pup_work
-  OPTS="lowerdir=,upperdir=/mnt/tmpfs/pup_rw,workdir=/mnt/tmpfs/pup_work,xino=on,index=on"
+  OPTS="lowerdir=,upperdir=/mnt/tmpfs/pup_rw,workdir=/mnt/tmpfs/pup_work,xino=on"
```
mavrothal, line 795: In current kernels  both OVERLAY_FS_INDEX and OVERLAY_FS_XINO_AUTO are not set (also OVERLAY_FS_INDEX),
but are good features to have. Better activate them in the kernel rather than remove them

mavrothal:
In my testing setup I now have OVERLAY and AUFS versions, sharing all SFSs/vmlinuz/initrd but the  puppysfs. However in beta8 bdrv and initrd.gz have different hashes but work fine. 
Is it because of timestamps or are indeed different?

dimkr:
> However in beta8 bdrv and initrd.gz have different hashes but work fine. Is it because of timestamps or are indeed different?

The files inside them have different modification times.

initrd-progs/0initrd/init
dimkr: OVERLAY_FS_INDEX and OVERLAY_FS_XINO_AUTO determine whether or not to auto-enable these features for a newly mounted overlay. Support for them is always included, and these features are active if the build-time option is enabled, or if enabled explicitly via mount options.

mavrothal: Even when kernel config options is "not set"?

mavrothal:
OK, THX.

You may want to take another look at sfs_load
Beat9 with 20dee90 added
under AUFS:
sfs_load shows bdrv and kbuild as extra SFSs loaded (right panel) and loadable (left panel).
Trying to unload bdrv fails (understandably)
Trying to unload kbuild succeeds i.e. the sfs IS unloaded but it remains showing as loaded in the right panel of sfs_load. And of course reloads upon reboot even if according to the respected function of sfs_load it should not.
So I guess sfs_load needs some teaching about bdrv and init under aufs may need to but kbuild in the extrasfs files or teachit that is a system file (kdrv???)

under OVERLAYFS:
again sfs_load shows bdrv and kbuild as extra SFSs loaded and loadable and on a call to unload them “does” in immediately reporting success, but of course nothing is unloaded in the present or the next session.

initrd-progs/0initrd/init
dimkr:
```
“xino”

Enabled with the mount option “xino=auto” or “xino=on”, with the module option “xino_auto=on” or with the kernel config option CONFIG_OVERLAY_FS_XINO_AUTO=y.
```

```
“inode index”

Enabled with the mount option or module option “index=on” or with the kernel config option CONFIG_OVERLAY_FS_INDEX=y.
```

dimkr:
> sfs_load shows bdrv and kbuild as extra SFSs loaded (right panel) and loadable (left panel).

WIP, see #2981

mavrothal:
will the #2981 patches of init and sfs_load apply cleanly on beta9 files (for testing)?

dimkr:
> will the #2981 patches of init and sfs_load apply cleanly on beta9 files?

No, you'll need to add an extra `""` argument to the `find_onepupdrv` calls - if unsure, just wait for beta10 and hold off your excitement 😸 

EDIT: https://github.com/dimkr/woof-CE/releases/tag/vanillaupup-22.03-beta10-overlay-kbuild

mavrothal:
In beta10 and Overlayfs
Trying to unloading an extra SFS sfs_load says “unloaded” but of course is not and still shows in the loaded (right panel)
It is unloaded upon reboot fine though.
So the gtk message must be improved to reflect the situation. Same thing for the loading message.
Also, when the SFS to be loaded is not at the root of the partition or the PSUBDIR, sfs_load (in ausf) offer to copy it there. Not when running in overlayfs, so on reboot is not loaded.

In beta 10 and Aufs
sfs_load has no issues and behaves as expected.

dimkr:
> So the gtk message must be improved to reflect the situation. Same thing for the loading message.

Yes, the wording in sfs_load is horrible. A "queued" SFS is "unloaded", etc'. But let's leave that aside, because sfs_load is not in great shape: it's a very long script that tries to do way too many things. It's a good idea to fix sfs_load, but not here, and not now: my time is limited and I need to focus.

Problems related to auto-loading of bdrv and kbuild ([b]with aufs[/b]) should be addressed in https://github.com/puppylinux-woof-CE/woof-CE/pull/2981, and all overlay-specific issues must be addressed here. The rest (in particular, UI issues) can be fixed later.

(Help is welcome, and if someone wants to take care of sfs_load, that would be great. I'm doing my best to minimize changes in this PR, to reduce risk and avoid conflicts with other PRs.)

dimkr:
@mavrothal https://github.com/dimkr/woof-CE/releases/tag/vanillaupup-22.03-beta11-overlay-kbuild contains 2ee35c8 and 8cfbf5f, the wording in sfs_load is still bad but it should work.

mavrothal:
> https://github.com/dimkr/woof-CE/releases/tag/vanillaupup-22.03-beta11-overlay-kbuild contains https://github.com/puppylinux-woof-CE/woof-CE/commit/2ee35c8fa709765e69419b414287d621a65fd75a and https://github.com/puppylinux-woof-CE/woof-CE/commit/8cfbf5f7775788f75ff957ef3f9d054d64655031, the wording in sfs_load is still bad but it should work.

Does it also include the sfs_load changes in #2963 ?

dimkr:
> Does it also include the sfs_load changes in #2963 ?

Yes

dimkr:
(Synced with testing after the merge of #2981, cleaned up some whitespace changes, then squashed)

mavrothal:
> https://github.com/dimkr/woof-CE/releases/tag/vanillaupup-22.03-beta11-overlay-kbuild 

beta11 and sfs_load in particular, look pretty good. 
sfs_load messages under overlayfs still need updating but it can wait I guess.
One message that could be fixed though is when you  unload an extra SFS, it reports that the SFS does not appear to be loaded. I guess because it can not  properly detect the relevant loop and mount point.
This can be confusing for the user and it needs more that wording.

Unfortunately I do not think I'll have any more time for this, but as far as I am concerned, this could be merged!
Does not appear to have any regressions for aufs-based puppies and is pretty good with overlayfs.
You may want for someone else to take a look or for a spin, I would not wait too long.

dimkr:
> One message that could be fixed though is when you unload an extra SFS, it reports that the SFS does not appear to be loaded.

I'll take a look. If it's a one-liner fix, I'll do it.

> Unfortunately I do not think I'll have any more time for this

Thanks for the feedback and the help with testing! Much appreciated.

> You may want for someone else to take a look or for a spin, I would not wait too long.

I'm a huge believer in "deliver fast, deliver often": the functionality is 80% of the work, the polish is the remaining 20%, but the polish takes 80% of the time, because it requires more intrusive changes, and its nuanced nature makes it much harder to test. It's easy to tell if saving in PUPMODE 13 works or not, but it's much harder to tell if it's possible to re-queue an unloaded SFS in all possible use cases and PUPMODE combinations.

It this PR works, and doesn't break aufs, IMHO, the polish can wait for later. sfs_load, like quicksetup, tries to do way too many things in one script, it's a mess, and it won't be easy to fix it.

01micko:
Why not break up `sfs_load` into `sfs_load_aufs` and `sfs_load_overlay` and make a symlink at an early stage?

I'm hearing you about `quicksetup`!

(edit)
OR.. keep `sfs_load` as a entry point and `if`[ under the right condition ] `exec sfs_load_aufs` `elif` [ the other right condition ] `exec sfs_load_overlay` `else`  "why are you running a full install?".

dimkr:
> Why not break up `sfs_load` into `sfs_load_aufs` and `sfs_load_overlay` and make a symlink at an early stage?

Tried that! It does so many things in one script, and some things work for the wrong reason: for example, loaded SFSs are considered to be loaded because it checks the output of `losetup -a` at some point, instead of using one of the global variables.

But maybe I can write a simple replacement, just a list of SFs with checkboxes (load next time or not).

> OR.. keep sfs_load as a entry point and if

Yes, like I did with snapmergepuppy - I think this is a better approach.

dimkr:
```
#!/bin/sh

. /etc/rc.d/BOOTCONFIG
. /etc/rc.d/PUPSTATE

export TEXTDOMAIN=sfs_load
export OUTPUT_CHARSET=UTF-8

queue() {
	. /etc/rc.d/BOOTCONFIG
	cat << EOF > /etc/rc.d/BOOTCONFIG.tmp
EXTRASFSLIST='`echo "$EXTRASFSLIST $1" | tr ' ' '\n' | sort | uniq | tr '\n' ' ' | sed -e 's/^ //' -e 's/ $//'`'
PREVUNIONRECORD='$PREVUNIONRECORD'
LASTUNIONRECORD='$LASTUNIONRECORD'
EOF
	mv -f /etc/rc.d/BOOTCONFIG.tmp /etc/rc.d/BOOTCONFIG
}

unqueue() {
	. /etc/rc.d/BOOTCONFIG
	cat << EOF > /etc/rc.d/BOOTCONFIG.tmp
EXTRASFSLIST='`echo "$EXTRASFSLIST $1" | tr ' ' '\n' | sort | uniq | fgrep -v $1 | tr '\n' ' ' | sed -e 's/^ //' -e 's/ $//'`'
PREVUNIONRECORD='$PREVUNIONRECORD'
LASTUNIONRECORD='$LASTUNIONRECORD'
EOF
	mv -f /etc/rc.d/BOOTCONFIG.tmp /etc/rc.d/BOOTCONFIG
}

export -f queue unqueue

LOADED=`losetup -a | sed 's~.*/~~g' | grep '\.sfs$'`

SFS_DIRS="/initrd${PUP_HOME}"
[ -n "$PSUBDIR" ] && SFS_DIRS="$SFS_DIRS /initrd${PUP_HOME}${PSUBDIR}"

BUTTONS=""
for SFS in `find $SFS_DIRS -maxdepth 1 -name '*.sfs' -printf '%f\n' | sort | uniq`; do
	DEF=false
	for BASE in $EXTRASFSLIST; do
		[ "$BASE" != "$SFS" ] && continue
		DEF=true
		break
	done

	if [ "$DEF" = "false" ]; then
		AUTO=0
		for BASE in $LOADED; do
			[ "$BASE" != "$SFS" ] && continue
			AUTO=1
			break
		done
		# if the user didn't ask to load this SFS but it's still loaded, it's
		# probably auto-loaded and we should hide it
		[ $AUTO -eq 1 ] && continue
	fi

	BUTTONS="$BUTTONS <togglebutton><label>$SFS</label><default>$DEF</default><action>if true queue \"$SFS\"</action><action>if false unqueue \"$SFS\"</action> </togglebutton>"
done

TITLE=xx
export DIALOG='<window title="SFS-Load" default_width="400">
	<vbox space-expand="true" space-fill="true">
		'`/usr/lib/gtkdialog/xml_info fixed package_sfs.svg 20 "$(gettext "Which SFS do you want to load?")"`'
		<vbox scrollable="true" height="400">
		'$BUTTONS'
		</vbox>
	</vbox>
</window>'

exec gtkdialog -p DIALOG
```

@01micko What do you think?

01micko:
It's a reasonable start.

Bear in mind some scripts call `sfs_load +my.sfs` to load, so that it can be added to the queue.

dimkr:
Using a upup built from this PR for weeks now and it works very very well. snapmergepuppy.overlay is pretty efficient, and it's much faster than the aufs equivalent. Application start times are great (much better than aufs, probably because pup_rw acts as a cache layer), and changes don't get lost like they sometimes do with aufs, because the inherent race condition between whiteouts handling in snapmergepuppy and the creation of new files is not a problem with overlay. IMHO, the PUPMODE 13 user experience with a slow flash drive, enough RAM and zram swap is much better compared to aufs (but not because overlay is 'better').

There are limitations compared to aufs, like the inability to load a SFS without having to reboot (which I don't find very limiting myself, I `apt install` my applications), and it's easier to accumulate enough files in pup_rw to fill RAM, so you have no choice but to save and reboot (to get a brand new, empty pup_rw).

mavrothal:
Still not merged or further tested? :-o

dimkr:
> Still not merged or further tested? :-o

I'm waiting for @01micko to decide (and I'm giving others time to raise objections) - I think this PR is not perfect, but good enough for a beta release followed by small fixes. I'm using a Puppy built from it as a daily driver on one of my laptops, and it works great.

01micko (commented):
Still have had no time to test this but I'm thinking about waving it through.

dimkr:
As long as it's not the default and doesn't break aufs in any way, I don't think it's risky.

dimkr:
I've been using overlay since March, and it still works very well.

The only problem I had was the small size of pup_rw: if you `apt install` many packages and it gets full, and you have to save+reboot to flush it back to disk, while with aufs, you need to save and don't have to reboot. But it's better since 03c4824: with a limit of 75% (for example, up to 6 GB of changes if you have 8 GB of RAM), it's really hard to fill pup_rw. Under normal use, without installing big packages, it rarely exceeds 1 GB even with uptime of several days, because most changes are browser cache files and they have a size limit.

dimkr:
@puppylinux-woof-CE/members @puppylinux-woof-CE/founders Objections to this PR?

mavrothal:
3 months since the last commit in the PR, is pleanty of time for review.
 I intent to merge this WE (if no one else does)

dimkr:
> 3 months since the last commit in the PR, is pleanty of time for review.

I squashed this PR about a week ago. I added a `punionfs=aufs/overlay` boot code to allow the user to override the default.

I want to cherry-pick this PR into Vanilla Dpup 9.2.x, as a "technology preview" for users who want to try overlay. Some users will find this useful, especially those hit by #3187.

mavrothal:
Forget cherry picking
From what I can see very few pups are build these days from woof
Merge and if it breaks anything,  you'll have the review!!!

dimkr:
> Merge and if it breaks anything, you'll have the review!!!

Ha, it will break PPM (because #3188 and #3115). I hate PPM and I don't use it: all my woof-CE builds include apt. I think I'll just make PPM show an error and quit if it's overlay at /, in a separate PR.

PR #2974, opened 2022-03-15
Add support for periodic TRIM on the save partition

dimkr:
This can improve performance and increase the lifespan of SSDs. I have an old SSD that becomes slower as the weeks go by, until I `fstrim`.

### DIFF
--- a/woof-code/rootfs-skeleton/etc/init.d/trim
+++ b/woof-code/rootfs-skeleton/etc/init.d/trim
@@ -33,6 +33,14 @@ LAST=$LAST" >> /etc/rc.d/TRIMSTATE.tmp
 
 case $1 in
  start)
+  DEV="`grep -m1 /initrd/mnt/dev_save /proc/mounts`"
+  [ -z "$DEV" ] && exit 0
+  DEV="${DEV#/dev/}"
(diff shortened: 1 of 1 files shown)


woof-code/rootfs-skeleton/etc/init.d/trim
```diff
@@ -0,0 +1,38 @@
+#!/bin/ash
+
+. /etc/rc.d/PUPSTATE
+[ $PUPMODE -eq 5 -o $PUPMODE -eq 6 ] && exit 0
+mountpoint -q /initrd/mnt/dev_save || exit 0
+
```
mavrothal, line 6: should also check if is a rotating device interrogating /sys/block/sdX/queue/rotational or using lsblk.
There is no point to trim rotating disks I think

dimkr: It doesn't do anything, so it's safe to try `fstrim` even if it's not supported. And skipping `fstrim` after one failure is probably not the best idea.

dimkr:
This PR had a bug, fixed in https://github.com/puppylinux-woof-CE/woof-CE/pull/2980.

PR #2977, opened 2022-03-17
Remove the duplicate initrd line in menu.lst

dimkr:


01micko:
Don't do it.

The 'errorcheck off' allows stupid stuff like this. If there is no `ucode.cpio` then grub4dos will not boot. I did my research on this, and tests on bare metal, plenty of them.

PR #2983, opened 2022-03-21
Make the 5.15.x x86_64 kernel more like Ubuntu 5.15.0-23.23

dimkr:
The current configuration is derived from Debian, and it's less desktop-oriented. An Ubuntu 22.04 based configuration will make Puppy more compatible with Ubuntu, and less likely to have Puppy-specific issues.


mavrothal:
Wow? Which problems are you trying to solve with support for 168 cores and kernel jumps?
Ubuntu makes kernels to support everything from entry level laptops to server farms and from to professional machines  to specialised instruments.
Why puppy needs that?
What are the limitations of the current kernels from what you see in puppyland?

dimkr:
> Ubuntu makes kernels to support everything from entry level laptops to server farms and from to professional machines to specialised instruments.

This configuration is based on the 'generic' kernel, not some other variant like the 'cloud' one.

If you look closely, you'll see that more storage device drivers are built-in, more multimedia drivers are included, and some settings are better for reduced power consumption and increased responsiveness.

mavrothal:
I looked a bit and I see a lot of exotic features 
As a matter of fact 2777 activated features, 590 of them builtin (and only 254 features removed).
Performance wise I hope you do not suggest support for 126+ cores 👅 
As I asked before, what is the use case or problem you are addressing with these 3000 changes, because any kernel issues I have noticed in puppy forum (though not following closely after the move) is mostly firmware or proprietary modules.
Unless you are thinking to use the Ubuntu kernel directly and get away with kernel building, I really do not see any good reason for this expansion.

dimkr:
> As I asked before, what is the use case or problem you are addressing with these 3000 changes

I can't answer this question, because I'm not trying to solve a specific problem. I want to make more stuff that works on Ubuntu work on Puppy, out of the box. Many guides you can find online don't work for Puppy.

And if you want a more particular example: the various eBPF based tools don't work on Puppy, because some kernel features are missing.

mavrothal:
> if you want a more particular example: the various eBPF

eBPF is aa very interesting new technology. 
But how do you see it working in puppy, a single user portable system running as root.

"Today, eBPF is used extensively to drive a wide variety of use cases: Providing high-performance networking and load-balancing in modern data centers and cloud native environments, extracting fine-grained security observability data at low overhead, helping application developers trace applications, providing insights for performance troubleshooting, preventive application and container runtime security enforcement, and much more." (From https://ebpf.io/)

dimkr:
> But how do you see it working in puppy, a single user portable system running as root.

It's useful for many things, see https://github.com/iovisor/bpftrace/blob/master/docs/tutorial_one_liners.md for examples. I can see many potential uses - for example, monitoring failed attempts to run applications as root, or automatically copying frequently accessed files from pup_ro1 to RAM, to speed up future I/O, etc'. Basically, it allows you to monitor all kinds of kernel events.

(And let's not forget, this is only one thing that's disabled in the current 5.15.x kernel)

dimkr:
I removed most changes, but kept various chipset drivers, sound card drivers, battery drivers and network card drivers that were missing. And most importantly, I kept scheduling and memory management related changes.

PR #2984, opened 2022-03-25
Suspend on lid close even if a USB device is mounted

dimkr:
This thing is driving me crazy!

It's been years since the last time I had a flash drive that complains it's disconnected and doesn't properly resume from suspend. Is this still a problem in 2022?

It takes a very long time for Puppy to boot from USB, and constantly shutting down, only because my laptop doesn't suspend when Puppy is booted from a flash drive, is not an option.


01micko (approved):

dimkr:
@mavrothal @peabee @mrfricks Objections? There must be some good reason why this restriction is still present.

01micko:
I'll merge this tomorrow if no objections are raised.

01micko:
Well, it's tomorrow well and truly.

PR #2989, opened 2022-03-28
.xinitrc: Pipewire improvements

rizalmart:
Run pipewire-media-session on login, pipewire cannot detect the audio device if pipewire-media-session is not running.

### DIFF
--- a/woof-code/rootfs-skeleton/root/.xinitrc
+++ b/woof-code/rootfs-skeleton/root/.xinitrc
@@ -55,6 +55,7 @@ if [ -e /usr/bin/wireplumber ]; then
         while [ ! -e /tmp/runtime-spot/pipewire-0 ]; do sleep 0.1; done
         run-as-spot pipewire-pulse &
         run-as-spot wireplumber &
+	run-as-spot pipewire-media-session &
     ) &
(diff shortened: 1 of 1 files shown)


woof-code/rootfs-skeleton/root/.xinitrc
```diff
@@ -55,6 +55,7 @@ if [ -e /usr/bin/wireplumber ]; then
         while [ ! -e /tmp/runtime-spot/pipewire-0 ]; do sleep 0.1; done
         run-as-spot pipewire-pulse &
         run-as-spot wireplumber &
+	run-as-spot pipewire-media-session &
```
dimkr, line 58: Isn't WirePlumber an alternative to pipewire-media-session? (so you don't need both)

dimkr: (btw, Bluetooth audio and the audio jack work just fine for me, with WirePlumber and without pipewire-media-session)

rizalmart: @dimkr 
Thanks for clarification I will close this PR and release new PR to avoid confusion

PR #2990, opened 2022-03-28
xinitrc: Pipewire improvements

rizalmart:
Use pipewire-media-session as fallback if wireplumber is not installed


dimkr (approved):
No objections to this PR, although all distros I looked at use WirePlumber and I don't think it's a good idea to use pipewire-media-session

PR #2991, opened 2022-03-28
10alsa: Added pipewire support

rizalmart:


woof-code/rootfs-skeleton/etc/init.d/10alsa
```diff
@@ -141,7 +141,7 @@ case "$1" in
 	fi
 	
 	
-	if [ "$(which pulseaudio)" != "" ]; then
+	if [ "$(which pulseaudio)" != "" ] || [ "$(which pipewire-pulse)" != "" ]; then
```
dimkr, line 144: I'm not sure if this works. PulseAudio and PipeWire are started much later, in .xinitrc. This condition can be true only if there's an init script that starts them (and there isn't), and if that init script is guaranteed to run before 10alsa.

PR #2992, opened 2022-03-28
mscw: Added pipewire support

rizalmart:


woof-code/rootfs-skeleton/usr/sbin/mscw
```diff
@@ -333,7 +329,7 @@ if [ "$(grep -v pcspeaker /proc/asound/pcm)" = "" ] ; then
 fi
 
 if [ ! -f /etc/asound.conf ] ; then
-    if [ "$(which pulseaudio)" == "" ]; then
+    if [ "$(which pulseaudio)" == "" ] && [ "$(which pipewire-pulse)" == "" ]; then
```
dimkr, line 332: Why not just `if [ -n "$PULSE_SERVER" ]; then`? Covers both cases and more efficient than running two processes.

(Also useful in other cases, where you want to distinguish between PulseAudio/PipeWire or ALSA)

PR #2995, opened 2022-03-28
Always start pcmanfm after JWM

dimkr:


peabee:
Presumably this DOES NOT affect cases where JWM is not the window manager??

dimkr:
> Presumably this DOES NOT affect cases where JWM is not the window manager??

It affects any window manager, but only if you use the pcmanfm petbuild.

01micko (approved):

peabee:
I feel the petbuild should only build the vanilla component - any puppy specific configuration should be done "elsewhere"....

dimkr:
> I feel the petbuild should only build the vanilla component - any puppy specific configuration should be done "elsewhere"....

The `pcmanfm` petbuild is designed to be a replacement for the `rox-filer` petbuild, that emulates its behavior as much as possible, as long as the user experience is similar or better. For example, single-click action is enabled and the default window dimensions are similar to the small default size of ROX-Filer windows, but the sidebar with mounted devices is enabled and not hidden away, because that's a vital UI element users are familiar with from other OSs and pretty much any popular file manager.

If you want a second petbuild, which doesn't do any of these things and leaves the default settings, feel free to add one. No one would object that.

peabee:
Perhaps the petbuild should be renamed to pcmanfm4puppy ?

Issue #2873, opened 2022-02-14
Petbuild of jwm fails


peabee:
Building ScPup32...........
```
Building jwm
checking for gcc... /usr/bin/ccache gcc
checking whether the C compiler works... no
configure: error: in `/tmp/jwm-2.3.7':
configure: error: C compiler cannot create executables
See `config.log' for more details
ERROR: failed to build jwm

# gcc --version
gcc (GCC) 10.3.0
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

```
Unable to find 'config.log' to see why....
Not sure what to check....

dimkr:
Your compiler is broken, not JWM and not the petbuild.

```
./configure || :
grep -C 10 error config.log
```

peabee:
sandbox3/petbuild-rootfs-complete is broken

/usr/bin/cpp is in sandbox3/devx

but is NOT in sandbox3/petbuild-rootfs-complete

```
# pwd
/mnt/sdc1/builds/sc32/t030222-woof-out_x86_x86_slackware_current/sandbox3/devx/usr/bin
# ls cpp
cpp
# 

# pwd
/mnt/sdc1/builds/sc32/t030222-woof-out_x86_x86_slackware_current/sandbox3/petbuild-rootfs-complete/usr/bin
# ls cpp
ls: cannot access 'cpp': No such file or directory
```
```
checking how to run the C preprocessor... /lib/cpp
configure: error: in `/mnt/sdc1/builds/sc32/t030222-woof-out_x86_x86_slackware_current/sandbox3/ccache-3.7.12':
configure: error: C preprocessor "/lib/cpp" fails sanity check
See `config.log' for more details
make: *** No targets specified and no makefile found.  Stop.
install: cannot stat 'ccache': No such file or directory
install: cannot stat '../petbuild-cache/ccache': No such file or directory

```

dimkr:
Does `chroot sandbox3/devx /lib/cpp --version` work?

peabee:
Yes (devx.sfs loaded):
```
# /lib/cpp --version
cpp (GCC) 11.2.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
[...]

PR #3012, opened 2022-04-01
Target i686 in petbuilds: libc targets i686 anyway

dimkr:
See https://github.com/puppylinux-woof-CE/woof-CE/issues/2873#issuecomment-1085556948.

Nobody runs a modern Puppy on an i486, i586 or K6, and the libc we take from other distros targets i686, so this `-march=i486` doesn't really make Puppy compatible with CPUs from the 90s.


peabee (approved):
Yes - thank you - fixes petbuilds for JammyPup32 ;-)

PR #3010, opened 2022-04-01
Add a minimal, dwl-based, kiosk Wayland compositor (fixes #3005)

dimkr:
tmp

This compositor allows more integration of Wayland-native elements into Puppy's traditional desktop. It supports layer-shell and the idle protocol, so swaylock can replace xlock (see #3007), gtkdialog's layer-shell support actually does something, and so on.

I want to experiment with the idea of running a fullscreen Xwayland in the background layer, run the ROX-Filer pinboard in it, but specify `XDG_SESSION_TYPE=wayland` and `GDK_BACKEND=wayland`, making all applications use their Wayland backend (unless they use GTK+ 2):

rox

PR #3014, opened 2022-04-02
Force children of JWM to use the X11 backend

dimkr:
More preparation for #3010. This will be useful soon, once dwl is patched to take more of JWM's responsibilities, like keyboard shortcuts. The applications started by .xinitrc, JWM and their children must use their X11 backend to integrate with JWM (for example, tray icons), while native Wayland applications started via the compositor should use the Wayland backend, or the compositor's "rootless" Xwayland as fallback.

### DIFF
--- a/woof-code/rootfs-petbuilds/cage/usr/bin/startxwayland
+++ b/woof-code/rootfs-petbuilds/cage/usr/bin/startxwayland
@@ -62,9 +62,6 @@ for F in $WAYLAND_DISPLAY $WAYLAND_DISPLAY.lock; do
 	mount --bind $XDG_RUNTIME_DIR/$F $SPOT_RUNTIME_DIR/$F
 done
 
-export GDK_BACKEND=x11
-unset SDL_VIDEODRIVER
-
 [ -e "$XDG_CONFIG_HOME/wmonitors/wmon_cmd" ] && . $XDG_CONFIG_HOME/wmonitors/wmon_cmd
 
 "$1"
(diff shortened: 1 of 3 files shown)


01micko:
I'm gathering that `XDG_SESSION_TYPE=wayland`? (set by wlroots)

dimkr:
> I'm gathering that `XDG_SESSION_TYPE=wayland`? (set by wlroots)

Yes, it should.

But .xinitrc always sets `XDG_SESSION_TYPE=x11`.

I want to make a second dwl petbuild, which doesn't run .xinitrc at all, so `XDG_SESSION_TYPE=wayland` and `GDK_BACKEND=wayland`. This PR is meant to decouple the compositor from .xinitrc, and allow dwl to be used in two mode, a "native Wayland" mode and a "fullscreen Xwayland" mode.

dimkr:
> I'm gathering that `XDG_SESSION_TYPE=wayland`? (set by wlroots)

Yes, it should.

But .xinitrc always sets `XDG_SESSION_TYPE=x11`.

I want to make a second dwl petbuild, which doesn't run .xinitrc at all, so `XDG_SESSION_TYPE=wayland` and `GDK_BACKEND=wayland`. This PR is meant to decouple the compositor from .xinitrc, and allow dwl to be used in two mode, a "native Wayland" mode and a "fullscreen Xwayland" mode.

01micko (approved):

PR #3022, opened 2022-04-03
Auto-enable swayidle and turn off all monitors

dimkr:
If ~/root/.config/swayidle/config is updated only when the compositor is started, it won't turn off a newly connected monitor.

Now, instead of building a swayidle configuration file in set_wl_monsleep, swayidle runs a script that turns off all monitors and a script that turns them back on and re-positions them.

In addition, the dwl session now runs swayidle, and the entire screensaver and screen locking mechanism is shared with labwc.


01micko:
Before we go too far the program [wlopm](https://sr.ht/~leon_plickat/wlopm/) should be considered in front of `wlr-randr` but it requires `zwlr-output-power-management-v1`

dimkr:
> Before we go too far the program [wlopm](https://sr.ht/~leon_plickat/wlopm/) should be considered in front of `wlr-randr` but it requires `zwlr-output-power-management-v1`

Should be easy to implement this protocol in dwl. But I think it's still a good idea to align labwc with dwl first, then do that.

EDIT: I'll try https://github.com/djpohly/dwl/compare/main...guidocella:output-power-management.patch

dimkr:
@01micko Done, the combination of swayidle, wlopm and swaylock works nicely on dwl.

01micko (approved):

PR #3025, opened 2022-04-04
Make the screensaver configuration menu entry work on Wayland

dimkr:
As a placeholder, `puplock` opens the swaylock configuration file in a text editor. In the future, it will become actually useful, or xlock_gui will be disabled in "pure Wayland" builds.


01micko:
All the old hard coded old school puppy code embedded into woof needs to go. I'll approve this but there is still plenty to be done.

01micko (approved):

PR #3033, opened 2022-04-06
When not building devx use existing petbuilds

peabee:
Building devx is slow and is not needed on every build.

### DIFF
--- a/woof-code/support/petbuilds.sh
+++ b/woof-code/support/petbuilds.sh
@@ -56,7 +56,10 @@ PKGS=
 # busybox must be first, so other petbuilds can use coreutils commands
 for NAME in $PETBUILDS; do
     HASH=`cat ../DISTRO_PKGS_SPECS-${DISTRO_BINARY_COMPAT}-${DISTRO_COMPAT_VERSION} ../DISTRO_COMPAT_REPOS ../DISTRO_COMPAT_REPOS-${DISTRO_BINARY_COMPAT}-${DISTRO_COMPAT_VERSION} ../DISTRO_PET_REPOS ../rootfs-petbuilds/${NAME}/petbuild 2>/dev/null | md5sum | awk '{print $1}'`
+    [ ! "$BUILD_DEVX" = "yes" ] && HASH="latest"
+    new_petbuild="no"
     if [ ! -d "../petbuild-output/${NAME}-${HASH}" ]; then
+        new_petbuild="yes"
         if [ $HAVE_ROOTFS -eq 0 ]; then
             echo "Preparing build environment"
             rm -rf petbuild-rootfs-complete
@@ -279,8 +282,7 @@ for NAME in $PETBUILDS; do
         find ../petbuild-output/${NAME}-${HASH} -name '.git*' -delete
     fi
 
-    rm -f ../petbuild-output/${NAME}-latest
-    ln -s ${NAME}-${HASH} ../petbuild-output/${NAME}-latest
+    [ "$new_petbuild" = "yes" ] && rm -f ../petbuild-output/${NAME}-latest && ln -s ${NAME}-${HASH} ../petbuild-output/${NAME}-latest
 
     PKGS="$PKGS $NAME"
 done


dimkr:
This is not a good solution, because if DISTRO_PKGS_SPECS has changed, and now devx can't be used to build a certain petbuild, you keep using the cached output from before the breakage, forever. Ensuring that devx is functional (so your build is reproducible) is one of the most important roles of petbuilds.sh.

(If you don't care about reproducibility, don't use petbuilds, just use .pet packages or something.)

The right solution is to avoid the extra overhead of building devx _when it's unused_, is to modify 3builddistro, so it doesn't build devx if `PETBUILDS` is non-empty, `BUILD_DEVX=no` and all petbuilds are built against the latest DISTRO_PKGS_SPECS.

dimkr (changes requested):
^

peabee:
Building the devx roughly doubles the build time and is not needed for many incremental builds.

/support/petbuilds.sh conflates 2 distinct functions:
1. performing petbuilds
2. copying petbuilds into the system build

Therefore if you want to do 2. but not do 1. (because you already have suitable petbuilds & you don't want to build devx) as things stand the logic must be put into /support/petbuilds.sh and cannot be put into 3builddistro.

So if the proposed changes are not acceptable, more extensive changes are needed so that functions 1. and 2. can be separated and called independently from 3builddistro.

> all petbuilds are built against the latest DISTRO_PKGS_SPECS

How can this be sensibly determined?? The hash-sum method seems too "sensitive" as it triggers if even the slightest change is made - it should really only be triggered if the devx components have changed.

dimkr:
> So if the proposed changes are not acceptable, more extensive changes are needed so that functions 1. and 2. can be separated and called independently from 3builddistro.

You can divide it into two scripts if you want. I don't see a good way to do this without much duplication of code. If you have an elegant solution - I'm all ears.

> How can this be sensibly determined?? The hash-sum method seems too "sensitive" as it triggers if even the slightest change is made - it should really only be triggered if the devx components have changed.

There are many corner cases, if you want to rebuild only if devx has changed. For example, if devx is unchanged, and `x-dev` is still included, but `x` is dropped from the main SFS, your previously built petbuild is broken (it will fail at runtime).

The are two main reasons why I haven't bothered handling these corner cases:
1. Using prebuilt petbuilds goes against the very essence of this feature. The idea behind petbuilds in woof-CE is to ensure every Puppy that replaces .pet packages with petbuilds, is reproducible by other woof-CE users, has all the latest fixes (for example, all the patches for the JWM menu generator) and has a functional devx that can be used to rebuild packages. If you wish to take the risk of a broken devx, don't want the extra fixes that go into petbuilds in the future, or don't care about reproducibility, just package your petbuild output as .pet packages and stick to these packages. petbuilds are not the right tool for your use case.
2. If the build takes too much time, you can use a faster disk, run woof-CE on a more powerful machine in the cloud, or just run woof-CE on GitHub Actions. If build times are the main issue, there's a variety of scalable solutions.

I'm strongly against the idea of crippling features, to speed up some use cases they were never designed for.

peabee:
No - I want to use petbuilds and to rebuild them when I rebuild devx. I see no need to rebuild them otherwise.

dimkr:
> No - I want to use petbuilds and to rebuild them when I rebuild devx. I see no need to rebuild them otherwise.

petbuilds.sh rebuilds the petbuilds **because** you changed devx. I explained why in my previous comment. The entire petbuilds mechanism is not suitable for your "build once and reuse forever" use case.

peabee:
That is NOT my use case....... it's "petbuild each time there is a new devx"..............

dimkr:
Do you have a way to reliably detect changes in devx, other than hashing DISTRO_PKGS_SPECS?

peabee:
Experience & knowledge............and monitoring what components have been updated

peabee:
How / why does the variable $PKGS at line 294 of petbuilds.sh differ from $PETBUILDS ?
Could line 294 be:
for NAME in $PETBUILDS; do
?

PR #3035, opened 2022-04-08
Make petbuilds.sh compatible with BUILD_DEVX=no

peabee:
All automated Woof-CE builds have BUILD_DEVX=yes so behaviour is unchanged.

Only applies to manual builds where BUILD_DEVX=no

### DIFF
--- a/woof-code/3builddistro
+++ b/woof-code/3builddistro
@@ -336,6 +336,7 @@ if [ -n "$PETBUILDS" ]; then
 		echo "Building without qemu-user-static"
 	fi
 
+	[ "$BUILD_DEVX" = "yes" ] && PETBUILD_ENABLE="enabled"
 	cd sandbox3
 	. ../support/petbuilds.sh
 	cd ..
--- a/woof-code/3builddistro
+++ b/woof-code/3builddistro
@@ -336,7 +336,7 @@ if [ -n "$PETBUILDS" ]; then
 		echo "Building without qemu-user-static"
 	fi
 
-	[ "$BUILD_DEVX" = "yes" ] && PETBUILD_ENABLE="enabled"
+	[ ! "$BUILD_DEVX" = "yes" ] && echo "WARNING - petbuilds require BUILD_DEVX=yes"
 	cd sandbox3
 	. ../support/petbuilds.sh
 	cd ..
--- a/woof-code/support/petbuilds.sh
+++ b/woof-code/support/petbuilds.sh
@@ -52,8 +52,11 @@ fi
 
 HERE=`pwd`
 PKGS=
+COPYPKGS="$PETBUILDS"
 
 # busybox must be first, so other petbuilds can use coreutils commands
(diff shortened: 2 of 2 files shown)


01micko:
What is the point of this?

`petbuilds` build something. To build something is to build something. If you don't have the bricks and mortar how can you build something?

Frankly, I don't care if it 'takes too long'. Build it on a raspberry pi or something less powerful, go to bed, wake up and voila! Built.

__The whole point of petbuild is to build packages.__

My plan in the begining when I created the original petbuilds was to incorporate it in the woof process, __and build fresh on *every* build__. @dimkr brought that to fruition in a far better way that I envisioned, but the same principle.

peabee:
The proposed changes DO NOT change anything for Woof-CE builds.........

They do allow the possibility for builders who do not want to build the devx and/or petbuilds every time to fulfill that desire!

Please support the changes - they are not negating any of the reasons you wanted petbuilds.

dimkr:
> They do allow the possibility for builders who do not want to build the devx and/or petbuilds every time to fulfill that desire!

You can't use petbuilds without building devx, by design. I've said this in your previous PR, and I'll say it again: petbuilds.sh makes sure petbuilds were built using the devx of the current woof-CE build, to protect you against shipping a broken devx, or a Puppy that uses packages that cannot be rebuilt.

Don't ruin this mechanism by wrapping it with a large `if`. What you did here is almost the same as skipping most of petbuilds.sh when `BUILD_DEVX=no`, and it's still a hack meant to make petbuilds.sh in a use case it wasn't designed for.

peabee:
> You can't use petbuilds without building devx, by design.

Indeed - and if you want to make new petbuilds you must set BUILD_DEVX=yes

> to protect you

I don't want to be protected thank you - if I make an error then I am happy to bear the consequences.

I hear all the (very negative) comments but I still want to go ahead with this change please.  It has NO IMPACT on existing Woof-CE builds but provides a degree of flexibility which I need. If I'm the only builder who makes use of this flexibility how does that impact anybody else?

dimkr:
You don't know if this change is going to impact others, this is an assumption.

If you add a quick n' dirty hack like this, at least fix the indentation of the code you wrapped with an `if`, add a warning, make builds on GitHub Actions fail if this hack is enabled, minimize the changes (`BUILD_DEVX=yes` entails `PETBUILD_ENABLE=enabled`, so you don't need both), and so on.

(I still oppose this PR, even if the changes are clean and minimal. But I would support a PR that auto-enables building of devx when it's needed to rebuild the petbuilds, but `BUILD_DEVX=no`, which is the right way to support your use case.)

peabee:
Requested changes made.

builds on GitHub Actions will fail with or without the changes if BUILD_DEVX=no

peabee:
I presume the absence of further comments means that I can merge this pull request now?

dimkr:
You didn't really address my comment, I think you misunderstood it. But I'm tired of this discussion and the constant resistance to changes in woof-CE from your side.

Feel free to merge it and even delete the entire petbuilds feature, I'm sick of this.

PR #3036, opened 2022-04-11
Add gtk2dialog symlink

peabee:
/rootfs-packages/jwm_config/usr/local/jwm_config/tray
now unconditionally uses gtk2dialog - see #2965
this change ensures gtk2dialog is present in a build - as a link to gtkdialog if necessary

### DIFF
--- a/woof-code/support/rootfs-hacks.sh
+++ b/woof-code/support/rootfs-hacks.sh
@@ -17,7 +17,7 @@ if [ ! -e ${SR}/usr/bin/rxvt-unicode ] && [ -f ${SR}/usr/bin/urxvt ] ; then
 fi
 
 # gtk2dialog symlink
-if [ ! -e ${SR}/usr/sbin/gtk2dialog ] && [ -f ${SR}/usr/sbin/gtkdialog ] ; then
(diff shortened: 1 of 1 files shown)


woof-code/support/rootfs-hacks.sh
```diff
@@ -16,6 +16,11 @@ if [ ! -e ${SR}/usr/bin/rxvt-unicode ] && [ -f ${SR}/usr/bin/urxvt ] ; then
 	ln -snfv urxvt ${SR}/usr/bin/rxvt-unicode
 fi
 
+# gtk2dialog symlink
+if [ ! -e ${SR}/usr/sbin/gtk2dialog ] && [ -f ${SR}/usr/sbin/gtkdialog ] ; then
```
dimkr: If this line is meant to deal with old gtkdialog .pet packages, you don't know if /usr/sbin/gtkdialog is a file. It could be a symlink, so you're better use `-e`.

(And I'd add a warning here, woof-CE can't fix others' problems forever. A user who insists on using some old .pet package should eventually rebuild it or start using a petbuild that's guaranteed to be compatible with woof-CE. This is a hack, and hacks are technical debt that needs to be removed at some point.)

PR #3040, opened 2022-04-18
Revert "Make petbuilds.sh compatible with BUILD_DEVX=no"

01micko:
Reverts puppylinux-woof-CE/woof-CE#3035

Don't just merge stuff!


dimkr:
See #3041, it cleans up the mess and protects builds covered by CI against breakage, while allowing @peabee to abuse this feature.

PR #3041, opened 2022-04-18
Fix bad indentation in petbuilds.sh and fail CI if affected by #3035

dimkr:


peabee:
@01micko has reverted #3035
as it made build (arm, debian, bullseye-veyron-speedy, 5.4.x-veyron-speedy) fail
although I can't see why....... as BUILD_DEVX=yes......

This seems to be a replacement for #3035 - thanks.

peabee (approved):
Seems to do what I wanted - thanks.

01micko (approved):

Issue #3043, opened 2022-04-19
/sys/fs/cgroup is missing in Puppy


dimkr:
This breaks some applications, like buildah.

PR #3044, opened 2022-04-19
Make buildah and podman work in jammy64 (fixes #3042, #3043)

### LINKED ISSUE #3042: /var/lib/containers should be a bind mount
Various container-related tools, including buildah and podman, expect the file system at /var/lib/containers to be something other than aufs or overlay, because they use it as a layer in another layered file system. Therefore, if this directory exists, it should be a bind mount to pup_rw or something else that's outside of /.

dimkr:
Various development tools run containers, and it's good to have stuff like https://github.com/89luca89/distrobox working on Puppy.

`apt install -y podman && run-as-spot podman run --rm -it alpine:3.15` works:

podman1
podman2

~This is not perfect, some things are still broken or spit errors/warnings. But this PR fixes 80% of the big problems, with minimum changes.~ Should work well

~~EDIT: one thing is still problematic, for some reason /dev/net is 600 and spot can't access /dev/net/tun~~


woof-code/rootfs-skeleton/etc/rc.d/rc.sysinit
```diff
@@ -557,15 +574,20 @@ udevadm trigger #--action=add --subsystem-match="input" --subsystem-match="sound
...
+) |
 while read F1 F2 F3 F4 F5 F6 F7 F8 ETC ; do
 	case $F1 in
 		"c!"*|"b!"*|"u!"*|"p!"*) ok=1 ;;
+		"z"*) chmod $F3 $F2 2>/dev/null ;;
 		*) continue ;;
 	esac
 	[ -z "$F7" ] && continue #precaution
 	# mknod: /dev/net/tun: No such file or directory
-	[ ! -d ${F2%/*} ] && mkdir -p -m $F3 ${F2%/*}
```
dimkr, line 568: Huge bug in this line. If /dev/net/tun is missing and needs to be 666, but /dev/net is missing, the latter should be 755 and not 666. Otherwise, spot can't `ls /dev/net` and /dev/net/tun is inaccessible.

01micko (approved):

PR #3067, opened 2022-04-30
Replace /root with /home/spot in XDG_* environment variables

dimkr:
Flatpak uses XDG_DATA_DIRS, and run-as-spot doesn't set this one.

(It's a file in /etc/profile.d that adds a directory under /root to XDG_DATA_DIRS, but we don't run a login shell inside run-as-spot, so this file doesn't get sourced again.)


01micko (approved):

01micko:
Hmm, found a typo in CONTRIBUTING.md

`s/there/their/`

PR #3056, opened 2022-04-25
ntfs support improved / corrected

peabee:
https://github.com/puppylinux-woof-CE/woof-CE/blob/2a23bfd7efa4b28f66255dbc2f87adaf30f761e0/kernel-kit/configs_x86/DOTconfig-5.10.60-i686-pae#L8809

ntfs support was not included

https://github.com/puppylinux-woof-CE/woof-CE/blob/2a23bfd7efa4b28f66255dbc2f87adaf30f761e0/kernel-kit/configs_x86_64/DOTconfig-5.10.60-x86_64#L8714

ntfs support was inconsistent

PR #3068, opened 2022-05-01
Change to CONFIG_FUSE_FS=y, to make ntfs-3g work

### LINKED ISSUE #3058: ntfs boot failure with kernels 5.10 and 5.15
Even after #3056 
slacko-8.2.2-174.iso  ([build-x86-slackware-15.0-174])
fails to boot from an ntfs formatted boot partition
5.10 and 5.15 both fail
5.4 is OK

dimkr:
This _should_ fix #3058.


peabee (changes requested):
5.4 has both
CONFIG_FUSE_FS=y
and
CONFIG_NTFS_FS=y
[hash] CONFIG_NTFS_DEBUG is not set
CONFIG_NTFS_RW=y

Shouldn't 5.10 and 5.15 be the same??

dimkr:
CONFIG_NTFS_FS=y doesn't matter, because Puppy uses ntfs-3g (a FUSE-based file system). #3056 didn't do anything.

peabee:
Lets try it - but I do feel that 5.4 5.10 and 5.15 should be as consistent as possible

**Note**: 5.15.36 aufs patches need an update and so 5.15 build will fail until released:
https://github.com/sfjro/aufs5-standalone/issues/12#issuecomment-1114352961

dimkr:
> Lets try it - but I do feel that 5.4 5.10 and 5.15 should be as consistent as possible

I disagree, the 5.4.x configuration has many unused drivers, like the NTFS one. It's a very crude configuration file.

PR #3087, opened 2022-05-15
xorg-autoconf: Added nvidia kernel module for modesetting

rizalmart:
Since the announcement of nvidia open source driver. Using modesetting might be feasible.

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/xorg-autoconf
+++ b/woof-code/rootfs-skeleton/usr/sbin/xorg-autoconf
@@ -621,7 +621,7 @@ drvfound=""
 [ -e $XORG_MODULE_PATH/drivers/${drv}_drv.so ] && drvfound="1"
 
 if [ "$drvfound" == "" ]; then
- if [ "$(echo "$MODLIST" | awk '{ print $1 }' | grep -E "^nouveau")" != "" ]; then
+ if [ "$(echo "$MODLIST" | awk '{ print $1 }' | grep -E "^nouveau|^nvidia")" != "" ]; then
   drv="modesetting"
  fi
 fi


woof-code/rootfs-skeleton/usr/sbin/xorg-autoconf
```diff
@@ -621,7 +621,7 @@ drvfound=""
 [ -e $XORG_MODULE_PATH/drivers/${drv}_drv.so ] && drvfound="1"
 
 if [ "$drvfound" == "" ]; then
- if [ "$(echo "$MODLIST" | awk '{ print $1 }' | grep -E "^nouveau")" != "" ]; then
+ if [ "$(echo "$MODLIST" | awk '{ print $1 }' | grep -E "^nouveau|^nvidia")" != "" ]; then
```
dimkr, line 624: Most users will stick to the closed-source one. How do you know which one is used? And, does it matter in first place?

rizalmart: @dimkr 

If you see the code, modesetting on nvidia videocard was just a second fallback if the nouveau or nvidia xorg driver was missing but nouveau/nvidia kernel module was loaded. Otherwise it will use vesa or fbdev if nvidia videocard was present but no nouveau/nvidia kernel module found. Loading those kernel modules was depend on the udev. Newer nvidia gpu's will have an official open source driver from nvidia. So using modesetting as fallback makes sense.

dimkr: But what about the closed source one? Most users have the closed source one, because they don't have a high-end, recent GPU. And NVIDIA's plan is to keep the closed source driver around, at least for some time. Does modesetting work with the closed source driver? If both drivers are called `nvidia`, maybe this logic can introduce problems.

(But I trust your judgement - I don't use X.Org myself, I'm only trying to help.)

01micko: @rizalmart any rebuttal for @dimkr 's comment? I don't mind the PR but I'd like to hear your response.

PR #3090, opened 2022-05-16
Drop support for PUPMODE 6

dimkr:
Users are having a hard time to adjust to anything other than ISO images and initrd.gz-less boot. And it looks like user don't care much about security and stability updates, or having a working, built-in installer.

If this code is unused (except in Vanilla Dpup 9.1.x, which has its own woof-CE branch), let's just remove it. The code that generates flash drive images instead of ISO images is still useful, though, so I'm keeping it.


mavrothal:
I would disagree with the drop.
I would actually argue to merge #2963

You have to realize that the 30-50 active forum members are mostly aspiring developers and/or active testers. So having multiple (but similar) pups and common routines that they can “debug”, is important for them.
Also continuous positive/friendly  interaction with the puppy builder is important.
Is more of a watering hole rather than a standard distro forum.

The code you propose to delete does not bother anything and would be easier to be adapted by another interested puppy builder if it is already there.

dimkr:
> I would disagree with the drop.
> I would actually argue to merge https://github.com/puppylinux-woof-CE/woof-CE/pull/2963

Why not both? Nobody uses this feature except me, and I want to drop PUPMODE 6 and frugalify in favor of overlay support.

The vast majority of users prefers a traditional Puppy release with initrd.gz, psubdir, etc'. Users are much more likely to adopt overlay if it's supported "natively" by initrd.gz, and not via PUPMODE 6.

> The code you propose to delete does not bother anything and would be easier to be adapted by another interested puppy builder if it is already there.

PUPMODE 6 is always a corner case - sometimes the desired behavior is same as PUPMODE 2, and sometimes it requires special handling. Less PUPMODEs means less code, less issues and easier maintenance, so I do see reasons why remove this code. (And again, nobody uses it except Vanilla Dpup 9.1.x, which has its own stable woof-CE branch anyway.)

dimkr:
(squashed this PR, so it's a single commit that can be reverted in case anyone wants PUPMODE 6 support to be restored)

EDIT: @mavrothal do you still have objections?

dimkr:
Vanilla Dpup 9.0.x and 9.1.x are dead, and nobody asks for this feature. Everyone just wants ISO releases with initrd.gz, and PUPMODE 6 (+frugalify) is nothing but a maintenance burden which needs special handling all the time.

If this changes in the future, this PR is a single commit that can be reverted.

PR #3113, opened 2022-05-26
Update dwl to 52e0d00

dimkr:
dwl now allows compile-time choice of the acceleration profile. Because the petbuild patches dwl so it takes these settings from environment variables, now the acceleration profile is configurable:

profile


01micko (approved):
Looks nice.

PR #3128, opened 2022-05-31
Set the default icon for windows without an icon

dimkr:
Most windows do have an icon, but those that don't use this ugly black rectangle. A Puppy-style icon looks much better.

Before:
before

After:
after


01micko (approved):

zigbert:
Nice!

PR #3129, opened 2022-05-31
snapmergepuppy: Exclude internet cache

rizalmart:


01micko (approved):

dimkr:
Please don't do this, losing the browser cache on every reboot will kill so many data plans 😿

PR #3130, opened 2022-06-01
Use the same input/keyboard settings across X.Org and Wayland

dimkr:
Puppy has no way to configure touchpad or mouse settings, and users try to use flsynclient. It's old and doesn't work with today's libinput based driver, so users have to fall back to the old xserver-xorg-input-synaptics or edit xorg.conf manually.

If both X.Org and Wayland compositors use libinput, they can share one configuration file and one wizard.

At some point in the future (when nobody uses the old X drivers anymore), the old code path that modifies xorg.conf, can be removed.

In addition, Wayland compositors use xkbcommon and support the same configuration keys and values as X, so keyboard settings can be moved to a shared location.

This PR moves keyboard-related settings away from xorg.conf, to a shared location that can be used by dwl and other compositors that respect the `XKB_DEFAULT_*` environment variables.

This will make it easier to remove xorgwizard in the future.

wizard


01micko (approved):
I don't know if `labwc` will like this but I'm prepared to give it a try.

dimkr:
> I don't know if `labwc` will like this but I'm prepared to give it a try.

It should, all wlroots based compositors use XKB_DEFAULT_LAYOUT, etc' unless they're misconfigured and use `""` instead of `NULL` as the default.

dimkr:
Forgot to mention, `LIBINPUT_DEFAULT_*` are supported in dwl only because I patched it. This is not something standard: labwc will need an alternative solution for input settings, or a patch similar to the dwl one. This PR unifies keyboard and pointer settings across X and dwl, while labwc respects only the keyboard part.

01micko:
Yeah `labwc` input config is in `~/.config/labwc/rc.xml` so I'll probably just add a block there and parse the values from `~/.inputrc`

dimkr:
Squashed, ready to merge. If it causes problems, this single commit can be easily reverted later.

PR #3132, opened 2022-06-02
Update JWM to 2.4.2

dimkr:
Seems to work just fine, after a quick glance and few theme switches.


01micko (approved):
I might be biased because I had a hand in fixing some of the 2.4.1 bugs :rofl:

peabee (commented):
Generally in favour however, what does

> and few theme switches

mean? Are there consequential changes? Have they been tested? If not, how do we test?
pTheme?
JWMDesk?
Themes?
all the JWM related packages in http://distro.ibiblio.org/puppylinux/pet_packages-noarch/ ?

Maybe this is a case where an update .pet should be made available on the forum for testing and feedback??

peabee:
Perhaps the existing petbuild should be renamed jwm237 and this should be a new petbuild jwm242.

jwm could be a symlink to one or the other depending on whether legacy builds could move to 242 ?

peabee:
I did a test build of JammyPup32 with jwm242........

Selecting theme Original-Pup in pTheme crashed out to the terminal............

peabee (changes requested):
More testing is needed

dimkr:
> Generally in favour however, what does
> 
> > and few theme switches
> 
> mean?

Changing themes using ptheme. All 5 themes I tried (412, 431, Flat-grey, Buntoo, Tahrpup) worked with zero changes in ptheme or the JWM themes.

EDIT: Stardust_bright, the JWM theme used by Original Pup, worked too.

peabee:
The problem comes from the wallpaper
Sky.svg
which Original Pup uses
which 237 is fine with
but it crashes 242

dimkr:
Should be fixed by https://github.com/joewing/jwm/pull/572.

01micko (approved):

bigpup1:
Need to look at reported issues in this forum topic:
https://forum.puppylinux.com/viewtopic.php?t=6091

dimkr:
https://github.com/joewing/jwm/pull/577 should fix broken dynamic menus.

dimkr:
@peabee More objections to this PR?

peabee:
Has sufficient testing been accomplished?
Have all problems reported on the forum been fixed?
Still feel it would be prudent to call this petbuild jwm242 and rename the current jwm -> jwm237 and fix legacy builds to use jwm237

dimkr:
> Has sufficient testing been accomplished?

I think so.

> Have all problems reported on the forum been fixed?

As far as I know, yes.

> Still feel it would be prudent to call this petbuild jwm242 and rename the current jwm -> jwm237 and fix legacy builds to use jwm237

If 2.4.x has no known issues, why keep 2.3.7 (which has known issues) around?

peabee:
> > Have all problems reported on the forum been fixed?
> 
> As far as I know, yes.
> 
??
https://forum.puppylinux.com/viewtopic.php?p=59101#p59101
https://forum.puppylinux.com/viewtopic.php?p=59118#p59118

dimkr:
> > > Have all problems reported on the forum been fixed?
> > 
> > 
> > As far as I know, yes.
> 
> ?? https://forum.puppylinux.com/viewtopic.php?p=59101#p59101 https://forum.puppylinux.com/viewtopic.php?p=59118#p59118

These are jwm_config issues that affect 2.3.x as well.

dimkr:
Still works fine, and I'm not getting any 2.4.x-specific complains from users.

peabee (approved):

PR #3139, opened 2022-06-04
Default to UTC for the system clock

dimkr:
If I'm not mistaken, distros with systemd (= all popular distros) do that, and this should make life easier for those who dual-boot.

### DIFF
--- a/woof-code/rootfs-skeleton/etc/clock
+++ b/woof-code/rootfs-skeleton/etc/clock
@@ -1,5 +1,5 @@
 #Set this to either 'utc' or 'localtime' based on which one your computer's
 #hardware clock uses.
-HWCLOCKTIME=localtime
-#HWCLOCKTIME=utc
+HWCLOCKTIME=utc
+#HWCLOCKTIME=localtime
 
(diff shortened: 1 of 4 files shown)


01micko (commented):
This makes the checkbox in `quicksetup` default to true.

dimkr:
> This makes the checkbox in `quicksetup` default to true.

That's intentional. The assumption is that most users either use only Puppy (so it doesn't matter if it's UTC or not) or Puppy and other OSs that use UTC (so "true" is the sane default). Those who dual boot Puppy and Windows (the corner case where using localtime in Puppy can be useful) will have to configure Windows to use UTC, or untick the checkbox.

01micko:
> > This makes the checkbox in `quicksetup` default to true.
> 
> That's intentional. The assumption is that most users either use only Puppy (so it doesn't matter if it's UTC or not) or Puppy and other OSs that use UTC (so "true" is the sane default). Those who dual boot Puppy and Windows (the corner case where using localtime in Puppy can be useful) will have to configure Windows to use UTC, or untick the checkbox.

I'm not so sure that dual booting puppy and windows is such a corner case. I think it may be more common than you think, or even less common than I think, so maybe we should do a forum poll or something before this is merged.

Don't get me wrong, I don't really care either way, but some might.

PR #3142, opened 2022-06-05
Allow "pure Wayland mode" with dwl using GDK_BACKEND=wayland in ~/.dwlrc

dimkr:
In this weird mode (disabled by default, until it's ready), the maximized Xwayland window is still there, and so is JWM. However, all windows started by JWM are native Wayland windows: JWM is unaware of their existence and they float on top of the Xwayland window.

The Puppy themes ported to GTK+ 3 have proper theming for CSD, and window decorations are more-or-less visually consistent with other UI elements.

The next step is to add some Wayland panel (yambar? sfwbar?), start it in ~/.dwlinitrc only if GDK_BACKEND=wayland, and disable JWM's trays when GDK_BACKEND=wayland. This will make JWM 100% invisible unless you start some legacy GTK+ 2 application, which runs in the Xwayland window.

I'm still working on implementing JWM-like keyboard shortcuts directly in dwl, and I want to disable them if GDK_BACKEND=x11 (to let JWM handle keyboard shortcuts).

native


01micko:
`sfwbar` with battery, keyboard and cpu temperature 'applets' (all my scripts hacked in to the config)

sfwbar-recent

dimkr:
> `sfwbar` with battery, keyboard and cpu temperature 'applets' (all my scripts hacked in to the config)
> 
> sfwbar-recent

I got sfwbar to work with dwl, I have an early version of wlr-foreign-toplevel support in a tiny patch. Maybe this PR should wait until 1) I add this protocol and 2) we have a sfwbar petbuild with Puppified configuration files, so I can start it in ~/.dwlinitrc.

dimkr:
Here's a demo of legacy GTK+ 2 applications (HexChat) in this "pure Wayland" mode. dwl is built without Xwayland support, DISPLAY is set and GTK+ 2 applications run inside the fullscreen Xwayland window, below native Wayland windows. JWM manages these X windows, not dwl.

hexchat

01micko (approved):

PR #3151, opened 2022-06-11
Move yambar to the top layer

dimkr:
This makes yambar more consistent with JWM's tray.

### DIFF
--- a/woof-code/rootfs-petbuilds/yambar/root/.config/yambar/config.yml
+++ b/woof-code/rootfs-petbuilds/yambar/root/.config/yambar/config.yml
@@ -16,6 +16,7 @@ bar:
   location: top
   spacing: 5
   margin: 7
+  layer: top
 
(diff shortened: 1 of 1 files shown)


01micko:
Does it work fine with fullscreen apps? If so then I approve.

dimkr:
> Does it work fine with fullscreen apps? If so then I approve.

I'll have to check. There are some behavior differences between labwc and dwl, and I'm not sure if their layer shell implementations are true to the spec.

dimkr:
> Does it work fine with fullscreen apps? If so then I approve.

It doesn't work fine: for some reason, dwl ignores yambar's request and maximized applications cover the entire monitor. Also, yambar sits on top of fullscreen applications.

01micko:
For some ideas see https://github.com/LBCrion/sfwbar/issues/24

PR #3153, opened 2022-06-12
Add bookworm64 template with dwl in "pure Wayland" mode

dimkr:
bullseye64 uses X.Org, jammy64 uses dwl in "Xwayland mode" and sid64 uses labwc. The first two are in good shape, usable as daily drivers, and sid64 is a good place to work on labwc integration.

I want to start working on Vanilla Dpup 10.x, based on Debian Bookworm and dwl in "Wayland mode". It needs a panel with theme support, keyboard shortcuts and various window management improvements in dwl (like smarter window placement).


01micko (approved):
A good starting point. No doubt stuff will change as development advances.

01micko (approved):

PR #3178, opened 2022-07-01
Avoid killing Xwayland to prevent SIGHUP in wmexit

dimkr:
wmexit kills Xwayland, which creates a race condition. wmexit receives a SIGHUP, and sometimes, this signal is received before wmexit kills dwl, leaving the user with a black screen with a cursor (because Xwayland, JWM, etc' are killed, while dwl is still running, with zero windows).

### DIFF
--- a/woof-code/rootfs-skeleton/usr/bin/wmexit
+++ b/woof-code/rootfs-skeleton/usr/bin/wmexit
@@ -71,7 +71,7 @@ sleep 0.2
 sync
 
          #xfce       #lxde     #LXQt        #window managers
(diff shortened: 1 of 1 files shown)


01micko (commented):
Why isn't `dwl` killed?

dimkr:
> Why isn't `dwl` killed?

If wmexit kills Xwayland and wmexit runs in a pty (for example, inside lxterminal), sometimes wmexit gets killed by a SIGHUP before it reaches the point where it kills dwl. The probability is at least 20% on my machine.

Therefore, wmexit should kill dwl **without killing Xwayland**: it will die when the compositor dies.

01micko (approved):

PR #3191, opened 2022-07-06
Prevent WeeChat from advertising its version number

dimkr:
This is bad for privacy (makes users easier to fingerprint) and security (makes it easier to detect known vulnerabilities).

After this PR, WeeChat no longer gives away information like its version number and the client's local time (tested with two clients, one before and one after this PR):

weechat


01micko (approved):

01micko:
This PR looks the goods but I have problem with missing python stuff in slacko. I'll open a separate issue as time permits.

PR #3209, opened 2022-07-17
Add support for dwl themes

dimkr:
This continues #3208 and adds window borders that match the yambar theme:

borders


01micko (approved):
Nice! I have some labwc themes to merge soon.

PR #3212, opened 2022-07-17
Add a yambar battery monitor based on powerapplet_tray

dimkr:


01micko (approved):
[Ha!](https://github.com/LBCrion/sfwbar/issues/32#issuecomment-1186832332)

PR #3213, opened 2022-07-17
Add dwl support in logout_gui

dimkr:


01micko (approved):
I have similar for labwc .. soon..

PR #3216, opened 2022-07-18
Add a foot package template

dimkr:


01micko (approved):
I really like foot. Daniel has done a great job with it.

dimkr:
> I really like foot. Daniel has done a great job with it.

I like it, too. It works very well with a tiling window manager. It's a good rxvt-unicode replacement: it doesn't have a tab bar that wastes spaces, and it supports both -e and --hold. I'll see if the lxterminal urxvt wrapper can be replaced with a symlink to foot.

dimkr:
Changed to `TERM=xterm-256color` so no additional terminfo files need to be added. This minimizes the risk of inconsistency between lxterminal and foot.

PR #3258, opened 2022-07-23
Try x-www-browser before Firefox

dimkr:
Users who don't like the preinstalled Firefox don't have to change defaultbrowser once they install their browser of choice, if defaultbrowser tries x-www-browser first.

### DIFF
--- a/woof-code/packages-templates/firefox_FIXUPHACK
+++ b/woof-code/packages-templates/firefox_FIXUPHACK
@@ -83,9 +83,18 @@ EOF
 
 (
 echo "echo '"'#!/bin/sh'"' > usr/local/bin/defaultbrowser"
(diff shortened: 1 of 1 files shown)


01micko (commented):
How does this work? puppy-apps?

dimkr:
> How does this work? puppy-apps?

Nope, it's a Debian thing (the "alternatives" mechanism).

dimkr:
Tested with Chrome, under bookworm64. The official Chrome package replaces the x-www-browser symlink, so the user doesn't have to do anything to make Chrome the default browser.

(btw, it uses Xwayland by default but supports Wayland with a command-line argument - works nicely either way)

01micko (approved):

PR #3277, opened 2022-07-26
Kill PipeWire before starting it

dimkr:
This is weird, pipewire and pipewire-pulse keep running after the spot session bus is killed, and after tty1 is dead. As a result, audio is broken until they're killed and restarted.


01micko (approved):
This probably affects labwc too.

dimkr:
> This probably affects labwc too.

startdwl and startlabwc need to be unified at some point, and that will fix this.

PR #3282, opened 2022-07-28
Use gsimplecal as a yambar calendar applet

dimkr:
gsimplecal


01micko:
I'm fine with this but why not use `timebar`?

dimkr:
> I'm fine with this but why not use `timebar`?

gsimplecal is small, resembles minixcal and closes when it loses focus. I think it's a better fit for a tiling compositor.

01micko (approved):

Issue #3320, opened 2022-08-09
sfs_load handling of symlinks is wrong


peabee:
#3027 from @rizalmart introduced checks on symlinks when loading sfs

For VoidPup64:
/lib and /usr/lib are real directories
/lib64 and /usr/lib64 are symlinks

The devx for VoidPup64 only contains real directories /lib & /usr/lib but fails to load (erroneously) due the checks in sfs_load

sfs_load should only fail if the sfs contains real directories which are going to overwrite symlinks

Maybe @dimkr can have a look?

dimkr:
That change in sfs_load is incompatible with VoidPup, that's all I can say. The only solution I see is removing all protections of this kind, because VoidPup is incompatible with both USR_SYMLINKS=yes and USR_SYMLINKS=no.

peabee:
The sfs_load logic appears to be wrong to me.........

At line 52 the checks are initiated if any of a list of symlinks is present in the running system however the symlink that has triggered the check is not remembered.

Then at line 65 the contents of the sfs are checked to see if any of the same list of objects is both a directory and not a symlink inside the sfs.

There is no logical connection between the 2 tests.

In the VoidPup64 devx case the first test is triggered by the presence of the /lib64 symlink whereas the 2nd test  is triggered by the /lib directory inside the devx.

The VoidPup64 devx sfs is correctly configured and can be safely installed but the erroneous sfs_load logic is stopping it being loaded.

peabee:
I propose a change lines 52-83:
```
sym_list=""
for fld2 in bin sbin lib lib64 lib32 libx32
do
  [ -L /$fld2 ] && sym_list="$sym_list $fld2"
done

if [ "$sym_list" != "" ]; then

 if [ "$1" ]; then
 
   SFSFILE="$(basename "$1")"
   MNTFLD="$(echo "${SFSFILE}-module" | sed -e 's#\ #_#g')"
   mkdir -p /mnt/"$MNTFLD"
   DEVLOOP="$(losetup -f)"
   losetup "$DEVLOOP" "$1"
   mount -r -t squashfs -o loop,ro "$DEVLOOP" /mnt/$MNTFLD
  
   FLD_INVALID=""
  
    for fld2 in $sym_list
    do
    if [ -d /mnt/$MNTFLD/$fld2 -a ! -L /mnt/$MNTFLD/$fld2 ]; then
      FLD_INVALID="$fld2"
      break
     fi
    done
    
[...]

PR #3027, opened 2022-04-05
sfs_load: Added handling of usrmerge

rizalmart:
Stop loading non-usrmerge file system hierarchy compliant SFS modules if the main puppy sfs was usrmerge compliant.

Fixes #3006 issue for sfs load

PR #3323, opened 2022-08-10
Update sfs_load

peabee:
Fixes Issue #3320

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/sfs_load
+++ b/woof-code/rootfs-skeleton/usr/sbin/sfs_load
@@ -50,7 +50,7 @@
 . /etc/rc.d/PUPSTATE
 
 sym_list=""
-for fld2 in bin sbin lib lib64 lib32 libx32
+for fld2 in bin sbin lib lib64 lib32 libx32 usr/bin usr/sbin usr/lib usr/lib64 usr/lib32 usr/libx32
 do
   [ -L /$fld2 ] && sym_list="$sym_list $fld2"
 done
@@ -80,8 +80,8 @@ if [ "$sym_list" != "" ]; then
     rm -rf /mnt/$MNTFLD
        
     if [ "$FLD_INVALID" != "" ]; then
-      /usr/lib/gtkdialog/box_yesno "sfs_load" warning "$(gettext "The directory $FLD_INVALID in this sfs will overwrite the system symlink. This may cause problems. Do you want to continue?")"
-      [ $? == 1 ] && exit 1
+      /usr/lib/gtkdialog/box_yesno "sfs_load" warning "/$FLD_INVALID $(gettext "directory in this sfs will overwrite the system symlink. This may cause problems. Do you want to continue?")"
+      [ $? -eq 1 ] && exit 1
     fi
  
  fi
--- a/woof-code/rootfs-skeleton/usr/sbin/sfs_load
+++ b/woof-code/rootfs-skeleton/usr/sbin/sfs_load
@@ -49,7 +49,13 @@
 
 . /etc/rc.d/PUPSTATE
 
-if [ -L /bin -o -L /sbin -o -L /lib -o -L /lib64 -o -L /lib32 -o -L /libx32 ]; then
+sym_list=""
+for fld2 in bin sbin lib lib64 lib32 libx32
+do
+  [ -L /$fld2 ] && sym_list="$sym_list $fld2"
+done
+
+if [ "$sym_list" != "" ]; then
 
  if [ "$1" ]; then
  
@@ -62,10 +68,10 @@ if [ -L /bin -o -L /sbin -o -L /lib -o -L /lib64 -o -L /lib32 -o -L /libx32 ]; t
   
    FLD_INVALID=""
   
-    for fld2 in bin sbin lib lib64 lib32 libx32
+    for fld2 in $sym_list
     do
     if [ -d /mnt/$MNTFLD/$fld2 -a ! -L /mnt/$MNTFLD/$fld2 ]; then
-      FLD_INVALID="y"
+      FLD_INVALID="$fld2"
       break
      fi
     done
@@ -74,14 +80,13 @@ if [ -L /bin -o -L /sbin -o -L /lib -o -L /lib64 -o -L /lib32 -o -L /libx32 ]; t
     rm -rf /mnt/$MNTFLD
        
     if [ "$FLD_INVALID" != "" ]; then
-      /usr/lib/gtkdialog/box_ok "sfs_load" error "$(gettext "This sfs should not be loaded because it has the wrong file structure")"
-      exit 1
+      /usr/lib/gtkdialog/box_yesno "sfs_load" warning "$(gettext "The directory $FLD_INVALID in this sfs will overwrite the system symlink. This may cause problems. Do you want to continue?")"
+      [ $? == 1 ] && exit 1
     fi
  
  fi
  
 fi 
-
 
 if [ "$PUNIONFS" = "overlay" ] ; then
 	exec sfs_load.overlay "$@"
--- a/woof-code/rootfs-skeleton/usr/sbin/sfs_load
+++ b/woof-code/rootfs-skeleton/usr/sbin/sfs_load
@@ -80,8 +80,8 @@ if [ "$sym_list" != "" ]; then
     rm -rf /mnt/$MNTFLD
        
     if [ "$FLD_INVALID" != "" ]; then
-      /usr/lib/gtkdialog/box_yesno "sfs_load" warning "/$FLD_INVALID $(gettext "directory in this sfs will overwrite the system symlink. This may cause problems. Do you want to continue?")"
-      [ $? -eq 1 ] && exit 1
+      /usr/lib/gtkdialog/box_ok "sfs_load" warning "/$FLD_INVALID $(gettext "directory in this sfs will overwrite the system symlink. This will cause problems. Unable to continue with load.")"
+      exit 1
     fi
  
  fi


woof-code/rootfs-skeleton/usr/sbin/sfs_load
```diff
@@ -74,15 +80,14 @@ if [ -L /bin -o -L /sbin -o -L /lib -o -L /lib64 -o -L /lib32 -o -L /libx32 ]; t
     rm -rf /mnt/$MNTFLD
        
     if [ "$FLD_INVALID" != "" ]; then
-      /usr/lib/gtkdialog/box_ok "sfs_load" error "$(gettext "This sfs should not be loaded because it has the wrong file structure")"
-      exit 1
+      /usr/lib/gtkdialog/box_yesno "sfs_load" warning "$(gettext "The directory $FLD_INVALID in this sfs will overwrite the system symlink. This may cause problems. Do you want to continue?")"
+      [ $? == 1 ] && exit 1
```
mavrothal: In which case would you think a user would like to overwrite with links and break the system?

woof-code/rootfs-skeleton/usr/sbin/sfs_load
```diff
@@ -62,10 +68,10 @@ if [ -L /bin -o -L /sbin -o -L /lib -o -L /lib64 -o -L /lib32 -o -L /libx32 ]; t
   
    FLD_INVALID=""
   
-    for fld2 in bin sbin lib lib64 lib32 libx32
+    for fld2 in $sym_list
     do
     if [ -d /mnt/$MNTFLD/$fld2 -a ! -L /mnt/$MNTFLD/$fld2 ]; then
```
dimkr, line 73: The logic here is flawed, if I understand it correctly. If /libx32 is a directory but devx/libx32 doesn't exist, sfs_load tells the user the devx SFS is "invalid".

woof-code/rootfs-skeleton/usr/sbin/sfs_load
dimkr: `$? -eq 1`

woof-code/rootfs-skeleton/usr/sbin/sfs_load
```diff
@@ -49,7 +49,13 @@
...
 . /etc/rc.d/PUPSTATE
 
-if [ -L /bin -o -L /sbin -o -L /lib -o -L /lib64 -o -L /lib32 -o -L /libx32 ]; then
+sym_list=""
+for fld2 in bin sbin lib lib64 lib32 libx32
+do
+  [ -L /$fld2 ] && sym_list="$sym_list $fld2"
+done
+
+if [ "$sym_list" != "" ]; then
```
dimkr: This entire block should be after `if [ "$1" ]; then`: there's no need to slow down sfs_load if we're not loading a SFS

woof-code/rootfs-skeleton/usr/sbin/sfs_load
```diff
@@ -74,15 +80,14 @@ if [ -L /bin -o -L /sbin -o -L /lib -o -L /lib64 -o -L /lib32 -o -L /libx32 ]; t
     rm -rf /mnt/$MNTFLD
        
     if [ "$FLD_INVALID" != "" ]; then
-      /usr/lib/gtkdialog/box_ok "sfs_load" error "$(gettext "This sfs should not be loaded because it has the wrong file structure")"
-      exit 1
+      /usr/lib/gtkdialog/box_yesno "sfs_load" warning "$(gettext "The directory $FLD_INVALID in this sfs will overwrite the system symlink. This may cause problems. Do you want to continue?")"
```
dimkr: This string is not translatable. It's not a fixed string that can go into .po/.mo files, because it contains `$FLD_INVALID`.

woof-code/rootfs-skeleton/usr/sbin/sfs_load
dimkr, line 58: `if [ -n "$sym_list" ]; then`

woof-code/rootfs-skeleton/usr/sbin/sfs_load
```diff
@@ -49,7 +49,13 @@
 
 . /etc/rc.d/PUPSTATE
 
-if [ -L /bin -o -L /sbin -o -L /lib -o -L /lib64 -o -L /lib32 -o -L /libx32 ]; then
+sym_list=""
+for fld2 in bin sbin lib lib64 lib32 libx32
```
dimkr: Is it enough? /usr/lib64 can be a symlink, too.

woof-code/rootfs-skeleton/usr/sbin/sfs_load
dimkr: That's a good question, #3027 was meant to protect users. If this SFS is dangerous, there's no reason to allow loading. If it's not dangerous, there's no reason to do this check.

woof-code/rootfs-skeleton/usr/sbin/sfs_load
peabee: Replying to all 7 comments made over a very short period of time.........
1. libx32 as a symlink with no libx32 directory in sfs works fine (tested)
2. the test is "-a" - both directory must exist AND it isn't a symlink in the sfs
3. the style changes are not explained/justified - why are they needed?
4. moving the "if" is a different additional update
5. the error message needs to provide information on why the sfs load is failing - if it can't be translatable it will need to be just English
6. I prefer to leave it to the user to decide whether they want to load an sfs having provided them with information - don't like things that hard fail with no get-out clause

woof-code/rootfs-skeleton/usr/sbin/sfs_load
peabee: Again - another different update - also /usr/lib ??

woof-code/rootfs-skeleton/usr/sbin/sfs_load
dimkr: > Replying to all 7 comments made over a very short period of time.........

Next time, if you don't want my review or intend to dismiss my comments, please don't tag me and don't ask me for help.

> libx32 as a symlink with no libx32 directory in sfs works fine (tested)

The logical bug happens when the main SFS has a directory at /libx32 (or any other directory in the list), while the devx doesn't have anything there.

> the test is "-a" - both directory must exist AND it isn't a symlink in the sfs

Enjoy this counter-example:

```
/tmp$ mkdir x
/tmp$ rm y
rm: cannot remove 'y': No such file or directory
/tmp$ [ -d x -a ! -L y ] && echo bug
bug
```

> the style changes are not explained/justified - why are they needed?

I wouldn't say they're absolutely _needed_, it's just cleaner, more efficient in some cases, and more consistent with non-Puppy shell scripts.

> the error message needs to provide information on why the sfs load is failing - if it can't be translatable it will need to be just English

How about changing this:

```
$(gettext "The directory $FLD_INVALID in this sfs will overwrite the system symlink. This may cause problems. Do you want to continue?")
```

to this?

```
$(gettext "The following directories this sfs will overwrite system symlinks. This may cause problems. Do you want to continue?")

-- list of directories goes here --
```

> I prefer to leave it to the user to decide whether they want to load an sfs having provided them with information - don't like things that hard fail with no get-out clause

There are cases where a SFS is **guaranteed** to break the system. There's no reason to allow this. People who want to ensure the destruction of their OS can just `rm -rf /; kill -9 1` or something.

woof-code/rootfs-skeleton/usr/sbin/sfs_load
dimkr: I'll quote you in #3320:

> /lib64 and /usr/lib64 are symlinks

If you're trying to fix the problem for /lib64, why not solve it for /usr/lib64 as well?

woof-code/rootfs-skeleton/usr/sbin/sfs_load
dimkr: I misread this condition, I understood as if it's `if [ -d $fld2 -a ! -L /mnt/$MNTFLD/$fld2 ]; then`, but my point still stands, in a slightly different way. sfs_load needs to check if both $fld2 and /mnt/$MNTFLD/$fld2 are symlinks, both are directories, or at least one of them is missing (but deal correctly with the corner case where `! -e` is true because it's a broken symlink, and `! -L` when it's missing).

peabee:
Can the pull request as amended proceed please?
Other perceived problems can be tackled in further pull requests.

dimkr:
> Can the pull request as amended proceed please?
> Other perceived problems can be tackled in further pull requests.

Not a fan of #3027 and this PR, but I have my own stable woof-CE fork to protect my builds against breaking changes. No objection from my side.

woof-code/rootfs-skeleton/usr/sbin/sfs_load
01micko: > I prefer to leave it to the user to decide whether they want to load an sfs having provided them with information - don't like things that hard fail with no get-out clause

Lowest common denominator? Imagine if traffic rules were legislated by your rationale.

01micko: `<hypothetical>`
APP: If you proceed all your funds run the risk of being transferred to a Mafia cartel. Do you wish to proceed?
USER: ** hits wrong button **
APP: **I'm fatter now**
`</hypothetical>`

Who is responsible?

peabee:
Warning reverted as it seemed to cause the most concern.

mavrothal:
> user freedom reverted as it seemed to cause the most concern.

:-o

I personally find the comment that "freedom causes concern" VERY offensive and I would appreciate if you retract it.

I also find strange that you believe that the people you want to "free", are stupid!  
Because the person that wants to for some reason go ahead and bomb the system loading a malformed  SFS or just test if the SFS is malformed, can certainly comment-out the exit line. 

Unless if by "free" you mean help people make detrimental mistakes.
Yeah, this "help" causes concern.

peabee:
Words removed.

mavrothal:
> Words removed.

English is not my best language but I did not ask to remove the words (which is useless in git anyway) I asked to retract the statement.
As far as I understand retraction is done by another statement explaining why the previous one does not accurately represents the facts or ideas of the original.
In other words I ask if you BELIEVE that freedom is a source of concern for me (and the participants in this discussion) or not.
If you do, removing the words makes no difference. If you don't do state so.

Issue #3306, opened 2022-08-03
Experimental builds should be distinct from established builds


peabee:
At the moment all builds run on Github are "lumped together" into a single Action class: builds.

Experimental builds should be in a different Action class to incremental builds of established Pup versions - e.g.

experimental builds
established builds

PR #3327, opened 2022-08-15
Add a watermark to development builds with Wayland

dimkr:
This will make life easier for victims of #3306.

watermark2
watermark3

### DIFF
--- a/woof-code/rootfs-petbuilds/activate-linux/etc/xdg/autostart/activate-linux.desktop
+++ b/woof-code/rootfs-petbuilds/activate-linux/etc/xdg/autostart/activate-linux.desktop
@@ -1,0 +1,7 @@
+[Desktop Entry]
+Version=1.0
+Name=activate-linux
+Comment=Activate Linux watermark for Wayland
+Exec=sh -c '. /etc/DISTRO_SPECS; GDK_BACKEND=wayland exec activate-linux "$DISTRO_NAME $DISTRO_VERSION" "This is an unstable development build"'
+Terminal=false
+Type=Application
--- a/woof-code/rootfs-petbuilds/activate-linux/pet.specs
+++ b/woof-code/rootfs-petbuilds/activate-linux/pet.specs
@@ -1,0 +1,1 @@
+activate-linux-1|activate-linux|1||Desktop|316||activate-linux-1.pet||Activate Linux watermark for Wayland|puppy|||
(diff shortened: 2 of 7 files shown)


peabee (commented):
Don't know what to say........... watermark is not very helpful - what does "not genuine" mean / infer??

Such a watermark is only useful after you've expended lots of effort downloading and installing an iso - you need to know before you expend such effort!

Don't see the point.

dimkr:
I don't see the point in #3306, then. All builds built using the latest woof-CE, from a branch called "testing", are available through a CI service called GitHub Actions (and not through a page labeled "Releases"), which requires signup+login to a website that describes itself as a "complete developer platform". A user has to ignore so much information and go through multiple steps to do the mistake of downloading a development build. How can the hypothetical fix for #3306 protect such stubborn users, who ignore so many warning signs?

woof-code/rootfs-petbuilds/activate-linux/etc/xdg/autostart/activate-linux.desktop
```diff
@@ -0,0 +1,7 @@
+[Desktop Entry]
+Version=1.0
+Name=activate-linux
+Comment=A full-featured GTK based trayicon UI for ConnMan
+Exec=sh -c '. /etc/DISTRO_SPECS; GDK_BACKEND=wayland exec activate-linux "$DISTRO_NAME $DISTRO_VERSION" "This copy of $DISTRO_NAME is not genuine"'
```
mavrothal: what about "is a development build for testing" instead of "is not genuine"
Given that "genuine" puppy builds are not defined anywhere I can find, "not genuine" does not mean anything really and certainly does not mean that this is a development build for testing

dimkr: It's "genuine" as in https://www.google.com/search?q=windows+is+not+genuine+watermark&tbm=isch

01micko:
Do you need a binary?

```sh
#!/bin/sh

. /etc/DISTRO_SPECS

cat > /tmp/${0##*\/}.css <<_S
window {
	background-color: rgba(0, 0, 0, 0);
	color: #313131;
	font-weight: bold;
	font-size: 20px;
}
_S

export P='<window edge="bottomright" width-request="400" height-request="200" layer="bottom">
 <vbox>
  <hbox>
   <text><label>'"$DISTRO_NAME $DISTRO_VERSION"'</label></text>
  </hbox>
  <hbox>
   <text><label>Activate this Pup</label></text>
  </hbox>
 </vbox>
</window>'

gtkdialog -p P --styles=/tmp/${0##*\/}.css
```

20220826_13h14m45s_grim

mavrothal:
wow
"your copy of <OS> is not genuine"
"Activate this <OS> "
I know MS owns github now but I did not expect such an immediate influence
😆

01micko:
I think the whole thing is a little (lot?) out of hand.

A binary file is just a waste of space and resources when 25 lines of shell (not including possible startup and config files) utilising puppy standard inclusions can easily distinguish a 'stable' vs 'experimental' release, if that's what we really need anyway.

Unfortunately, this entire thing is a waste of valuable human resources too. I'm knocking it on the head.

dimkr:
> wow "your copy of is not genuine" "Activate this " I know MS owns ggithub now but I did not expect such an immediate influence laughing

Don't give MS too much credit, the concept of irony predates MS.

PR #3337, opened 2022-08-21
Speed up tofi by avoiding expensive font lookup

dimkr:
See https://github.com/philj56/tofi/blob/aeefb7cd446cbc32f6bd68f3f776a032979b3b92/doc/config#L12.

@01micko Is it safe to assume any Puppy has this font at this path?


woof-code/packages-templates/tofi_FIXUPHACK
```diff
@@ -6,7 +6,7 @@ height = 26
 horizontal = true
 font-size = 10
 prompt-text = "\$ "
-font = monospace
+font = /usr/share/fonts/TTF/DejaVuSansMono.ttf
```
01micko, line 9: Yes

01micko (approved):

Issue #3334, opened 2022-08-20
GTK+ 3 doesn't respect Xft.dpi under Xwayland


dimkr:
```
~$ cat ~/.Xresources 
!*font: -misc-fixed-medium-r-semicondensed--13-120-75-75-c-60-*-*
!*font: -misc-fixed-medium-r-normal--13-120-75-75-c-70-*-*
*font: -misc-fixed-medium-r-normal--14-130-75-75-c-70-*-*
!*font: -misc-fixed-medium-r-normal--15-140-75-75-c-90-*-*

Xft.dpi: 192
```

![xftdpi](https://user-images.githubusercontent.com/1471149/185744012-00dec6ed-3c97-4740-aefa-843d84de82c4.png)

dimkr:
Not fixing this one, see #3417.

Issue #3332, opened 2022-08-20
yambar uses the same dimensions when the scaling factor is 2


dimkr:

dimkr:
wp-fractional-scale-v1 support across the board will fix this.

PR #3331, opened 2022-08-20
Make tofi visually consistent with dwl

dimkr:
This PR adds a 1px bright green border around tofi. dwl doesn't draw the green border around this window and it steals the focus until closed.

bookworm has tofi 0.4.0 and will probably receive philj56/tofi@ea64fc43a6a321fcc7417f35d5ce50a2c7ee9d7e shortly after the next tofi release. This commit makes tofi visually consistent with dwl: without it, tofi doesn't scale the 1px border to 2px when the scale factor is 2.

PR #3339, opened 2022-08-23
Run Xwayland with the default DPI until #3334 is fixed

dimkr:
Some applications, including JWM, respect the `-dpi` value, while others rely on `Xft.dpi`. Until #3334 is fixed and GTK+ 3/4 applications use a larger font size, let's use the same font size everwhere.

(The "pure Wayland" setup is unaffected, and scaling mostly works under Wayland, with the notable exceptions of #3331 and #3332.)


dimkr:
Now that #3358 is merged and Xwayland is stretched, this one is needed.

PR #3358, opened 2022-08-31
Increase the dwl scale factor only in >= 1080p displays

dimkr:
Tiny virtual machine windows satisfy the condition of >= 96 DPI, not because they are dense, but because they are tiny.


dimkr:
Merging this one, I think it's tested well enough.

Issue #3187, opened 2022-07-05
PUPMODE 13 with aufs corrupts browser settings and history


dimkr:
When using aufs, snapmergepuppy copies binary files (like the sqlite database used to store history) to pup_ro1 while the browser writes to them, then deletes them from pup_rw. This means pup_ro1 contains revision x of the file, while the browser is writing revision x+1 to a deleted file. Next time snapmergepuppy runs, the file is missing, so revision x+1 doesn't get copied to pup_ro1 and it's lost.

PR #3359, opened 2022-08-31
snapmergepuppy: Avoid active opened files

rizalmart:
Don't move the file to save file/folder if the file was currently opened by a process

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/snapmergepuppy
+++ b/woof-code/rootfs-skeleton/usr/sbin/snapmergepuppy
@@ -204,12 +204,18 @@ do
 		if [ -L "$N" ] && [ -d "$BASE/$N" ] ; then
 			rm -rf "$BASE/$N"	# SFR: in case if folder has been replaced with a symlink (cp won't overwrite a dir with a symlink)
 		fi
-		cp -af "$N" "$BASE/$N" 2>>/tmp/snapmergepuppy-error
-		rm "$N"
+		#Move if the file was not opened by a process
+		if [ "$(lsof | awk '{ print $3 }' | grep "/" | grep -vE "\/dev|\/proc|\/sys|\/tmp|\/run|\/mnt|\/media" | sort | uniq | grep "$N")" == "" ]; then
+		 cp -af "$N" "$BASE/$N" 2>>/tmp/snapmergepuppy-error
+		 rm "$N"
+		fi
 	else
 		if [ ! -e "$BASE/$N" ] || [ ${NCTIME%%.*} -gt `stat -c %Z "$BASE/$N"` ] ; then
+		    #Move if the file was not opened by a process
+		    if [ "$(lsof | awk '{ print $3 }' | grep "/" | grep -vE "\/dev|\/proc|\/sys|\/tmp|\/run|\/mnt|\/media" | sort | uniq | grep "$N")" == "" ]; then
 			cp -af "$N" "$BASE/$N" 2>>/tmp/snapmergepuppy-error
 			rm "$N"
+		    fi
 		fi
 	fi
 


woof-code/rootfs-skeleton/usr/sbin/snapmergepuppy
```diff
@@ -204,12 +204,18 @@ do
 		if [ -L "$N" ] && [ -d "$BASE/$N" ] ; then
 			rm -rf "$BASE/$N"	# SFR: in case if folder has been replaced with a symlink (cp won't overwrite a dir with a symlink)
 		fi
-		cp -af "$N" "$BASE/$N" 2>>/tmp/snapmergepuppy-error
-		rm "$N"
+		#Move if the file was not opened by a process
+		if [ "$(lsof | awk '{ print $3 }' | grep "/" | grep -vE "\/dev|\/proc|\/sys|\/tmp|\/run|\/mnt|\/media" | sort | uniq | grep "$N")" == "" ]; then
```
dimkr, line 208: `lsof` is a pretty heavy operation, and you do it per file. Maybe it's a better idea to run `lsof` once and redirect to a file, before this loop, and grep the file.

dimkr, line 208: Isn't `lsof | grep -m1 "^$N\$"` considerably faster than `lsof | awk '{ print $3 }' | grep "/" | grep -vE "\/dev|\/proc|\/sys|\/tmp|\/run|\/mnt|\/media" | sort | uniq | grep "$N"`?

dimkr, line 208: I think you mean `grep "^$N\$"` - i.e. `/a/b/c` is not opened if the output contains `/a/b/cd`; only exactly `/a/b/c` means `/a/b/c` is open

```
~$ (echo abc; echo abcd) | grep abc
abc
abcd
~$ (echo abc; echo abcd) | grep '^abc$'
abc
```

woof-code/rootfs-skeleton/usr/sbin/snapmergepuppy
```diff
@@ -204,12 +204,18 @@ do
...
+		 cp -af "$N" "$BASE/$N" 2>>/tmp/snapmergepuppy-error
+		 rm "$N"
```
dimkr, line 210: Maybe just copy but without deleting, if the file is open? That's safer against power failure. For example, if the browser is constantly open, the user will lose all browsing history since bootup when the battery drops to zero. On the other hand, there's a race condition between the process that writes to the file and `cp`, so the result of the copying is not guaranteed to be a valid file.

dimkr:
@rizalmart I don't see a way to fix #3187 without a redesign of snapmergepuppy. It needs some radical solution, like a FUSE-based file system that blocks writing to files under /home/spot and /root until snapmergepuppy stops running.

woof-code/rootfs-skeleton/usr/sbin/snapmergepuppy
rizalmart: > Maybe just copy but without deleting, if the file is open?

Not a good idea for me.

rizalmart:
> @rizalmart I don't see a way to fix #3187 without a redesign of snapmergepuppy. It needs some radical solution, like a FUSE-based file system that blocks writing to files under /home/spot and /root until snapmergepuppy stops running.

fuse-based fs can make puppy performance slower it seems

dimkr:
Definitely, FUSE is super slow. But you need a file system (doesn't matter if it's in userspace or not) that blocks access for writing while snapmergepuppy is running, if you want to ensure consistency.

PR #3361, opened 2022-09-01
snapmergepuppy: Avoid opened files 2.0

rizalmart:
Avoid opened files based from @dimkr's recommendation from previous PR #3359

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/snapmergepuppy
+++ b/woof-code/rootfs-skeleton/usr/sbin/snapmergepuppy
@@ -215,14 +215,14 @@ do
 		#Check if the file was opened or shutting down
 		if [ "$(echo $OPF | grep -m1 "^$N\$")" == "" ] || [ "$SHUTDOWN" == "yes" ]; then
 		  cp -af "$N" "$BASE/$N" 2>>/tmp/snapmergepuppy-error
-		  rm "$N"
+		  [ $? -eq 0 ] && rm "$N"
 		fi
 	
 	else
 		if [ ! -e "$BASE/$N" ] || [ ${NCTIME%%.*} -gt `stat -c %Z "$BASE/$N"` ] ; then
 		     if [ "$(echo $OPF | grep -m1 "^$N\$")" == "" ] || [ "$SHUTDOWN" == "yes" ]; then
 			cp -af "$N" "$BASE/$N" 2>>/tmp/snapmergepuppy-error
-			rm "$N"
+			[ $? -eq 0 ] && rm "$N"
 		     fi
 		fi
 	fi
--- a/woof-code/rootfs-skeleton/usr/sbin/snapmergepuppy
+++ b/woof-code/rootfs-skeleton/usr/sbin/snapmergepuppy
@@ -165,6 +165,10 @@ FREEBASE=`df -B 1 | grep -w "$BASEMTP"| head -n 1 | tr -s ' ' | cut -f 4 -d ' '`
 rm -f /tmp/snapmergepuppy-nospace #110222 shinobar
 rm -f /tmp/snapmergepuppy-error   #140102 SFR
 
+#Get list of last opened files
+OPF="$(lsof | awk '{ print $3 }' | grep "/" | grep -vE "\/dev|\/proc|\/sys|\/tmp|\/run|\/mnt|\/media|\/memfd:" | sort | uniq)"
+
+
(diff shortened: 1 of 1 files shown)


dimkr:
@01micko I see nothing wrong with this second revision of this PR, but I'm not sure I understand the motivation for this, as it doesn't solve #3187 but slows down snapmergepuppy a bit.

01micko:
I'll wait for @rizalmart 's response as to whether to merge or not.

rizalmart:
> @01micko I see nothing wrong with this second revision of this PR, but I'm not sure I understand the motivation for this, as it doesn't solve #3187 but slows down snapmergepuppy a bit.

Just a thought here. Since aufs and overlayfs creates modified version of files at writable layer/upper folder. It much better to avoid to those actively opened files when moving the modified files to save file/folder to avoid I/O errors between process and target file. Almost the same concern as dimkr's #3187.

dimkr:
I think I proposed this in V1, but why not just copy the file to pup_ro1 without deleting it from pup_rw, if a file is writing to it? This way, if the file is corrupt because it was copied in some intermediate state (before all changes are flushed), it will get "fixed" in a future snapmergepuppy.

#3187 happens not because files are copied while a process is writing to them, but because they're deleted while a process is writing to them. snapmergepuppy copies file revision x to pup_ro1 while the process is writing revision x+1, then snapmergepuppy deletes the file, the data written to the file is discarded (so revision x+1 is lost), and next time snapmergepuppy runs it doesn't have a file to copy. Copying a file while a process is writing to it can produce a corrupt file, but that corruption is only temporary if the file is kept in pup_rw and gets copied later.

rizalmart:
>but why not just copy the file to pup_ro1 without deleting it from pup_rw, if a file is writing to it? This way, if the file is corrupt because it was copied in some intermediate state (before all changes are flushed), it will get "fixed" in a future snapmergepuppy.

The problem with this approach was it will fill the ramdisk space quite easily.

rizalmart:
> #3187 happens not because files are copied while a process is writing to them, but because they're deleted while a process is writing to them. snapmergepuppy copies file revision x to pup_ro1 while the process is writing revision x+1, then snapmergepuppy deletes the file, the data written to the file is discarded (so revision x+1 is lost), and next time snapmergepuppy runs it doesn't have a file to copy. Copying a file while a process is writing to it can produce a corrupt file, but that corruption is only temporary if the file is kept in pup_rw and gets copied later.

Thats why the file that was currently opened by process should be avoided in the first place for copying files to save file

dimkr:
> The problem with this approach was it will fill the ramdisk space quite easily.

I'm not sure, because the ramdisk already holds all files opened for writing. And you don't save memory by deleting files if they're currently open: they're can't be re-opened, but they remain there (although they're inaccessible) until all currently running processes close them.

woof-code/rootfs-skeleton/usr/sbin/snapmergepuppy
```diff
@@ -215,14 +215,14 @@ do
 		#Check if the file was opened or shutting down
 		if [ "$(echo $OPF | grep -m1 "^$N\$")" == "" ] || [ "$SHUTDOWN" == "yes" ]; then
 		  cp -af "$N" "$BASE/$N" 2>>/tmp/snapmergepuppy-error
-		  rm "$N"
+		  [ $? -eq 0 ] && rm "$N"
```
dimkr, line 218: Copying will succeed even if a process is writing to the file (it will copy a potentially corrupt file), so this check doesn't do anything

rizalmart:
> I'm not sure, because the ramdisk already holds all files opened for writing.

Probably not, aufs and overlayfs only resides modified/altered/deleted file on writable layer/upperdir.

> And you don't save memory by deleting files if they're currently open: they're can't be re-opened, but they remain there (although they're inaccessible) until all currently running processes close them. 

What about inert added/modified files which was not opened by any process? They consume ramdisk space

dimkr:
> What about inert added/modified files which was not opened by any process? They consume ramdisk space

They won't appear in the output of `lsof`, so you'll just copy and delete them.

EDIT: in case my proposal wasn't clear, I propose to copy and delete files that don't appear in `lsof` output, and copy the files that do appear there without deleting them.

rizalmart:
> They won't appear in the output of `lsof`, so you'll just copy and delete them.
> 
> EDIT: in case my proposal wasn't clear, I propose to copy and delete files that don't appear in `lsof` output, and copy the files that do appear there without deleting them.

Oh I get it, however the concern here was since the file was currently opened. There is a chance that the file copied on save file/folder might be corrupted since the file was currently active with the process and accidentally copy that file in the middle of I/O operation of a process with the file

dimkr:
> There is a chance that the file copied on save file/folder might be corrupted

Exactly, and if you don't delete the file from pup_rw, the next snapmergepuppy run will copy it again. If this time it's not corrupted, then this second run will fix the corruption in pup_ro1.

PR #3363, opened 2022-09-01
Automatically fall back to overlay if aufs is unsupported

dimkr:
This doesn't break anything for anyone, and doesn't change the current default of aufs.

If (and only if) aufs is unavailable, the init script falls back to overlay and allows Puppy to boot instead of failing; currently, our kernels are built with aufs, and this PR makes it possible to swap the kernel with one without aufs.

This means UNIONFS is redundant: the init script can just default to aufs if supported. As before, the user can force of use of overlay when using a kernel that supports aufs, by adding `punionfs=overlay` to the kernel command-line.

See https://github.com/puppylinux-woof-CE/woof-CE/discussions/3362.


01micko (commented):
Ok

PR #3371, opened 2022-09-04
Replace {f,e}grep with grep -{F,E}

dimkr:
https://www.phoronix.com/news/GNU-Grep-3.8-Stop-egrep-fgrep

### DIFF
--- a/kernel-kit/firmware_picker.sh
+++ b/kernel-kit/firmware_picker.sh
@@ -158,7 +158,7 @@ do
 	for F in $SRC_FW_DIR/$fw_top_dir/*/hw*;do
 		fw_subdir=${F#$SRC_FW_DIR/}
 		[ -e $FIRMWARE_RESULT_DIR/$fw_subdir ] && continue
-		fgrep -qlm1 ${fw_subdir#$fw_top_dir/} /tmp/modstrings || continue
+		grep -Fqlm1 ${fw_subdir#$fw_top_dir/} /tmp/modstrings || continue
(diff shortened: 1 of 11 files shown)


01micko (approved):

01micko:
Funny, I started coding in shell around 2007 and never used egrep/fgrep simply because I knew about -E/-F

dimkr:
> Funny, I started coding in shell around 2007 and never used egrep/fgrep simply because I knew about -E/-F

I started around that time too, and I'm using fgrep for 2-3 years I think 😊

peabee:
FYI - Slackware "comment":
a/grep-3.8-i586-2.txz:  Rebuilt.
  Folks, I rarely veto upstream, but I'm not going to entertain this nonsense.
  The egrep and fgrep commands were part of Unix since the 70s, continue to be
  included with the BSDs, and frankly, aren't hurting anything. GNU grep
  declared them deprecated in 2007 and when they were changed into shell
  scripts around 8 years ago I figured that's where it would end. I can see no
  logical justification to have these scripts start making noise and then to
  eventually pull the rug out from under any code that might be using them, so
  I've placed non-noisy versions of them into the package sources and will be
  installing those during the build. Given that the -F and -E options are part
  of the POSIX standard, these scripts will continue to work fine. That said,
  we will be continuing to change our own code over to the recommended syntax
  to avoid the minimal overhead incurred compared to using grep directly.

01micko:
@peabee I agree with PV's sentiment but woof supports more than Slackware.

PR #3373, opened 2022-09-04
Change focal64 to kernel 5.10

peabee:
Maybe temporary until 5.4 aufs kernel-kit can be fixed

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -56,7 +56,7 @@ jobs:
           - arch: x86_64
             compat-distro: ubuntu
             compat-distro-version: focal64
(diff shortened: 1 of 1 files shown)


dimkr (approved):
No objection, although focal64 is unmaintained and very different from 9.5. I don't think anyone is using the focal64 CI build to test stuff.

Dropping all kernel builds not officially supported by aufs and migrating all builds (not just focal64) to newer kernel branches (say, >=5.10.x) sounds (to me) like a better direction to go. I'd drop the 4.19.x and 5.4.x builds, then drop the aufs variant of every kernel branch that loses official aufs support. aufs is a big maintenance burden with the frequent breakage and the new support policy.

Issue #3333, opened 2022-08-20
Xwayland contents are blurry when the scaling factor is 2


dimkr:
The fullscreen Xwayland window is scaled, and not the icons and text of applications running inside it.

![scaled](https://user-images.githubusercontent.com/1471149/185743751-5611c748-26bd-430d-861b-ee1646401f86.png)

Native Wayland applications (like l3afpad and activate-linux, see #3327) are scaled properly:

![native](https://user-images.githubusercontent.com/1471149/185743834-84b82413-ac8e-495b-8941-1eee364e5291.png)

The cursor is scaled properly, when it's inside a native window:

![xcursor](https://user-images.githubusercontent.com/1471149/185743878-3f3261c5-ea14-45be-b352-2a6f11e65ee6.png)
![wcursor](https://user-images.githubusercontent.com/1471149/185743877-57b37f74-09dd-42e3-8195-45a007ca9ae5.png)

PR #3378, opened 2022-09-05
Set the Xwayland font DPI via ~/.Xdefaults (fixes #3334)

### LINKED ISSUE #3334: GTK+ 3 doesn't respect Xft.dpi under Xwayland

dimkr:
To use ~/.Xresources, we need xrdb, but it comes from x11-xserver-utils, which depends on parts of GCC.

~/.Xdefaults is a smaller solution.

To ensure consistency with applications running as root, I copied root's ~/.Xdefaults to /home/spot (for example, connman-ui  runs as spot).

Now font scaling works correctly under Xwayland (example with 144 DPI):

EDIT: I also disabled automatic choice of the scaling factor, when dwl runs in kiosk mode. Scaling should be handled the (awful) X way, by increasing font DPI, and the compositor should not stretch the Xwayland window.


01micko (approved):

dimkr:
Oh no, this fix doesn't work 100%!

1. VLC ignores .Xdefaults 😢 
2. VLC increases the text size after `xrdb -merge ~/.Xresources`, but it's very weird. If I run `xrdb` via /etc/xdg/autostart, it works just fine, but only for applications started after JWM. If I run `xrdb` right after Xwayland starts, it doesn't have any effect. If I run it before JWM, it affects JWM and applications started after JWM, but not ROX-Filer and the tray icons, which are started before JWM.

dimkr:
Extracted the fix for #3333 into #3384, will fix #3334 later.

PR #3391, opened 2022-09-10
Add support for UEFI boot using Ubuntu's signed shim

dimkr:
This makes #3321 work for Ubuntu as well and replaces #2912 with a more generic solution, that  works for older Ubuntu releases too.

Frugalpup's old binaries are used only if Ubuntu's newer, signed binaries don't.


01micko:
We need a `grub2` petbuild really. @jamesbond3142 sent me the puppy.cer signed certificate an age ago and all the tools *should* be available to sign the shims. I'll allow this PR as an interim solution.

AlienBob uses a similar method in his `liveslak` project using either ubuntu, debian or fedora signed shims.

01micko (approved):

gyrog:
I think it's the grub ".efi" file that needs to  be signed with 'puppy.cer'. The shim is already signed.

PR #3395, opened 2022-09-12
Remove unwanted \

peabee:
causes new grep WARNINGS

### DIFF
--- a/woof-code/0setup
+++ b/woof-code/0setup
@@ -68,7 +68,7 @@ if [ "$1" = "" ] ; then
 fi
 #======================================================================
 
-if wget --help | grep -q '\-\-show\-progress' ; then
+if wget --help | grep -q '--show-progress' ; then
 	WGET='wget --no-check-certificate -q --show-progress'
 else
 	WGET='wget --no-check-certificate -q'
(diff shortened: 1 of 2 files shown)


01micko (dismissed):

woof-code/0setup
```diff
@@ -68,7 +68,7 @@ if [ "$1" = "" ] ; then
 fi
 #======================================================================
 
-if wget --help | grep -q '\-\-show\-progress' ; then
+if wget --help | grep -q '--show-progress' ; then
```
dimkr, line 71:
```
~$ wget --help | grep -q '--show-progress'
grep: unrecognized option '--show-progress'
Usage: grep [OPTION]... PATTERNS [FILE]...
Try 'grep --help' for more information.
```

dimkr: You'll need to be very careful.

dimkr: You need `--` to stop option parsing, i.e. `wget --help | grep -q -- '--show-progress'`

01micko:
There's probably many more. It _was_ standard practice to escape `-` especially if the 'grep' arg was surrounded in `".."`

woof-code/0setup
01micko: Oh, so there _is_ a case for escape.

dimkr: Yes, but you don't need that escaping if you make grep understand this is not an option the right way 😄

01micko: Also leaving off the '--' works

`wget --help | grep -q 'show-progress'`

dimkr:
> There's probably many more. It _was_ standard practice to escape `-` especially if the 'grep' arg was surrounded in `".."`

See #3397

peabee:
Replaced by #3397

PR #3404, opened 2022-09-15
Use the foot default of TERM=foot

dimkr:
foot supports many useful features, and tools like chafa support them if TERM=foot. Compatibility with rxvt-unicode is nice, but comes at the cost of incompatibility or degraded functionality in some modern console applications.


woof-code/support/rootfs-hacks.sh
```diff
@@ -228,6 +228,12 @@ done
 
 # need to enforce pterminfo: xterm -- see /etc/profile
 if [ -d ${SR}/usr/share/terminfox ] ; then
+	for FOOT in foot foot-direct; do
```
dimkr: TODO: drop this if #3408 is merged

01micko (approved):

PR #3419, opened 2022-09-24
Optionally restrict file system access for applications running as spot

dimkr:
Landlock is a new kernel API that allows a process to restrict its file system access: for example, it allows a process to disallow all access to /root, allow reading from /tmp and allow writing to /tmp/runtime-spot, then exec() an application.

This is a nice security feature that can be used as a second layer of defense for applications running as spot. Some Puppy packages mess up directory permissions and ownership, and nobody notices that because everything works just fine for applications running as root, and extra permissions for spot don't break anything. In some Puppy releases, spot can read files under /root, maybe even write to them! Landlock is a great, fine-grained security solution that can be used to prevent spot from doing things it shouldn't do, even if file system permissions say otherwise.

Before this PR, pkexec is a root SUID executable that shows a yes/no prompt (using a restricted yad process that runs as spot), then runs a process as root if the user agrees. However, Landlock requires the PR_SET_NO_NEW_PRIVS bit to be set, which means a process that uses Landlock cannot run SUID executables: a process running as spot which uses Landlock will run pkexec as spot despite the SUID bit.

Therefore, this PR moves most functionality from pkexec to a daemon that runs as root, pkexecd. pkexec talks to this daemon over a Unix socket, doesn't do anything on its own and doesn't need to be a SUID executable.

To enable:
1. Switch to kernel >= 5.15.x
2. Build the kernel with
```
CONFIG_SECURITY_LANDLOCK=y
CONFIG_LSM="landlock"
```
(CONFIG_LSM should contain `landlock`, doesn't have to be exactly `landlock` - currently, Puppy doesn't use any LSM AFAIK)
3. Enable the `spot-pkexec` petbuild

If the kernel is too old, Landlock is missing or Landlock is disabled, `spot-sandbox` continues with PR_SET_NO_NEW_PRIVS and without Landlock.

If `spot-sandbox` is not available, run-as-spot doesn't use it.

When `spot-sandbox` is available, processes running as spot can:
- Read from /, minus /root and /home
- Read from and write to /home/spot, /dev, /proc and /tmp

If /tmp loses its 1777 permissions, spot can't write to it. In other words, Landlock is a second layer of security on top of file permissions: if file permissions disallow a certain action, these rules don't apply.

After `chmod 777 /root`, with and without this Landlock-based sandboxing:

image
image

Chrome installation through gdebi+pkexec+sandboxed Firefox running as spot:

image
image
image
image

EDIT: had to add write permissions for /tmp, because Chrome creates temporary files directly under /tmp.
EDIT 2: now pkexecd handles multiple requests in parallel

## TODO

- [x] Test Firefox with WebGL
- [x] Test Chrome with WebGL
- [x] Test .deb package installation through gdebi, when gdebi is started by downloading a package through the browser

### DIFF
--- a/woof-code/rootfs-petbuilds/spot-pkexec/etc/xdg/autostart/pkexecd.desktop
+++ b/woof-code/rootfs-petbuilds/spot-pkexec/etc/xdg/autostart/pkexecd.desktop
@@ -1,0 +1,7 @@
+[Desktop Entry]
+Version=1.0
+Name=pkexecd
+Comment=pkexec daemon
+Exec=pkexecd
+Terminal=false
+Type=Application
--- a/woof-code/rootfs-petbuilds/spot-pkexec/petbuild
+++ b/woof-code/rootfs-petbuilds/spot-pkexec/petbuild
@@ -4,4 +4,10 @@ download() {
 
 build() {
     $CC $CFLAGS -D_GNU_SOURCE pkexec.c $LDFLAGS -o /usr/bin/pkexec
+    $CC $CFLAGS -D_GNU_SOURCE pkexecd.c $LDFLAGS -o /usr/bin/pkexecd
+    if [ -e /usr/include/linux/landlock.h ]; then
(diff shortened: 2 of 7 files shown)


01micko (approved):
Merge when you're ready

PR #3421, opened 2022-09-25
Ensure fusermount is SUID root (fixes #3420)

### LINKED ISSUE #3420: AppImages fail because FUSE doesn't work as spot
spot is not allowed to mount FUSE file systems, so AppImages can't run as spot. Some applications (like VLC) don't want to run as root, so such AppImages don't work on Puppy.

~Maybe fusermount needs to be SUID root?~ EDIT: yes

dimkr:


dimkr:
Seems to work nicely and at least three users already enjoy this fix after I cherry-picked it into Vanilla Dpup 9.2.x.

PR #3427, opened 2022-09-27
Update release.yml

lakshayrohila:
Made changes so that DISTRO_VERSION doesn't include 64 even if 64 is typed.

For ex.-
Input:- jammy64
DISTRO_VERSION:- jammy

In previous versions of the file:-
Input:- jammy64
DISTRO_VERSION:- jammy64 (cannot be found in ubuntu's repos)


dimkr:
DISTRO_VERSION != DISTRO_COMPAT_VERSION. DISTRO_VERSION is the Puppy version (for example, you can put 9.5 or 10.0 there) and DISTRO_COMPAT_VERSION is the compatible distro (= Ubuntu) version, and that should be something like `jammy` or `focal`.

PR #3440, opened 2022-09-30
Make rc.sysinit faster if modules are builtin and X uses libinput

dimkr:


woof-code/rootfs-skeleton/etc/rc.d/rc.sysinit
```diff
@@ -455,39 +455,44 @@ if [ $PUPMODE -eq 2 ] ; then #full hd installation.
 fi
 
 [ ! -d /proc/acpi ] && modprobe apm #load apm.ko if no acpi.
-zcat /proc/config.gz | grep -m 1 -q '^CONFIG_INPUT_LEDS=m' && modprobe input-leds #for >= k4.2x
+modprobe input-leds 2>/dev/null #for >= k4.2x
```
dimkr, line 458: Loading `input-leds` if `CONFIG_INPUT_LEDS=m` is the same as trying to load `input-leds` and letting it fail if `CONFIG_INPUT_LEDS=y` or `CONFIG_INPUT_LEDS=n`, but much slower, because modprobe uses the cache generated by depmod (which includes a list of builtin modules and .ko paths) and module lookup is very fast - faster than decompressing /proc/config.gz

woof-code/rootfs-skeleton/etc/rc.d/rc.sysinit
```diff
@@ -455,39 +455,44 @@ if [ $PUPMODE -eq 2 ] ; then #full hd installation.
...
 TOUCHMOD=`lsmod|grep touch`
 ISITANXO=`cat /sys/class/dmi/id/product_name` #mavrothal: Check if it is an OLPC XO
 ISITAMAC=`grep -i -m 1 '^Apple' /sys/class/dmi/id/board_vendor`
-[ -f /lib64/libc.so.6 ] && lsuffix=64 || lsuffix= #slackware64
-if [ ! -f /usr/lib${lsuffix}/xorg/modules/input/kbd_drv.so -a ! -f /usr/lib${lsuffix}/xorg/modules/input/mouse_drv.so ] ; then
-  echo "XORG_UDEV=1: Xorg kbd_drv.so and mouse_drv.so are missing"
-  XORG_UDEV=1
-fi
-
-TOUCHSCREEN="$(udevadm info --export-db | grep -m 1 ID_INPUT_TOUCHSCREEN=1)"
```
dimkr, line 472: This check is slow, over 150ms on a relatively fast laptop. Why do it if `XORG_UDEV=1` already?

woof-code/rootfs-skeleton/etc/rc.d/rc.sysinit
```diff
@@ -517,7 +522,9 @@ echo "WAIT MODULES LOADED"
 #previous module loading may not have completed...
 echo -n "Waiting for modules to complete loading..." >/dev/console
 PCMCIABUILTIN='no'
-if [ "`modinfo yenta_socket 2>/dev/null`" = "" ];then   #Marvmod shut this up.
+if modinfo yenta_socket | grep -m 1 -q '(builtin)';then
```
dimkr, line 525: `yenta_socket` can be builtin although the output of `modinfo yenta_socket` is not empty, so the `lspci` can be skipped

woof-code/rootfs-skeleton/etc/rc.d/rc.sysinit
```diff
@@ -536,11 +543,11 @@ fi
 
 #note, if initrd then usb-storage will have already loaded...
 USBBUILTIN='no'
-if [ "`modinfo ehci_hcd 2>/dev/null`" = "" ];then #110712 shut this up.
+if modinfo ehci_hcd | grep -m 1 -q '(builtin)' ;then #kmod new version
```
dimkr, line 546: `ehci_hcd` is more likely to be builtin in non-ancient kernels, so there's no reason to run `modinfo` twice with most Puppy kernels

01micko (approved):

PR #3443, opened 2022-09-30
is_mbp is always empty

dimkr:
~~@rizalmart Maybe that's not the only bug introduced?~~

Whoa, it's been this way since 66b42e6 @wdlkmpx 

If this entire thing is broken since at least 2018 (= never worked) and nobody complains, maybe it's a better idea to just remove it? Fixing something broken for years can introduce new issues with today's kernels, and maybe it's not needed anymore. EDIT: see #3447.

### DIFF
--- a/woof-code/rootfs-skeleton/etc/rc.d/rc.sysinit
+++ b/woof-code/rootfs-skeleton/etc/rc.d/rc.sysinit
@@ -633,7 +633,7 @@ fi
 ##############platform adjustments################## see fd64's #rc.platform
 ###  - do it before we configure network and services
 # detect if it's MacBookPro (mavrothal)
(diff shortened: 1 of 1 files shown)


mavrothal:
Suppose to be empty unless puppy is running on a MacBookPro retina. 
Did you try it on a macbookpro?

dimkr:
> Suppose to be empty unless puppy is running on a MacBookPro retina. Did you try it on a macbookpro?

But it's *always* empty, even on one of these Macs.

-q means:

```
       -q, --quiet, --silent
              Quiet; do not write anything to standard output.
```

So `if [ "$is_mbp" ] ; then` is always false, on any hardware.

See https://github.com/puppylinux-woof-CE/woof-CE/pull/3447#issuecomment-1264615553.

PR #3447, opened 2022-10-01
Revert "rc.sysinit: add code from fatdog's rc.platform"

dimkr:
Alternative to #3443.

66b42e6 doesn't work because of that `-q`, and fixing 4+ years later can be more dangerous than removing this broken code that doesn't do anything.


mavrothal:
That's an old kernel problem. Without this hack any linux distro will fry a MacBookPro from 2012-15.
Of course I'm not sure if anyone else is running Puppy on a MBP so if is such a problem it can be removed. Though I'm not sure what do you mean does not work because of ```-q ```. This is still a valid grep option I think. Or can use ```-s``` instead.

dimkr:
> Though I'm not sure what do you mean does not work because of `-q`


Without -q, what the author intended:

```
~$ echo 1macbookpro2 | grep -i -m 1 macbookpro
1macbookpro2
```

With -q, not what the author intended, no output and `is_mbp` is empty even if `dmesg` contains `macbookpro`:

```
~$ echo 1macbookpro2 | grep -q -i -m 1 macbookpro
```

In other words, `if [ "$is_mbp" ] ; then` is always false.

mavrothal:
No, you are right. 
I have my own set of patches/mods for my MacBook and ```-q``` is not there 😊 
I guess indeed no one else runs puppies on MacBooks...
Remove ```-q``` or better replace with ```-s```

dimkr:
The question is, what's safer - drop this unused feature which is always disabled (this PR), or fix it (#3443).

If this feature finally starts working, it needs real-world testing on one of these Macs. And even after testing on Macs, it does introduce some risk on other machines, because looking for a single word in `dmesg` is prone to false-positives. For example, I can create a partition labeled `macbookpro` on my Thinkpad, and that will activate this Mac-specific voodoo, which can be dangerous.

But if nobody complains since <= 2018, why bother fixing this feature (which never worked) or adding extra protection against false positives, if we can remove it?

PR #3452, opened 2022-10-03
Remove the old and broken continuous TRIM implementation in favor of /etc/init.d/trim

dimkr:
According to various sources, periodic TRIM is better for SSD health, and that's why we have /etc/init.d/trim. I have an old SSD that becomes unreliable and forces me to reboot every once in a while if a partition is mounted with `discard`.

(And it looks like `pfix=trim` doesn't do anything, because PTRIM is ignored.)

EDIT: and this old support for continuous TRIM is broken since 2019 anyway!

### DIFF
--- a/initrd-progs/0initrd/sbin/mountpartition
+++ b/initrd-progs/0initrd/sbin/mountpartition
@@ -29,14 +29,7 @@ mkdir -p ${MNT_DIR}
 
 #-----------------------------------------------------
 
-. /etc/rc.d/functions_x
 MNT_O='noatime'
-MNT_DSK=$(fx_get_drvname $MNT_DEV)
-if [ "$MNT_FS" = "ext4" -o "$MNT_FS" = "f2fs" ] ; then
-  if [ "$(cat /sys/block/${MNT_DSK}/queue/rotational 2>/dev/null)" = "0" ] ; then
-    MNT_O="-o discard,noatime" # enable SSD TRIM
-  fi
-fi
 
 case $MNT_FS in
   ntfs)
--- a/initrd-progs/0initrd/README.txt
+++ b/initrd-progs/0initrd/README.txt
@@ -226,11 +226,10 @@ pimod=<, separated list of kernel module names>
    But sometimes init needs to request input from the user via the keyboard.
    Specifying kernel modules in this parameter will cause init to load them before any possible keyboard interaction.
 
-pfix=<ram, nox, trim, nocopy, fsck, fsckp, rdsh, <number>>
+pfix=<ram, nox, nocopy, fsck, fsckp, rdsh, <number>>
    The pfix parameter is a ',' separated list of 1 or more of the above sub-parameters.
    ram:      run in ram only (do not load ${DISTRO_FILE_PREFIX}save).
    nox:      do not start X.
-   trim:     add "discard" to mount options if SSD.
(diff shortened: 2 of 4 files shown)


initrd-progs/0initrd/sbin/mountpartition
```diff
@@ -29,15 +29,6 @@ mkdir -p ${MNT_DIR}
 
 #-----------------------------------------------------
 
-. /etc/rc.d/functions_x
-MNT_O='noatime'
-MNT_DSK=$(fx_get_drvname $MNT_DEV)
-if [ "$MNT_FS" = "ext4" -o "$MNT_FS" = "f2fs" ] ; then
-  if [ "$(cat /sys/block/${MNT_DSK}/queue/rotational 2>/dev/null)" = "0" ] ; then
-    MNT_O="-o discard,noatime" # enable SSD TRIM
```
dimkr, line 37: How does this even work? If line 37 works, then line 33 is broken. If I understand correctly, later in this file we pass `-o $MNT_O` to `mount`, and `-o -o discard,noatime` doesn't work (syntax error).

dimkr: Amazing, now I understand why that extra `-o` is not a problem: this piece of code is broken since its birth at ea50d51e87 ([this line](https://github.com/puppylinux-woof-CE/woof-CE/commit/ea50d51e87278372e19f913a8674466558b25613#diff-10f6bca9f7b05a23796c659a01bca009790e1bab43c0bd24ab8506840ac6c0c3L134) was not copied correctly when the code was moved to a separate file, probably an attempt to make the code prettier). The condition in line 36 is always false, so the value of MNT_O is never changed to the bad value that won't work.

initrd-progs/0initrd/init has this line, which passes `/dev/whatever` and not `whatever` to mountpartition:

```
 /sbin/mountpartition /dev/${1} $ONE_MP $ONE_FS #-t $ONE_FS /dev/$ONE_PART $ONE_MP
```

Then, mountpartition puts the first argument in MNT_DEV and passes it to fx_get_drvname:

```
MNT_DEV="$1"  #ex: /dev/sda1

...

xdev=${MNT_DEV##*/} #basename

...

MNT_DEV=/dev/$xdev

...

MNT_DSK=$(fx_get_drvname $MNT_DEV)

...

  if [ "$(cat /sys/block/${MNT_DSK}/queue/rotational 2>/dev/null)" = "0" ] ; then

...
```

This `if` is always false, because fx_get_drvname expects `sdb1`, not `/dev/sdb1`. It returns `/de` and `/sys/block/de/queue/rotational` doesn't exist. Therefore, TRIM is never enabled.

```
~$ MNT_DEV=/dev/sdb1
+ MNT_DEV=/dev/sdb1
~$ xdev=${MNT_DEV##*/} #basename
+ xdev=sdb1
~$ MNT_DEV=/dev/$xdev
+ MNT_DEV=/dev/sdb1
~$ MNT_DSK=$(fx_get_drvname $MNT_DEV)
++ fx_get_drvname /dev/sdb1
++ case $1 in
++ echo /de
+ MNT_DSK=/de
~$ cat /sys/block/${MNT_DSK}/queue/rotational
+ cat /sys/block//de/queue/rotational
cat: /sys/block//de/queue/rotational: No such file or directory
```

01micko (approved):

PR #3457, opened 2022-10-05
Move mount points even when using overlay

dimkr:


gyrog:
There may be a need to re-establish the P_MP variable, if it's the PDRV that was moved.
Sorry, can't remember for sure.

dimkr:
> There may be a need to re-establish the P_MP variable, if it's the PDRV that was moved.

Maybe. If needed, this can be fixed separately. I want to see if `-o move` works first - I don't remember why I changed to `-o bind` with overlay. So far, everything looks OK, and I really wonder why I did that.

gyrog:
What about lines 1393, 1394?
Should lines be replaced, or should $OPT be defined?

PR #3459, opened 2022-10-05
Remove overlay whiteouts during pupsave upgrade

dimkr:
Easier to read with whitespace changes hidden, https://github.com/puppylinux-woof-CE/woof-CE/pull/3459/files?diff=unified&w=1


gyrog (approved):
It looks like it should delete overlayfs "whiteout" files not in '/root/.config/autostart',
if that's what it's suppposed to do, then I approve it.

PR #3461, opened 2022-10-05
Skip hwclock --show on x86 to speed up rc.country

dimkr:
On my Thinkpad, the `hwclock --show` check takes somewhere between 400ms and 1s.

Instead of running `hwclock --show` on any computer that runs Puppy just in case it's a Pi, the logic should be:
1. Set the time using /var/local/shutdown_date_saved or DISTRO_SPECS, as fallback in case set_hwclock_type fails or it's a Pi
2. If it's ARM, run `hwclock --show`
3. Skip set_hwclock_type if this is ARM and `hwclock --show`'s output is empty

### DIFF
--- a/woof-code/rootfs-skeleton/etc/rc.d/rc.country
+++ b/woof-code/rootfs-skeleton/etc/rc.d/rc.country
@@ -83,15 +83,21 @@ fi
 #======================================
 
 #120629 raspberry pi does not have a hw clock, set date to reasonable value...
-HWDATE="`hwclock --show 2>/dev/null`" #ex: "Fri 29 Jun 2012 07:45:28 AM WST  -0.725833 seconds"
-if [ "$HWDATE" = "" ];then
-	if [ -f /var/local/shutdown_date_saved ];then #see /etc/rc.d/rc.shutdown
-		date -s "`cat /var/local/shutdown_date_saved`"
-	else
-		#either of these formats can set the date: "29 JUN 2012 10:00:00" "2012-06-28 16:20:08"
-		date -s "`stat -c %z /etc/DISTRO_SPECS | cut -f 1 -d '.' | cut -f 1,2 -d ' '`" #creation date of build. ex: "2012-06-28 16:20:08"
-	fi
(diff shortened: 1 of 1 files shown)


gyrog:
I always run Puppies using a savefolder, pupmode=12 on an Intel i3 desktop machine,
so I have a working hardware clock AND my savefolders always contain a '/var/local/shutdown_date_saved' file.

Which part of your proposed code is going to stop my clock being set on every boot to the date and time I last shutdown this particular Puppy?

dimkr:
> Which part of your proposed code is going to stop my clock being set on every boot to the date and time I last shutdown this particular Puppy?

No, no difference. Same behavior, only faster. NTP will set the clock anyway.

gyrog:
You did not answer my question.

In the old code:

HWDATE="`hwclock --show 2>/dev/null`" #ex: "Fri 29 Jun 2012 07:45:28 AM WST  -0.725833 seconds"
if [ "$HWDATE" = "" ];then

these 2 lines meant that my clock was not set.

In  the new code:

if [ -f /var/local/shutdown_date_saved ];then #see /etc/rc.d/rc.shutdown
	date -s "`cat /var/local/shutdown_date_saved`"

these 2 lines ensure that the coick on my machine will be set.
I would call that a change in behaviour.

dimkr:
I'm not sure I understand. set_hwclock_type sets the clock: if you run `date` before and after it you'll see the system clock has changed.

Before this change, the code uses shutdown_date_saved **if the output of `hwclock --show` is empty**, otherwise runs set_hwclock_type.

After the change, the code uses shutdown_date_saved unconditionally, then runs set_hwclock_type: if the latter failed because there's no hardware clock, the behavior is same as before: the clock is set to the value of shutdown_date_saved. If set_hwclock_type succeeds, same behavior as before.

The only difference is that `hwclock --show` is skipped, because it's slow.

@gyrog Makes sense?

01micko:
I'm happy to merge but I'll await @gyrog 's response

dimkr:
@gyrog ping

PR #3471, opened 2022-10-08
chmod +x /usr/local/pmusic/* (fixes #3462)

### LINKED ISSUE #3462: pmusic - no executable permissions
Executable permissions not set on files in:
rootfs-packages/pmusic/usr/local/pmusic

dimkr:


zigbert:
@dimkr 
box_splash is missing in this commit, and has not changed permission. I could not find the status at github, but I assume that is not executable since it was uploaded together with the other files.

Else, everything looks good.
THANK YOU!

01micko:
Yes, `box_splash` is still 644

dimkr:
I did this using ```grep -l '^#\!/bin' `find woof-code/rootfs-packages/pmusic -type f | grep -v pinstall.sh` | while read x; do git update-index --chmod=+x $x; done```. Looks like box_splash and maybe some other files use `#! /bin/bash` (with a space) and not `#!/bin/bash`, that's why I missed it.

Fixed.

PR #3472, opened 2022-10-08
Drop --indicator-idle-visible until flickering is fixed upstream

dimkr:


01micko:
This behaviour and other output issues have disappeared for me as wlroots edges closer to 16.0

PR #3484, opened 2022-10-10
Put a sandboxed application running as spot in a user namespace

dimkr:
Another layer of security hardening can't hurt. After this change, sandboxed applications running as spot don't see users other than spot: files owned by root look just like files owned by `nobody`, and UIDs other than spot's UID are invalid.


01micko (approved):
While my `C` is still pretty fundamental, this looks OK to me.

PR #3487, opened 2022-10-11
Add support for /etc/environment

dimkr:
Users are asking about input methods and some need environment variables like GTK_IM_MODULE. Some online tutorials instruct users to define them in /etc/environment and this won’t work in Puppy because Puppy’s login doesn’t use PAM.

This is a good opportunity to move environment variables like GDK_BACKEND outside of the compositor and simplify things.


01micko (approved):
I can probably make this work for `labwc` by symlinking `/etc/environment` to `~/.config/labwc/environment` if labwc fails to pick up any of these that are needed for labwc.

dimkr:
It will "just work" on labwc and even X, because these environment variables are exported by the login shell  👍🏿

PR #3499, opened 2022-10-13
Remove glib-memory-alloc

dimkr:
`G_SLICE=always-malloc` is a debugging feature and shouldn't be enabled normally.

@rizalmart Why is it needed? What applications depend on this debugging feature?


01micko (approved):

01micko (approved):

dimkr:
Still haven't found an application that needs this. Merging.

PR #3500, opened 2022-10-15
If not using ROX-Filer, let the file manager define mimeapps.list

dimkr:
The way xdg-utils works in Puppy is a hack:
1. There are hidden menu entries for `default%s` applications
2. These hidden pseudo-applications are configured as the default for various MIME types and URL schemes in woof-code/rootfs-skeleton/usr/share/applications/mimeapps.list
3. ROX-Filer has its own list of handlers (in /etc/xdg/rox.sourceforge.net/MIME-types): it uses `default%s` as well
4. Applications other than ROX-Filer respect mimeapps.list and not the ROX-Filer scripts, so they use the same handlers
5. A Puppy-specific application configures default MIME type handlers **by modifying `default%s`, rather than modifying ~/.config/mimeapps.list** like GNOME, Xfce, etc' do
6. xdg-open is replaced with a stub that runs ROX-Filer, because mimeapps.list defines only a subset of the handlers in ROX-Filer's internal list of handlers

This pile of hacks replaces the source of truth, which should be mimeapps.list, with the `default%s` applications, and only because ROX-Filer doesn't respect mimeapps.list.

This PR removes this hack **if ROX-Filer is missing**:
1. Puppy's special xdg-open is still there, and still uses the file manager to open files
2. The file manager shows a dialog that lets the user choose the MIME type handler and adds it to mimeapps.list, if no handler is configured, otherwise it invokes the handler
3. The `default%s` applications invoke configured MIME type handlers by querying them via `xdg-mime`

This logic should work equally well with pcmanfm and spacefm, making Puppy more compatible with DEs and applications that obey the XDG standards, while retaining backward compatibility with old Puppy tools that  invoke `default%s`.

### DIFF
--- a/woof-code/3builddistro
+++ b/woof-code/3builddistro
@@ -934,20 +934,55 @@ exec $value \"\$@\"" > ${DRV_TGT}/usr/local/bin/$field
 	chmod 755 ${DRV_TGT}/usr/local/bin/$field
 done
 
-echo "Checking for missing default apps ..."
-for default in rootfs-complete/usr/local/bin/default* ; do
-    app=${default##*/}
-    app=${app/default/}
-    dexec=`grep '^exec' $default | cut -f 2 -d " "`
-    chroot rootfs-complete sh -c "command -v $dexec >/dev/null 2>&1" && continue
+if [ -e rootfs-complete/usr/local/apps/ROX-Filer/AppRun ]; then
+	echo "Checking for missing default apps ..."
+	for default in rootfs-complete/usr/local/bin/default* ; do
+		app=${default##*/}
+		app=${app/default/}
+		dexec=`grep '^exec' $default | cut -f 2 -d " "`
+		chroot rootfs-complete sh -c "command -v $dexec >/dev/null 2>&1" && continue
 
-    echo -n "$dexec "
-    cat << EOF > $default
+		echo -n "$dexec "
+		cat << EOF > $default
 #!/bin/sh
 exec missingdefaultapp $app
 EOF
-done
-echo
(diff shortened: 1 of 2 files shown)


peabee:
What does "using" and "missing" mean?

What happens if both ROXfiler and say PcManFM are installed but PcManFM is set as the default file manager?

dimkr:
> What does "using" and "missing" mean?

If you're referring to the PR title, "If not using ROX-Filer, let the file manager define mimeapps.list", "using" means "ROX-Filer is present" and "missing" means "ROX-Filer is not present".

> What happens if both ROXfiler and say PcManFM are installed but PcManFM is set as the default file manager?

You get the original behavior, with the ROX-Filer specific hacks. As I described, these hacks are dropped, to make other file managers happy, only if ROX-Filer is missing.

01micko (commented):
Looks OK to me

dimkr:
@01micko It **will** break some stuff, but not for builds with ROX-Filer (which won't be affected at all). For example, I see that firefox.desktop specifies even image/png. The gpicview petbuild will need `xdg-mime default gpicview.desktop image/png` in pinstall.sh. There will be more cases like this, but not something that can't be fixed easily later.

dimkr:
Works fine after several days of testing. Merging.

@peabee Let me know if it causes any issues, it should change absolutely nothing in a Puppy that has `/usr/local/apps/ROX-Filer/AppRun`. I've checked multiple Puppy releases and all of them have ROX-Filer at the same directory, so I don't think it's prone to false-positives.

PR #3503, opened 2022-10-17
Reduce verbosity when using kernels from before 5998641 (fixes #3502)

### LINKED ISSUE #3502: udevd warnings displayed during boot sequence (new behaviour)
Behaviour has changed between:
BUILD_FROM_WOOF='testing;31102a7f1;2022-10-05 13:18:17 +0000'
and
BUILD_FROM_WOOF='testing;42f1deedd;2022-10-13 19:34:27 +0300'

These udevd warnings have always been present in dmesg but are now displayed during the boot sequence causing concern to users (but boot completes OK):
```
[   18.171838] udevd[1091]: specified group 'input' unknown
[   18.172385] udevd[1091]: NAME="%k" is ignored, because it breaks kernel supplied names, please remove it from /etc/udev/rules.d/55-cups.rules:5
[   18.177522] udevd[1091]: IMPORT{builtin}: 'net_setup_link' unknown /lib/udev/rules.d/80-net-setup-link.rules:9
[   18.177797] udevd[1091]: unknown key 'DEVTYPE' in /etc/udev/rules.d/88-puppy-autodetect.rules:6
[   18.177852] udevd[1091]: invalid rule '/etc/udev/rules.d/88-puppy-autodetect.rules:6'
[   18.177925] udevd[1091]: unknown key 'DEVTYPE' in /etc/udev/rules.d/88-puppy-autodetect.rules:10
[   18.177997] udevd[1091]: invalid rule '/etc/udev/rules.d/88-puppy-autodetect.rules:10'
[   18.179441] udevd[1092]: starting eudev-3.2.9
[   18.487193] udevd[1092]: specified group 'input' unknown
[   18.487703] udevd[1092]: NAME="%k" is ignored, because it breaks kernel supplied names, please remove it from /etc/udev/rules.d/55-cups.rules:5
[   18.491951] udevd[1092]: IMPORT{builtin}: 'net_setup_link' unknown /lib/udev/rules.d/80-net-setup-link.rules:9

```

dimkr:


peabee (commented):
Not really sure why the line was there if it can be deleted ........... but if it fixes #3502 then go for it.

dimkr:
> Not really sure why the line was there if it can be deleted ........... but if it fixes #3502 then go for it.

The kernel you use defaults to loglevel 7. When LOGLEVEL=7, which means debugging output is requested, that extra verbosity is enabled.

After this change, the init script ignores the default the kernel was built with and always forces loglevel 3 if no `loglevel=` parameter was specified, compensating for this kernel misconfiguration.

peabee:
Errrrrr /kernel-kit/build.sh lines 641 et al already do this.......

dimkr:
Try `modprobe configs; zcat /proc/config.gz | grep CONFIG_CONSOLE_LOGLEVEL_DEFAULT`

dimkr:
4.14.295, 5.10.149, 5.15.74 and 6.1-rc1 have this line:

```
#define CONSOLE_LOGLEVEL_DEFAULT CONFIG_CONSOLE_LOGLEVEL_DEFAULT
```

4.9.330 has this line:

```
#define CONSOLE_LOGLEVEL_DEFAULT 7 /* anything MORE serious than KERN_DEBUG */
```

build.sh looks for a line that got removed at some point before 4.14.0, so the printk-related part of the code block you're referring to is broken since at least 2017.

EDIT: changed my mind, the pattern uses `.*` and not `[0-9]`, so the `sed` for this line should work. But maybe it doesn't work because it doesn't deal with `MESSAGE_LOGLEVEL_DEFAULT` and other `CONFIG_*` references.

Issue #3489, opened 2022-10-11
extrasfs fail to load at startup in VDpup 10.0.43


mavrothal:
As the title says.
The sfs is written in BOOTCONFIG but never loads, not is shows in the console during the boot process.
Loading manually ot viewing is OK but still not present after reboot.

dimkr:
Please attach BOOTCONFIG and /initrd/tmp/bootinit.log

mavrothal:
Attached
[BOOTCONFIG.zip](https://github.com/puppylinux-woof-CE/woof-CE/files/9761877/BOOTCONFIG.zip)
[bootinit.log.zip](https://github.com/puppylinux-woof-CE/woof-CE/files/9761878/bootinit.log.zip)

dimkr:
Can you reproduce this against 10.0.44 with `loglevel=7` and attach bootinit.log?

mavrothal:
I will some time tonight  but till then I was looking at it briefly and looks like that ```setup_onepupdrv``` returns ```3``` for empty file at ```/mnt/pdrv/<filename>.sfs``` where ```find_onepupdrv``` mounts it.

dimkr:
I'm not sure I understand - why would this file be empty, and how can it be mounted if it's empty?

mavrothal:
The sfs
 is NOT empty (you can actually click on it and see its content)
May be something in the ```setup_onepupdrv``` logic
Do extra SFSs mount OK for you (or others)?

mavrothal:
Wasn't at some point that the extrasfs mounting moved (for aufs builds) to the rc.sysinit or I remember something else?

dimkr:
> Do extra SFSs mount OK for you (or others)?

It used to work, but I can't rule out the possibility of a regression.

> Wasn't at some point that the extrasfs mounting moved (for aufs builds) to the rc.sysinit or I remember something else?

If SFS loading is moved to rc.sysinit, it won't with with overlay unless it's implemented using some inefficient trick (like #3398).

mavrothal:
I think the problem is in ```setup_onepupdrv``` that while the sfs is mounted in ```/mnt/dev_save/<filename>.sfs```, is looking for it in ```/mnt/pdrv/<filename>.sfs``` and ```/mnt/pdrv``` does not exist!
This is because ```find_drv_file``` sets ```"$3"``` as ```p```  but pdrv was never made.

dimkr:
Looks fine here:

[...]

PR #3513, opened 2022-10-19
Move overlay support to init.overlay (preparation for #3489)

dimkr:
Current init minus aufs support is copied to init.overlay and #2963 in init is reverted, so aufs and overlay now live in separate files.

Now overlay-related work can continue in a separate file, without disturbing anyone. The next step is narrowing down the scope of init.overlay to PUPMODE 5, 12 and 13, then coding style cleanup, then bug fixes and efficiency improvements on top of a (cleaner and shorter) code base.


gyrog (approved):
Well done, now we have 2 1,400 line monoliths.

I still think this is a backward step.
But go for it.

PR #3528, opened 2022-10-22
Merge VoidPup64 into testing

dimkr:
Based on 22.02+10c.

The build passes locally and kind-of-works, still trying to see why it fails on GitHub but I'm almost there.

(The prebuilt adrv is omitted, because there's nowhere to download it from)


01micko (commented):
LGTM

peabee (commented):
Thank you for this......
A lot to review........ and difficult to review without an iso to load and test
Some brief comments:
"Kind-of-works"??
VoidPup comes in both 32-bit and 64-bit versions - this only builds VoidPup64 - I'm unsure how I will keep my local versions (which I prefer to keep) and this "cloud" version synchronised
Why is version 10.0? Whats wrong with 22.10?
I'm not a fan of version specific inline code buried in 3builddistro....
The fdrv is handled the same as the adrv in my local builds

dimkr:
> "Kind-of-works"??

Yes, it's without the prebuilt adrv, and I don't know if the result is exactly the same as the builds you produce locally.

> VoidPup comes in both 32-bit and 64-bit versions - this only builds VoidPup64

The 32-bit build can be added later, most changes are shared between the two.

> I'm unsure how I will keep my local versions (which I prefer to keep) and this "cloud" version synchronised

You can use woof-CE as-is (clone the repo) or create a fork of testing plus extra changes. When you want to, you can upstream these changes via pull requests.

> Why is version 10.0? Whats wrong with 22.10?

I even changed the name to something generic, so users who see this don't confuse development builds with VoidPup releases.

> I'm not a fan of version specific inline code buried in 3builddistro....

Me neither, but that's the cleanest solution I see in woof-CE as it is today, and sometimes it's easier to understand where the generalizations fail when the exceptions are explicit.

Plus, I don't like the idea of a file for distro x version y that's placed in woof-code/support and not woof-distro/x/y: the devx hacks thing needs to be somewhere if it's a separate file.

peabee:
I have uploaded the adrv and fdrv to:
https://smokey01.com/peebee/voidpup/

/voidpup/voidpupxx/sfs-extra/

peabee:
I would be happy for devx-hacks to be in woof-distro/x86_64/void
or
woof-code/support/void

the path can be specified in _00build.conf  EXTRA_PROCESSING= ?

peabee:
Did a test build locally on LxPupSc64 (lots of grep \ warnings!)....

Quite a lot of missing dependencies: [checkdeps_system.txt-false.gz](https://github.com/puppylinux-woof-CE/woof-CE/files/9846782/checkdeps_system.txt-false.gz) but otherwise looking very promising.

screenie

dimkr:
> Quite a lot of missing dependencies: [checkdeps_system.txt-false.gz](https://github.com/puppylinux-woof-CE/woof-CE/files/9846782/checkdeps_system.txt-false.gz) but otherwise looking very promising.

Are you sure you don't see the same output in VoidPup releases? DISTRO_PKGS_SPECS should be identical.

peabee:
No - for VoidPup64 its completely clear
As an example djvulibre is a pet package which is not making it into rootfs-complete - it has things in /usr/lib64 so maybe the transfer to /usr/lib isn't working quite correctly?? status/findpkgs_FINAL_PKGS-void-current says it has been included but its got deleted somehow.

```
VoidPup64-22.02
Start: Fri Oct  7 00:52:24 +08 2022
* Checking system dirs: /bin /usr/libexec /sbin /usr/bin /usr/sbin /usr/games /usr/local/bin


==================================================

* Checking system dirs: /lib /usr/lib /usr/local/lib


```

peabee:
p.s. support/void/makepkgdb.sh probably needs last 2 lines removed rather than the conditional exit

peabee:
Sorry - just noticed
sandbox3/rootfs-complete/var/db/xbps/pkgdb-0.38.plist

is not how it should be - tiny content - should be c. 1MB

dimkr:
> should be c. 1MB

makepkgdb.sh spits errors

peabee:
I think the lib deletion problem is because the inline code is much earlier in 3builddistro than the running of the combined pinstall.sh (which is where the copying currently should happen from the pinstall from 00_vpup64_fix)

I think the inline code needs changing....

peabee:
> makepkgdb.sh spits errors

I think? void-hacks-4woofce-211022/64/pre-0setup/0setup has not been implemented so index.plist does not exist

The code at lines 78-90 of 3builddistro seems to be somewhat different to that in void-hacks-4woofce-211022/64/devx-hack/void64/devx-hack.sh.....? It is also much earlier than where /sandbox3/devx is populated.

peabee:
Made progress.............

put devx-hack.sh   void0setup  into support/void

edited cvt.sh   makepkgdb.sh  in support/void

commented out lines 300-304 in 0setup - replaced with support/void/void0setup

commented out lines 78-90 in 3builddistro (no replacement)

changed _00build.conf:
EXTRA_PROCESSING="support/void/devx-hack.sh"
ADRV_SFS_URL=http://smokey01.com/peebee/voidpup/voidpup64/sfs-extra/adrv_vpup64_22.02.sfs
FDRV_SFS_URL=http://smokey01.com/peebee/voidpup/voidpup64/sfs-extra/fdrv_vpup64_22.02.sfs

rebuilt - now clear checkdeps - no unsatisfied dependencies
/var/db/xbps/pkgdb-0.38.plist is now correct
adrv is included so Light works

however downloaded fdrv does not replace the fdrv in the kernel-kit - can this be done?

dimkr:
My schedule has changed a bit. I'll continue working on this next week or in the week after 🤒

01micko:
> however downloaded fdrv does not replace the fdrv in the kernel-kit - can this be done?

Probably not. Where are you getting this firmware? Old firmware doesn't really cut it. If you want _some_ firmware make a list and cull it in `kernel-kit` with a specific script. We can probably accommodate that .

peabee:
> Where are you getting this firmware? 

A cut down version of that supplied by gyrog: https://www.forum.puppylinux.com/viewtopic.php?t=2302
that I've used in all my builds for a long time.

There is a facility in _00build.conf to download an fdrv but it seems to be ignored.

01micko:
Those files are huge! ~340MB. What logic do you use to cut it down? What don't you trust about the `kernel-kit` firmware build? It's all basically from the same source. And, firmware is often kernel specific.

peabee:
Historic precedent from previous pups
Not a trust issue - just too big - my fdrv is c. 20MB

> firmware is often kernel specific

Really?? Which firmware are you thinking about? I have seen no indications ever that firmware is kernel specific! Firmware is code which is loaded into devices - why would it have anything to do with the kernel? I have used the same firmware with many different kernels with no problems (for me or reported).

peabee:
Here are the files I had to change:
[void-hacks-4woofce-261022.sfs-false.gz](https://github.com/puppylinux-woof-CE/woof-CE/files/9867628/void-hacks-4woofce-261022.sfs-false.gz)

dimkr:
Firmware is kernel-specific. Some drivers expect a specific version of firmware, so an older fdrv won't work. Newer computers need more firmware (generally) and some machines need more firmware than others, so maybe you don't see the impact of "missing" firmware.

Sticking to a specific fdrv, especially if it's not tailor-made for the kernel you're using and not reproducible, is a super bad idea IMO.

> What logic do you use to cut it down?

If we don't have the script (or the algorithm) to produce this fdrv, it's not reproducible. The firmware can't be updated to fix issues or add support for hardware that needs newer firmware, and the selection of firmware can't be changed to include an extra firmware blob or two.

Why trust a random, "closed source" SFS from the forums more than the algorithm in kernel-kit?

dimkr:
Deprioritized on my end. If anyone wants to take it from here - feel free to do so.

PR #3540, opened 2022-10-23
Make releases workflow set prefix

lakshayrohila:
Latest commits to the github workflow system disabled a user to set prefix by running directly the releases workflow in github.

Resolved the issue by making `releases` workflow file call `build` workflow file while setting `prefix` variable as well.


dimkr:
Thanks @lakshayrohila!

PR #3547, opened 2022-10-24
Fill LASTUNIONRECORD and PREVUNIONRECORD on first boot

dimkr:
Otherwise, applications can't tell if devx, docx and nlsx are auto-loaded on first boot (#3506).

### DIFF
--- a/initrd-progs/0initrd/init
+++ b/initrd-progs/0initrd/init
@@ -1398,8 +1398,8 @@ if [ "$PUPSAVE" ];then
  echo "$xBOOTCONFIG" > /pup_new/etc/rc.d/BOOTCONFIG
  echo "LASTUNIONRECORD='$(echo -n $NEWUNIONRECORD)'" >> /pup_new/etc/rc.d/BOOTCONFIG
 else
- echo "" > /pup_new/etc/rc.d/BOOTCONFIG
- echo "LASTUNIONRECORD=''" >> /pup_new/etc/rc.d/BOOTCONFIG
+ echo "PREVUNIONRECORD='$(echo -n $NEWUNIONRECORD)'" > /pup_new/etc/rc.d/BOOTCONFIG
+ echo "LASTUNIONRECORD='$(echo -n $NEWUNIONRECORD)'" >> /pup_new/etc/rc.d/BOOTCONFIG
 fi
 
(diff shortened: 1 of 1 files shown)


mavrothal:
I do not see the harm, other that would be confusing (or even suspicious if paranoid)  to someone that might look the PREVUNIONRECORD on first boot, but what is the reasoning for that?
Which problem are you trying to solve with this hack?

dimkr:
> Which problem are you trying to solve with this hack?

#3506 
#3545

mavrothal:
Isn't #3506 going to autoload these SFSs if found regardless of ```BOOTCONFIG``` ?
Regarding #3545 why don't you just add  a condition, say in rc.sysinit, to update the caches if in PUPMODE 5 and nlsx is leaded, instead of running the update script?

dimkr:
> Isn't #3506 going to autoload these SFSs if found regardless of `BOOTCONFIG` ?

Yes, these SFSs are auto-loaded if found (like adrv and friends).

> why don't you just add a condition, say in rc.sysinit, to update the caches if in PUPMODE 5 and nlsx is leaded, instead of running the update script?

rc.update updates the cache and rc.sysinit needs a way to tell if nlsx is loaded.

mavrothal:
> rc.update updates the cache and rc.sysinit needs a way to tell if nlsx is loaded.

what about ```[ -d /initrd/pup_nls -a "$PUPMODE" = "5" ]``` ?

dimkr:
> what about `[ -d /initrd/pup_nls -a "$PUPMODE" = "5" ]` ?

1. It's hacky
2. What about docx, nlsx and any other SFS?

mavrothal:
> It's hacky

while having ```PREVIOUSONRECORD``` on first boot is not?

You could even count the mounted loops and run the update if more

dimkr:
> while having `PREVIOUSONRECORD` on first boot is not?

PREVIOUSONRECORD is meaningful only if PUPMODE != 5, and this PR fixes this inconsistency.

dimkr:
Is there any reason why keep `PREVIOUSONRECORD=""` on first boot? I don't see anything that uses this variable in PUPMODE 5 and may break by fixing this inconsistency.

PR #3573, opened 2022-11-03
Revive pre-ptheme tools and use them if PTHEME is unspecified

dimkr:
This is something I've wanted to do for a very long time. I don't like ptheme, it tries to do way too many things in one tool, it's very buggy and it looks like users just want the old-style layout of one tray at the bottom, or one of the alternatives that conflict with ptheme.

If I understand correctly, choose_themes is in this half-broken, zombie state since the introduction of ptheme (2014?), so it's been a very long time since the days it was possible to build a Puppy without ptheme using woof-CE, without ugly hacks like replacing the .jwm directory. I had to make some adjustments like setting the GTK+ 3 theme, adding Xwayland support and changing paths to what Puppy uses nowadays.

ptheme_gtk is in very bad shape and even has some syntax errors, but we need some GTK+ theme switcher even when ptheme is disabled. I made ptheme_gtk work without refactoring and cleanup, ~~then moved it to rootfs-skeleton~~.

(This brings dpup even closer to what I consider to be the golden age of Puppy, the 4.1.x-4.3.x era)

desktop

### DIFF
--- a/woof-code/rootfs-packages/ptheme/usr/sbin/ptheme_gtk
+++ b/woof-code/rootfs-packages/ptheme/usr/sbin/ptheme_gtk
@@ -6,7 +6,7 @@ export TEXTDOMAIN=ptheme
 export OUTPUT_CHARSET=UTF-8
 
 #check is X is runnning
-[ "`grep -F 'xwin' <<< $(ps -eo pid,command)`" ] && export X_IS_RUNNING=true || export X_IS_RUNNING=false
+pidof -s X Xorg dwl-kiosk labwc > /dev/null && export X_IS_RUNNING=true || export X_IS_RUNNING=false
 
 export WORKDIR=$HOME/.ptheme/tmp/ptheme
 [ ! -d $WORKDIR ] && mkdir -p $WORKDIR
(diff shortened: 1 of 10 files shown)


woof-code/rootfs-packages/ptheme/usr/sbin/ptheme_gtk
```diff
@@ -151,8 +151,8 @@ _EOF
 	elif [ "$ACTIVE_ICON_THEME" ]; then
 		sed -i '/gtk-icon-theme-name/d' $HOME/.gtkrc-2.0
 		echo "gtk-icon-theme-name = \"$ACTIVE_ICON_THEME\"" >> $HOME/.gtkrc-2.0
-		set_gtk3_theme "ACTIVE_$ICON_THEME"
```
01micko, line 154: So bugfix here

woof-code/support/choose_themes
```diff
@@ -7,11 +7,14 @@ if ! [ "$BUILDSYS" ] ; then
...
-#	THEME_GTK2=''      # /usr/share/themes/*/gtk-2.0 (* = theme)
-#	THEME_JWM=''       # /root/.jwm/themes/*-jwmrc (* = theme)
-#	THEME_DESK_ICONS=' # /usr/local/lib/X11/themes
-#	THEME_OPENBOX=''   # /usr/share/themes/*/openbox-3
+#	THEME_WALLPAPER=''   # /usr/share/backgrounds
+#	THEME_GTK2=''        # /usr/share/themes/*/gtk-2.0 (* = theme)
+#	THEME_JWM=''         # /usr/share/jwm/themes/*-jwmrc, /root/.jwm/themes/*-jwmrc (* = theme)
+#	THEME_JWM_BUTTONS='' # /usr/share/jwm/themes_window_buttons
+#	THEME_GTK_ICONS=''   # /usr/share/icons
+#	THEME_DESK_ICONS='   # /usr/local/lib/X11/themes
```
01micko: missing `'`

01micko (changes requested):
Didn't really check dpup specific stuff but other than the missing `'` LGTM

woof-code/rootfs-packages/ptheme/usr/sbin/ptheme_gtk
dimkr: Typo since 79356aa110

01micko (approved):

dimkr:
> Didn't really check dpup specific stuff but other than the missing `'` LGTM

The dpup-specific stuff is just reduction of available artwork - only 5 matching sets of themes are kept and everything else is deleted.

I've minimized the changes here in several cycles of cleanup, then moved these to separate PRs:

#3667 
#3666
#3664
#3663

PR #3575, opened 2022-11-04
If present, offer Synaptic instead of PPM in missingdefaultapp

dimkr:


peabee (changes requested):
I don't think that specific apps should be hard coded like this. Why only synaptic? What about other "alternative" PM's?

dimkr:
> I don't think that specific apps should be hard coded like this. Why only synaptic? What about other "alternative" PM's?

If Synaptic is present, there's no reason to recommend anything else because 1) PPM is super broken and 2) if you have Synaptic, you don't have vpm, yum, dnf, pacman or whatever.

lakshayrohila (changes requested):
You'll have to change the labels as well (PPM would be shown otherwise).

dimkr:
> You'll have to change the labels as well (PPM would be shown otherwise).

I agree, but missingdefaultapp doesn't support translations and I tried not to make the problem worse by adding more strings that need to be translated 😄

lakshayrohila:
> I tried not to make the problem worse by adding more strings that need to be translated 😄

Will those strings be needed to be translated as well? I mean they are names.

And even if they were to be translated, you could use a general name like 'Package Manager' instead of using name of a specific application? 💡

dimkr:
> you could use a general name like 'Package Manager' instead of using name of a specific application? bulb

Yep, that's why I kept the "Puppy Package Manager" label.

lakshayrohila:
> > you could use a general name like 'Package Manager' instead of using name of a specific application? bulb
> 
> Yep, that's why I kept the "Puppy Package Manager" label.

Ok, I understand it now. What you mean by PPM is not the application PPM but rather package manager for puppy. I think the PR is good otherwise (way better than my approach)!

lakshayrohila (approved):

dimkr:
@peabee Would this change (or something else, in this spirit) address your concerns?

```
PM=`command -v synaptic`
[ -z "$PM" ] && PM=`command -v vpm`                 <--
[ -z "$PM" ] && PM=ppm
```

peabee:
Not really - I think we need a much more general mechanism - which we already have - we just need a "defaultpackagemanager" added to the set in /usr/local/bin (rootfs-skeleton) and an additional setting for defaultpackagemanager in DEFAULTAPPS= in _00build.conf defaulting to ppm unless set by the system builder.
puppyapps may also need looking at.

dimkr:
Do you see a reason why would anyone build a Puppy with PPM and (objectively superior) package manager x, but with PPM as the so-called default?

peabee:
Yes because PackageManageX is likely to be much less flexible than PPM - i.e. only able to install one type of package from one repo location - whereas PPM can install many types of package from many different repos and although this might be considered "dangerous" it is what the type of people who use Puppy want to do.
If you don't want the flexibility that Puppy can provide then you're probably better off sticking to one of the more mainstream (and possibly more boring) distros.
I also think some of your rhetoric about how awful PPM is, is overblown and biased.
Having said all that, I very rarely use Package Managers of any sort as I prefer to keep my savearea unpolluted with code, especially large complex apps. And I certainly don't want dependencies being installed without my explicit agreement.

dimkr:
This message is for new users, not for advanced users who know they should use PPM when they want to break the system. Why lead new users to a trap?

dimkr:
4ec3256 is good enough for me.

PR #3577, opened 2022-11-05
Avoid excessive use of FTP

ChrysoliteAzalea:
Hello everyone! Today, I've looked at the woof-CE source code, and found that it excessively uses insecure protocols that don't verify the integrity of received data and therefore vulnerable to malicious injections from untrusted networks. I would like to avoid that, because woof-CE deals with very sensitive software code (such as the kernel).

### DIFF
--- a/woof-distro/x86_64/slackware64/15.0/DISTRO_PET_REPOS
+++ b/woof-distro/x86_64/slackware64/15.0/DISTRO_PET_REPOS
@@ -29,16 +29,13 @@ z|https://raw.githubusercontent.com/puppylinux-woof-CE/woof-CE/${WCE_BRANCH}/woo
 #   ex: Packages-puppy-4-official (note, url paths are in the database)
 
 PET_REPOS='
-z|http://distro.ibiblio.org/puppylinux|Packages-puppy-*-official
-z|http://ftp.nluug.nl/ftp/pub/os/Linux/distr/puppylinux|Packages-puppy-*-official
-z|http://ftp.cc.uoc.gr/mirrors/linux/puppylinux|Packages-puppy-*-official
-z|ftp://mirror.aarnet.edu.au/pub/puppylinux|Packages-puppy-*-official
-z|http://distro.ibiblio.org/puppylinux|Packages-puppy-*-official
-z|http://ftp.nluug.nl/ftp/pub/os/Linux/distr/puppylinux|Packages-puppy-*-official
-z|ftp://ftp.vcu.edu/pub/gnu+linux/puppylinux|Packages-puppy-*-official
-z|http://ftp.cc.uoc.gr/mirrors/linux/puppylinux|Packages-puppy-*-official
-z|http://mirror.aarnet.edu.au/pub/puppylinux|Packages-puppy-*-official
-z|http://mirror.internode.on.net/pub/puppylinux|Packages-puppy-*-official
+z|https://distro.ibiblio.org/puppylinux|Packages-puppy-*-official
+z|https://ftp.nluug.nl/ftp/pub/os/Linux/distr/puppylinux|Packages-puppy-*-official
+z|https://ftp.cc.uoc.gr/mirrors/linux/puppylinux|Packages-puppy-*-official
+z|https://distro.ibiblio.org/puppylinux|Packages-puppy-*-official
+z|https://ftp.nluug.nl/ftp/pub/os/Linux/distr/puppylinux|Packages-puppy-*-official
+z|https://ftp.cc.uoc.gr/mirrors/linux/puppylinux|Packages-puppy-*-official
+z|https://mirror.aarnet.edu.au/pub/puppylinux|Packages-puppy-*-official
 z|https://raw.githubusercontent.com/puppylinux-woof-CE/woof-CE-noarch/master|Packages-puppy-noarch-official
 '
 
--- a/woof-distro/x86_64/slackware64/15.0/DISTRO_PET_REPOS
+++ b/woof-distro/x86_64/slackware64/15.0/DISTRO_PET_REPOS
@@ -32,9 +32,6 @@ PET_REPOS='
 z|https://distro.ibiblio.org/puppylinux|Packages-puppy-*-official
 z|https://ftp.nluug.nl/ftp/pub/os/Linux/distr/puppylinux|Packages-puppy-*-official
 z|https://ftp.cc.uoc.gr/mirrors/linux/puppylinux|Packages-puppy-*-official
-z|https://distro.ibiblio.org/puppylinux|Packages-puppy-*-official
-z|https://ftp.nluug.nl/ftp/pub/os/Linux/distr/puppylinux|Packages-puppy-*-official
-z|https://ftp.cc.uoc.gr/mirrors/linux/puppylinux|Packages-puppy-*-official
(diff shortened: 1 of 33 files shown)


woof-code/packages-templates/fonts-noto_FIXUPHACK
```diff
@@ -13,7 +13,7 @@ cat << EOF > etc/fonts/conf.d/01-noto.conf
 <?xml version="1.0"?>
 <!DOCTYPE fontconfig SYSTEM "fonts.dtd">
 <fontconfig>
-  <its:rules xmlns:its="http://www.w3.org/2005/11/its" version="1.0">
+  <its:rules xmlns:its="https://www.w3.org/2005/11/its" version="1.0">
```
dimkr: Changes like this are breaking changes

dimkr (changes requested):
1. Please roll back changes that have mo security implications, like changes to comments
2. Without a script that checks that every HTTP->HTTPS URL both supports HTTPS and has a valid (= not self-signed) certificate, this PR is super dangerous

dimkr:
Word of caution: at some point I tried to remove the `--no-check-certificate` in https://github.com/puppylinux-woof-CE/woof-CE/blob/38e2eaf135598b2a8f48d1e0955f6955f069c566/woof-code/support/download_file.sh#L32, but it looks like in some cases, woof-CE uses HTTPS URLs with broken certificates.

ChrysoliteAzalea:
OK, I've got that. I think I'll leave all **http://** links intact for some time, and only change the **ftp://** and **git://** links (there is a **git:// link that points to GitHub, that supports HTTPS), because FTP is a really old and insecure protocol.

woof-code/packages-templates/fonts-noto_FIXUPHACK
ChrysoliteAzalea: Done -- all http links are left intact now.

dimkr:
AFAIK git:// is plaintext too

woof-distro/x86_64/ubuntu/jammy64/DISTRO_PET_REPOS
```diff
@@ -32,10 +32,10 @@ PET_REPOS='
 z|http://distro.ibiblio.org/puppylinux|Packages-puppy-*-official
 z|http://ftp.nluug.nl/ftp/pub/os/Linux/distr/puppylinux|Packages-puppy-*-official
 z|http://ftp.cc.uoc.gr/mirrors/linux/puppylinux|Packages-puppy-*-official
-z|ftp://mirror.aarnet.edu.au/pub/puppylinux|Packages-puppy-*-official
+z|https://mirror.aarnet.edu.au/pub/puppylinux|Packages-puppy-*-official
 z|http://distro.ibiblio.org/puppylinux|Packages-puppy-*-official
 z|http://ftp.nluug.nl/ftp/pub/os/Linux/distr/puppylinux|Packages-puppy-*-official
-z|ftp://ftp.vcu.edu/pub/gnu+linux/puppylinux|Packages-puppy-*-official
+z|https://ftp.vcu.edu/pub/gnu+linux/puppylinux|Packages-puppy-*-official
```
dimkr: Bad certificate

dimkr: But the download from this mirror **will** work because of [woof-CE/woof-code/support/download_file.sh](https://github.com/puppylinux-woof-CE/woof-CE/blob/38e2eaf135598b2a8f48d1e0955f6955f069c566/woof-code/support/download_file.sh#L32) 👎🏿

ChrysoliteAzalea: I wonder, should I remove that, or change it back to FTP? My main issue with FTP is its lack of integrity checks. If they are performed internally in woof-CE, using FTP can be... not that bad idea. For example, Debian uses HTTP because all packages are signed and checked after each download. Also, as far as I know, Wine can download some packages through HTTP and then check their checksums. So, if woof-CE checks the integrity of its downloads, then HTTP/FTP/Git Protocol aren't that much of an issue (in that case, I can simply close this pull request).

dimkr: woof-CE doesn't do GPG verification and integrity checks (hashes, etc') are not an option here, because these are not static files that don't change

dimkr: IMO 2-3 HTTPS mirrors in different continents are more than enough, so those that only speak FTP and HTTP can be removed

ChrysoliteAzalea: OK, all git:// and ftp:// were switched to https://. Also, I've switched some http:// to https://, and removed all mirrors I was not sure about.

ChrysoliteAzalea: Resolving the conversation as the problem mirror has been removed in the pull request.

ChrysoliteAzalea:
OK, all git:// and ftp:// were switched to https://. Also, I've switched some http:// to https://, and removed all mirrors I was not sure about.

woof-distro/x86_64/slackware64/15.0/DISTRO_PET_REPOS
```diff
@@ -29,16 +29,13 @@ z|https://raw.githubusercontent.com/puppylinux-woof-CE/woof-CE/${WCE_BRANCH}/woo
...
-z|http://distro.ibiblio.org/puppylinux|Packages-puppy-*-official
-z|http://ftp.nluug.nl/ftp/pub/os/Linux/distr/puppylinux|Packages-puppy-*-official
-z|ftp://ftp.vcu.edu/pub/gnu+linux/puppylinux|Packages-puppy-*-official
-z|http://ftp.cc.uoc.gr/mirrors/linux/puppylinux|Packages-puppy-*-official
-z|http://mirror.aarnet.edu.au/pub/puppylinux|Packages-puppy-*-official
-z|http://mirror.internode.on.net/pub/puppylinux|Packages-puppy-*-official
+z|https://distro.ibiblio.org/puppylinux|Packages-puppy-*-official
+z|https://ftp.nluug.nl/ftp/pub/os/Linux/distr/puppylinux|Packages-puppy-*-official
+z|https://ftp.cc.uoc.gr/mirrors/linux/puppylinux|Packages-puppy-*-official
+z|https://distro.ibiblio.org/puppylinux|Packages-puppy-*-official
```
dimkr: Same as line 32?

ChrysoliteAzalea: Sorry, I've overlooked a lot of duplicate strings. Fixed.

ChrysoliteAzalea:
Also, as far as I checked, [archive.ubuntu.com](http://archive.ubuntu.com) doesn't support HTTPS, so in that case, HTTP is the only option.

I'm also researching the possibility of building a Puppy distribution from user's local sources, so it could be built without active Internet connection.

dimkr:
> I'm also researching the possibility of building a Puppy distribution from user's local sources, so it could be built without active Internet connection.

If you use file:// URLs, it will mostly work. There will be some exceptions (like petbuilds, the ad blocker's pinstall.sh, etc').

PR #3581, opened 2022-11-05
Update man cache if missing

dimkr:
Minimal fix, alternative to #3545 and #3547.

(Only works if using overlay and docx and nlsx are auto-loaded by init, and that's the case in dpup. AFAIK, dpup is the only Puppy configured like this.)


dimkr:
Now it's a clean, dpup-specific fix. Merging.

Issue #3590, opened 2022-11-08
Linux kernel reading guide: link broken in kerenel-kit's README


lakshayrohila:
There is one link broken in `kernel-kit/README`, line `100`, which says- **Linux Kernel Reading Guide**.

PR #3592, opened 2022-11-09
Allow merge2out-gui to be scrollable

lakshayrohila:
`merge2out-gui` currently doesn't use scrollbar, and hence there is no way that users may be able to see all of the options/buttons due to this. Nor does `gtkdialog` window resize or move on some (or all) versions.

This PR makes `merge2out-gui` use scrollbar to resolve this issue.

This PR also resolves (https://github.com/puppylinux-woof-CE/woof-CE/issues/3590).


lakshayrohila:
Is there anything I need to reconsider here, @dimkr?

dimkr:
Nope, looks fine IMO

lakshayrohila:
> Nope, looks fine IMO

Ok, thanks! I thought that the PR was not being accepted because of it having some issues. Relieved that there wasn't any.

dimkr:
I'm doing my best not to break stuff during @01micko's absence. It's always good to have a second pair of eyes before changes are merged.

I'll abuse my merge rights again, because these are two single-line changes.

PR #3621, opened 2022-11-17
Add group input;x;31;

peabee:
Required by Slackware eudev - /lib/udev/rules.d/50-udev-default.rules  i.e lines 29-30
SUBSYSTEM=="input", GROUP="input"
SUBSYSTEM=="input", KERNEL=="js[0-9]*", MODE="0664"

value from:
https://patchwork.ozlabs.org/comment/883148/

### DIFF
--- a/woof-code/rootfs-skeleton/etc/group
+++ b/woof-code/rootfs-skeleton/etc/group
@@ -28,3 +28,4 @@ tape::108:root
 tty::109:root,spot
 plugdev::110:root,spot
 bluetooth::111:root,spot
+input:x:31:
--- a/woof-code/rootfs-skeleton/etc/group
+++ b/woof-code/rootfs-skeleton/etc/group
@@ -28,4 +28,4 @@ tape::108:root
 tty::109:root,spot
 plugdev::110:root,spot
 bluetooth::111:root,spot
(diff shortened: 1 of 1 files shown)


dimkr (approved):
Looks fine to me, although I'd 1) uses the GID used by Slackware or some other distro and 2) add spot to this group if other distros normally add human users (i.e. not users like `nobody` and `daemon`) to this group

peabee:
> Looks fine to me, although I'd 1) uses the GID used by Slackware or some other distro and 2) add spot to this group if other distros normally add human users (i.e. not users like `nobody` and `daemon`) to this group

Good points - but how do you find info on 1) & 2) ???

woof-code/rootfs-skeleton/etc/group
```diff
@@ -28,3 +28,4 @@ tape::108:root
 tty::109:root,spot
 plugdev::110:root,spot
 bluetooth::111:root,spot
+input:x:31:
```
01micko: input:x:31:root,spot

01micko: stupid github!

`input:x:31:root,spot`

PR #3622, opened 2022-11-18
Align dpup with 0778965 and c10d742

dimkr:


peabee:
I decided this wasn't needed for group-debian because the 50-udev-default.rules was different to Slackware and didn't mention group input.....

dimkr:
dpup has an `input` group, it's added during the build. I only wanted to have spot in it, for consistency with Slacko.

PR #3635, opened 2022-11-21
Generate SHA512 checksums

lakshayrohila:
This PR allows Github Actions to generate and upload `SHA512` checksums for `.iso` and `.sfs` files.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -165,6 +165,10 @@ jobs:
       run: |
         for i in `ls ../woof-out_*/woof-output-*/*.{sfs,iso} 2>/dev/null`; do sudo mv -f $i .; done
         sudo mv -f ../woof-out_*/kernel-kit/output/kernel_sources-*.sfs .
+    - name: Generate SHA1 checksums
+      run: |
+        find . -type f -name '*.sfs' -exec sha1sum {} \; > SHA1checksums.sha1
+        find . -type f -name '*.iso' -exec sha1sum {} \; >> SHA1checksums.sha1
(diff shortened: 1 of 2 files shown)


dimkr:
SHA1 is insecure: there are tools that can generate a malicious file with the same SHA1 as a given file (https://shattered.io/) in finite time, and some applications warn when you try to use SHA1.

Switch to SHA256 or SHA512.

Also, I'd change to `.txt` or something, because some people download these files from Windows.

dimkr:
Also, I'd move this from build.yml to release.yml, because end-users are not supposed to download build artifacts, only releases. No reason to slow down all builds, if these files are unused.

In addition, if you want this PR to do anything, you'll need to upload the text file containing the hashes, here:

https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/.github/workflows/release.yml#L76-L78

lakshayrohila:
> SHA1 is insecure: there are tools that can generate a malicious file with the same SHA1 as a given file (https://shattered.io/) in finite time, and some applications warn when you try to use SHA1.
> 
> Switch to SHA256 or SHA512.
> 
> Also, I'd change to `.txt` or something, because some people download these files from Windows.

Done.

> Also, I'd move this from build.yml to release.yml, because end-users are not supposed to download build artifacts, only releases. No reason to slow down all builds, if these files are unused.
> 
> In addition, if you want this PR to do anything, you'll need to upload the text file containing the hashes, here:
> 
> https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/.github/workflows/release.yml#L76-L78

Done.

Issue #3639, opened 2022-11-22
Error while installation from PPM in jammy64. Possibly petget error.


lakshayrohila:
I tried building jammy64 with current defaults and installed some programs using it. Don't know how many it successfully installed, but installing few it broke the system. I tried installing `apt` once for getting the same results and the next time `nano`. The common thing here, as I saw it, is `/bin/`. The executables of both programs are put in `/bin` and using `ls /bin/` after that listed only that program (atleast for `nano`; I think after installing `apt` even `ls` was not even found). **A PPM bug maybe?**

I was reading [this](https://github.com/puppylinux-woof-CE/woof-CE/discussions/3305) discussion some other day, and hence I see why there might be a bug:-
> Nobody here is working towards improving PPM, and it's essentially unmaintained for years.

**So, what might be the future of PPM in next releases?** I think it might be good to uninstall it. **But, how?** And which package manager to use instead of it? I tried building with `pkg` but it took forever to even update its repos without showing any movement, as if it was hanged; maybe due to  some bad settings used in builds? And what about `apt`? Could that be added as well (I'd actually like if we could)?

Actual question:-
1. What's the error (see https://github.com/puppylinux-woof-CE/woof-CE/issues/3639#issuecomment-1323985937) and how to resolve it?

Additional questions are (if this `petget` issue will never be handled):-
1. Should `PPM` be removed?
2. Should `pkg` be added?
3. Should `apt` be added?

lakshayrohila:
`pkg ru` seems to work fine now after I tried it once again. Just took nearly 10-20secs. Still getting an error, which seems to say:-
```
Checking pkgs
/usr/local/petget/0setup: line 337: support/findpkgs: No such file or directory
Error 2: line 7853, Code error!
```

lakshayrohila:
This error seems to be related to `petget`. Renaming the error.

dimkr:
You can't just install apt through PPM to get apt support, it requires more than that. And PPM is so bad at installing packages with deep dependency trees I wouldn't be surprised if this breaks your system.

> And what about apt?

It's possible to add apt to jammy64, it's supported and works well as long as `USR_SYMLINKS=yes`.

[...]

PR #3640, opened 2022-11-22
Allow apt support even if using .pet packages and enable in jammy64

dimkr:
See #3639.


lakshayrohila (commented):
I think the PR is great! Just add in the woof-CE documentation about .pet conflicts with `apt`.

Issue #3646, opened 2022-11-23
Palemoon petbuild fails to download files.


lakshayrohila:
I recently found that a `jammy64` build had failed due to palemoon not being able to download its package. Found that it was due to palemoon structure of links. What it does is provide link to the latest version of the package at different site; and then when a new release is available, it moves it to `archive.palemoon.org`. Not just this, the latest release is also not available under its archives. So clearly, there is no way we could keep an eye on palemoon everytime in case it had been updated and the file moved.

But for a complete fix, I have dropped a post on `forum.palemoon.org` describing about our plans for including palemoon in puppylinux, and also have requested them to upload their latest releases to `archive.palemoon.org`; so as to create more 'stable' links. The post will be first reviewed by some executives there on the forum, then uploaded to [General Discussion](https://forum.palemoon.org/viewforum.php?f=65) sub-forum, then reviewed by someone and then worked upon. This will take sometime.

https://github.com/puppylinux-woof-CE/woof-CE/pull/3645 tries to mitigate the problem for sometime.

lakshayrohila:
**Update:** My post has been accepted. Here it is- https://forum.palemoon.org/viewtopic.php?f=65&t=29106.

Waiting for someone to reply there.

PR #3647, opened 2022-11-23
Fix https://github.com/puppylinux-woof-CE/woof-CE/issues/3646.

lakshayrohila:
Add `archive.palemoon.org` link as well, as a fallback, if `linux.palemoon.org` one is not found.


lakshayrohila:
@dimkr please review the PR.

PR #3654, opened 2022-11-26
Add 'Refresh Menus' menu item.

lakshayrohila:


woof-code/rootfs-skeleton/etc/xdg/templates/_root_.jwmrc
```diff
@@ -29,6 +29,8 @@ PUPPYMENU jwm-xdgmenu /etc/xdg/menus/puppy-fun.menu
 
 <Separator/>
 
+<Program label="Refresh Menus" icon="menu.svg">/usr/sbin/fixmenus</Program>
```
dimkr: This doesn't do anything because (AFAIK) fixmenus doesn't run `jwm -reload` - maybe you'd want `sh -c "fixmenus; jwm -reload"` here to avoid changes to fixmenus

dimkr: And I wonder what's the motivation for this PR - if the menu generator has a bug that requires some manual modification of .desktop files followed by `fixmenus`, why don't we fix the menu generator instead?

lakshayrohila: I think some people (including me) might want to make changes to .desktop files manually. There are many programs out there which might not run as is from puppy and may need manual modification from the user. Just for example- most electron apps would need `--no-sandbox` option added.

I think it might be great to allow the users to run this menu item instead of reloading window everytime.

dimkr: I bet most users who know why and how to fix broken .desktop files already know how to `fixmenus; jwm -reload` and don't really **need** this menu entry. However, the Puppy menu is already super long and annoying with small displays, so an extra entry doesn't change much.

I have no objection to this PR, but @01micko and @peabee might have different opinions.

lakshayrohila:
I think this PR is not very much needed. I see that sometimes (or maybe most of the times) the menu gets reloaded.

PR #3655, opened 2022-11-26
Add DOCS_URL support.

lakshayrohila:
This PR includes a new variable, named `DOCS_URL`. This shall allow the developers to overwrite the current woof-CE docs according to their own needs (though it is good, still at times people may need it).

woof-CE `README.md` is updated accordingly, also giving the developers warning to use the option carefully and letting them know to keep in mind about ensuring the integrity of the Puppy they build.


lakshayrohila:
I am confused here whether I should use `unzip doc.zip -d sandbox3/rootfs-complete/usr/share/` or ```
unzip doc.zip
mv doc sandbox3/rootfs-complete/usr/share/
```?

The second option would be more restrictive so that the users are able to modify docs only with this option, but still when I think twice about this, I remember that it does not matter to woof-CE and the developers should themselves remember to maintain the integrity of the Puppy they build.

What do you think?

File: woof-code/3builddistro [hunk 1]
```diff
@@ -493,6 +493,16 @@ for i in $(ls sandbox3/rootfs-complete/usr/share/locale); do
 	case $i in en|en_US|locale.alias) continue ;; esac
 	rm -rf sandbox3/rootfs-complete/usr/share/locale/$i
 done
+
+#overwrite current documentation if DOCS_URL present
+if [ ! -z "$DOCS_URL" ]; then
```
dimkr: `! -z` is the same as `-n`

woof-code/3builddistro
```diff
@@ -493,6 +493,16 @@ for i in $(ls sandbox3/rootfs-complete/usr/share/locale); do
...
+	echo "Overwriting current documentation with developer provided $DOCS_URL..."
+	rm -rf sandbox3/rootfs-complete/usr/share/doc
+	wget -t 3 -T 60 -O doc.zip "$DOCS_URL"
```
dimkr: 1. Add caching of some sort (download to local-repositories and skip download if already there?)
2. Fail 3builddistro if wget failed

lakshayrohila: @dimkr, please give any example for (1).

dimkr: https://github.com/puppylinux-woof-CE/woof-CE/blob/4c122eac026f65501b4a85c6b8c08a28377ce309/woof-code/support/pc_image.sh#L62

lakshayrohila: @dimkr, I have tried to add a caching system, which saves downloaded files as `../../local-repositories/$DOCS_URL.doc.zip`. But I am unsure if this will work or not, as the link may also contain characters such as `:` and `/`.

dimkr: You can generate a unique file name instead, i.e.`$(echo "$DOCS_URL" | sha256sum | cut -f 1 -d ' ').zip`

dimkr: Or, just take the base name and assume it's unique, i.e. `../../local-repositories/${DOCS_URL##*/}.doc.zip`

woof-code/3builddistro
```diff
@@ -493,6 +493,23 @@ for i in $(ls sandbox3/rootfs-complete/usr/share/locale); do
 	case $i in en|en_US|locale.alias) continue ;; esac
 	rm -rf sandbox3/rootfs-complete/usr/share/locale/$i
 done
+
+#overwrite current documentation if DOCS_URL present
+if [ -n "$DOCS_URL" ]; then
+	echo "Overwriting current documentation with developer provided $DOCS_URL..."
+	rm -rf sandbox3/rootfs-complete/usr/share/doc
+
+	[ -f "../../local-repositories/$DOCS_URL.doc.zip" ] || wget_output=$(wget -t 3 -T 60 -O "../../local-repositories/$DOCS_URL.doc.zip" "$DOCS_URL")
```
dimkr: The correct way to check if wget has failed is:

```
	wget -t 3 -T 60 -O "../../local-repositories/$DOCS_URL.doc.zip" "$DOCS_URL"
	if [ $? -ne 0 ]; then
```

(exit code, not output)

woof-code/3builddistro
```diff
@@ -493,6 +493,23 @@ for i in $(ls sandbox3/rootfs-complete/usr/share/locale); do
...
+
+	if [ $wget_output -ne 0 ]; then
+		echo "ERROR: wget failed. Read above lines to see the error."
+		exit 1
+	fi
+
+	unzip doc.zip -d sandbox3/rootfs-complete/usr/share/
```
dimkr: These two lines are broken, you're downloading to one place and attempting to extract from another

woof-code/3builddistro
lakshayrohila: Updated with a newer caching system.

woof-code/3builddistro
lakshayrohila: Does it check with output? I thought `$()` returned exit code. Will backticks (`) return exit code then?

(Sorry, I am new to bash, actually am a programmer of C.)

woof-code/3builddistro
lakshayrohila: I updated it now. The earlier caching system was completely bad.

woof-code/3builddistro
```diff
@@ -508,9 +508,11 @@ if [ -n "$DOCS_URL" ]; then
 	if [ "$CACHE_PERSONAL_DOCS" = 'y' ]; then
 		mkdir -p "../../local-repositories/personal_docs/$DOCS_URL"
 		DOWNLOADED_DOC_FILENAME="../../local-repositories/personal_docs/$DOCS_URL/doc.zip"
-		[ -f $DOWNLOADED_DOC_FILENAME ] || wget_output=$(wget -t 3 -T 60 -O $DOWNLOADED_DOC_FILENAME "$DOCS_URL")
+		[ -f $DOWNLOADED_DOC_FILENAME ] || wget -t 3 -T 60 -O $DOWNLOADED_DOC_FILENAME "$DOCS_URL"
+		wget_return_code=$?
```
dimkr: `wget_return_code` is undefined if `[ -f $DOWNLOADED_DOC_FILENAME ]`!

```
                wget_return_code=0         <--
		[ -f $DOWNLOADED_DOC_FILENAME ] || wget -t 3 -T 60 -O $DOWNLOADED_DOC_FILENAME "$DOCS_URL"
		wget_return_code=$?
```

woof-code/3builddistro
```diff
@@ -508,9 +508,11 @@ if [ -n "$DOCS_URL" ]; then
 	if [ "$CACHE_PERSONAL_DOCS" = 'y' ]; then
```
dimkr, line 525: IMO this cache should be enabled by default, not the other way around. I know @peabee does local builds and prefers to cache as much as possible, and repeated downloads will kill the GitHub Actions traffic limits.

woof-code/3builddistro
lakshayrohila: How can I fix this? Use `if` maybe?

dimkr: Either initialize it before the condition (see the example I gave you), or just use `if`

lakshayrohila: I have used `if`. Seems like my bash scripting knowledge is still a trash :frowning_face:! Sorry for constant silly bugs.

woof-code/3builddistro
```diff
@@ -493,6 +493,46 @@ for i in $(ls sandbox3/rootfs-complete/usr/share/locale); do
...
+	y|n) echo -n ;;
+	*) CACHE_PERSONAL_DOCS=y ;;
+esac
+
+if [ -n "$DOCS_URL" ]; then
+	echo "Overwriting current documentation with developer provided $DOCS_URL..."
+	rm -rf sandbox3/rootfs-complete/usr/share/doc
+	
+	if [ "$CACHE_PERSONAL_DOCS" = 'y' ]; then
+		mkdir -p "../../local-repositories/personal_docs/$DOCS_URL"
```
dimkr: If `DOCS_URL` is `https://google.com/a/b/c`, does this line try to create `../../local-repositories/personal_docs/https://google.com/a/b/c"`? This won't work.

lakshayrohila: It creates `https:/google.com/a/b/c`. I have tested this part of code in a seperate test directory. First `https:` directory is created, inside it `google.com` and so on.

I did it this way because I couldn't think of any better option.

dimkr: > I did it this way because I couldn't think of any better option.

Why not `${DOCS_URL##*/}`, as I suggested in some other comment?

lakshayrohila:
I converted it into a draft because I got a better idea to allow the user to modify docs.

lakshayrohila:
I have completed with my idea. Here is what I want:- https://github.com/lakshayrohila/woof-CE/tree/Add-DOCS_URL#getting-documentation-from-a-specified-url.

@dimkr, @01micko, @peabee please review. Thanks!

lakshayrohila:
@dimkr, @peabee, @01micko anything wrong in this PR?

dimkr:
> @dimkr, @peabee, @01micko anything wrong in this PR?

I didn't take a super close look after the changes. But at a quick glance - I think you should use .tar.{gz,xz,zstd} and not .zip because tar preserves metadata like file ownership and permissions.

IMO this PR is generally a bad idea, because Puppy's history proves that anything that comes from outside of woof-CE (like stuff shared on the forums) simply rots away and runs out of sync with Puppy. Fossapup was built with modifications that were never merged into woof-CE, so now it's' impossible to rebuild it. Old .pet packages have bugs that won't get fixed because nobody knows how to build them, and the source code of some is lost. This external documentation - who has access to it? Is there any formal process for updating the documentation when woof-CE changes? Does it have some kind of version control? Is there a way to take this documentation and create a "branch" for Puppy x rather than y?

lakshayrohila:
> I didn't take a super close look after the changes. But at a quick glance - I think you should use .tar.{gz,xz,zstd} and not .zip because tar preserves metadata like file ownership and permissions.
IMO this PR is generally a bad idea, because Puppy's history proves that anything that comes from outside of woof-CE (like stuff shared on the forums) simply rots away and runs out of sync with Puppy. Fossapup was built with modifications that were never merged into woof-CE, so now it's' impossible to rebuild it. Old .pet packages have bugs that won't get fixed because nobody knows how to build them, and the source code of some is lost. This external documentation - who has access to it? Is there any formal process for updating the documentation when woof-CE changes? Does it have some kind of version control? Is there a way to take this documentation and create a "branch" for Puppy x rather than y? 

I wanted to create this feature because as you said once in puppy forums, there will be few changes needed for the documentation for certaing puppys. I do not want to create it for my own use, at least not for now, but someone sometime may need it. If you say so, I'll close this PR.

dimkr:
> If you say so, I'll close this PR.

Don't take my opinion so seriously 😄

lakshayrohila:
> I didn't take a super close look after the changes. But at a quick glance - I think you should use .tar.{gz,xz,zstd} and not .zip because tar preserves metadata like file ownership and permissions.

@dimkr, I have updated the file with .tar.{gz,xz,bz2} file extensions.

lakshayrohila:
There are still few bugs. I'll correct them and then open this PR again.

lakshayrohila:
I have tested this in an environment very much like what 3buildistro will get (setting up local-cache directories and sandbox3/..., so that the PR can be thoroughly examined). It works completely fine now.

@dimkr, @peabee please review. Here are my plans (or what I want this PR to do):- https://github.com/lakshayrohila/woof-CE/tree/Add-DOCS_URL#getting-documentation-from-a-specified-url.

dimkr:
My opinion hasn't changed, I still don't like the idea of documentation maintained outside of woof-CE, or at least, outside of the puppylinux-woof-CE organization. It's impossible to collaborate on a .zip file, it will run out of sync with woof-CE, and it won't be reusable in other woof-CE build configurations.

(Haven't looked at code changes since the last time I reviewed this PR)

PR #3664, opened 2022-11-27
Make the default tray layout closer to the BionicPup and FossaPup one

dimkr:
This shouldn't affect any Puppy built with pTheme enabled.


lakshayrohila (approved):

lakshayrohila:
Looks great!

PR #3666, opened 2022-11-27
Remove symlink emblems from the default ROX-Filer pinboard

dimkr:
This shouldn't affect  any Puppy built with pTheme enabled.


lakshayrohila (approved):
Looks great to me! Many distros would actually need this.

dimkr:
> Looks great to me! Many distros would actually need this.

Actually, no. pTheme is sort of mandatory in every woof-CE build, because the pre-pTheme tools are broken. And pTheme always replaces the pinboard with one of its presets, so currently nobody uses the PuppyPin modified by this PR.

This PR will be useful after https://github.com/puppylinux-woof-CE/woof-CE/pull/3573 is merged and won't change anything for any current woof-CE user.

PR #3671, opened 2022-11-29
Make upup look more like FossaPup

dimkr:
Uses Numix (not the "dusk" variant, because only the former is available as an Ubuntu package), PMaterial, KDE cursors and @01micko's 431-dark.svg. I'd say this combination looks good, demonstrates how Puppy can be customized to be not-so-ugly, and it's pretty close to FossaPup.

jammy64


lakshayrohila (approved):
I always loved Fossapup theme over the theme I set for Jammypup. Looks great! Thanka for the PR!

PR #3672, opened 2022-11-29
Switch to the Pale Moon petbuild in {bionic,focal}64 too

dimkr:
@lakshayrohila Any reason not to do this?


lakshayrohila:
@dimkr, I don't think this might cause any problem. Maybe, you need to first try. I am currently accepting this PR, I don't think this will break anything.

lakshayrohila (approved):

dimkr:
Tested on jammy64 and bionic64, works fine. It's good to get rid of that ancient browser!

PR #3683, opened 2022-12-03
Add s15pup builds

peabee:


dimkr:
Do you want me to add them to the schedule of periodic builds? I can reduce the frequency of `slackware{,64}/15.0` to once a month and add a bi-weekly build of s15pup{32,64}.

woof-distro/x86/slackware/s15pup32/_00build.conf
```diff
@@ -0,0 +1,323 @@
...
+##   but you can specify a default app if you wish...
+## If you specify a value it will override anything that previously
+##   set that value in the corresponding script...
+## These are the current default*apps (scripts) in /usr/local/bin
+DEFAULTAPPS="
+defaultarchiver=
+defaultaudioeditor=
+defaultaudiomixer=
+defaultaudioplayer=
+defaultbrowser=light
```
dimkr, line 90: You can use @lakshayrohila's palemoon petbuild if you want a full-featured but relatively small browser without duplicate effort. The Ubuntu-based builds already use it and it works all the way back to 18.04.

woof-distro/x86_64/slackware64/s15pup64/_00build.conf
```diff
@@ -0,0 +1,304 @@
...
+# build docx, nlsx? yes/no - any other value = no
+BUILD_DOCX=no
+BUILD_NLSX=yes
+
+## include devx SFS in ISO?
+DEVX_IN_ISO=no
+
+## packages to build from source
+PETBUILD_GTK=2
+PETBUILDS="busybox aaa_pup_c gtkdialog jwm puppy_flat_icons puppy_standard_icons gtk_theme_flat_grey_rounded gtk_theme_gradient_grey gtk_theme_polished_blue xdg-puppy-jwm firewallstatus powerapplet_tray mtpaint"
```
dimkr, line 49: There are so many more that you can use instead of ancient .pet packages ... xlockmore, rxvt-unicode, transmission, Xdialog, yad, netmon_wce ... the list is long. Some petbuilds carry important fixes if you care about CPU usage, security and battery life - for example, the xlockmore petbuild supports power saving while the old .pet packages don't.

woof-distro/x86/slackware/s15pup32/_00build.conf
lakshayrohila: @dimkr That palemoon PETBUILDS is only for x86_64.

peabee:
This is a record of what has been built - future changes may be possible if they are needed due to user feedback.

I prefer to stick with the Light adrv sfs - it performs the functions of a basic browser and is half the size of Palemoon and easily replaced - sfs's of up-to-date versions of Chromium, Firefox and Palemoon are available to download.

I prefer to stick with the current mix of pets and petbuilds - petbuilds are great but slow down the build process considerably - if user feedback shows that an "ancient" (but well tested) .pet needs to move to a petbuild then that can be considered.

I don't think a periodic build is necessary as the S15Pup build is 100% Woof-CE with no added wrinkles so periodic builds won't help Woof-CE and could be confused with the official builds which will continue to be locally built and uploaded to SourceForge.

dimkr:
> if user feedback shows that an "ancient" (but well tested) .pet needs to move to a petbuild then that can be considered.

Why wait for users to tell you the old xlockmore package doesn't support DPMS and eats battery life of old laptops with bad batteries, if we already know this? (And that's only one example)

> petbuilds are great but slow down the build process considerably

A Thinkpad X220 with an old SSD can complete a clean woof-CE run in 10-15m. But oh well, I guess there's no way on earth I can convince you to reconsider getting rid of those old and buggy packages.

> I don't think a periodic build is necessary as the S15Pup build is 100% Woof-CE

If it's not built periodically, it will break over time and rot away just like all Ubuntu-based releases. If we don't have automated periodic builds, then it's up to you to ensure these configuration files keep working in your local setup.

dimkr:
@peabee https://github.com/puppylinux-woof-CE/woof-CE/pull/2953 is a good example of a bug that's super likely to be present in S15Pup but fixed in a petbuild. (Again, this is only one example: the petbuilds fix multiple decade-old bugs in Puppy.)

peabee:
> @peabee #2953 is a good example of a bug that's super likely to be present in S15Pup but fixed in a petbuild. (Again, this is only one example: the petbuilds fix multiple decade-old bugs in Puppy.)

line 292 of __00build.conf:
sed -i 's%#DISABLE_LOCK%DISABLE_LOCK%' etc/acpi/acpi.conf

probably prevents the "bug"....

dimkr:
> probably prevents the "bug"....

I don't see how exactly.

Boot S15Pup on a laptop, hit Super+L and close the lid. The laptop won't sleep and you'll see a password prompt when you open the lid. That's bug 1.

Close the window instead of entering a password. Hit Super+L  again, and nothing will happen. Close the lid, and nothing will happen. That's bug 2.

peabee:
Tweaks can come once the basics are merged - can I merge as is? Also the voidpup pull request?

dimkr:
> can I merge as is?

IMO yes

peabee:
Both S15Pup64 and S15Pup32 build correctly. Thanks.

PR #3689, opened 2022-12-04
Change to flat connman-ui icons

dimkr:
The current icons don't well work with some themes because they have too much detail and don't scale nicely, or the contrast is bad.

Before:

before

After:

after


dimkr:
Meh, f498125 is not good. The network tray icon is centered vertically but too small.

image

EDIT: the old netmon_wce icons are much better

image

dimkr:
@lakshayrohila What do you think about this one? It makes Numix and Breeze more usable.

lakshayrohila (approved):
@dimkr, seems great! 👍

PR #3692, opened 2022-12-05
Add Kubuntu theme.

lakshayrohila:
I created this theme sometime earlier, which seems to resemble theme that of Kubuntu, which I now use in all of the puppies I use.


lakshayrohila:
Seems like I created a branch from wrong branch. Correcting it and reopening the PR.

peabee (commented):
I think this should be a separate rootfs-package rather than part of jwm_config ....

PR #3694, opened 2022-12-05
Add void build dependencies to CI

dimkr:


lakshayrohila (approved):
@peabee, this PR will do it for you! You could add other build dependencies at line 88 of build.yml in the same way.

peabee (approved):
thanks

PR #3700, opened 2022-12-06
Installer improvements

dimkr:
This PR adds:
- An important reliability fix for bootflash
- Fix for bootflash failure when vmlinuz and initrd.gz are present in the ESP but not in pdrv
- efilinux support in bootflash, if present (= only dpup)
- syslinux support in bootflash, if present (= only dpup)
- Cleanup of broken buttons if grub4dos, cdrecord and xorriso are not present (= only dpup)


dimkr:
@peabee Objections to this PR? Shouldn't affect your builds, only adds two bug fixes

peabee (commented):
No comment. Maybe some worry about how much testing these sort of changes get.

dimkr:
> No comment. Maybe some worry about how much testing these sort of changes get.

The two fixes are super easy to test, they're safe and well-tested.

The new features (which are enabled only in dpup right now) are tested too, with multiple computers and both BIOS and UEFI. Everything seems to work well, including support for microcode loading, F2FS support, installation to a flash drive and installation from this flash drive to another.

Issue #3701, opened 2022-12-06
apt: key not found; Jammy64


lakshayrohila:
Don't know if this error is related to woof-CE or not, but the error at the bottom is related. This is the error:-
```
apt update
Hit:1 http://archive.ubuntu.com/ubuntu jammy InRelease
Get:2 http://packages.microsoft.com/repos/code stable InRelease [10.4 kB]
Get:3 http://archive.ubuntu.com/ubuntu jammy-updates InRelease [114 kB]
Err:2 http://packages.microsoft.com/repos/code stable InRelease
  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY EB3E94ADBE1229CF
Get:4 http://archive.ubuntu.com/ubuntu jammy-security InRelease [110 kB]
Get:5 http://archive.ubuntu.com/ubuntu jammy-updates/main amd64 Packages [755 kB]
Get:6 http://archive.ubuntu.com/ubuntu jammy-updates/main i386 Packages [389 kB]
Get:7 http://archive.ubuntu.com/ubuntu jammy-updates/main Translation-en [169 kB]
Get:8 http://archive.ubuntu.com/ubuntu jammy-updates/main amd64 c-n-f Metadata [11.3 kB]
Get:9 http://archive.ubuntu.com/ubuntu jammy-updates/universe i386 Packages [555 kB]
Get:10 http://archive.ubuntu.com/ubuntu jammy-updates/universe amd64 Packages [761 kB]
Get:11 http://archive.ubuntu.com/ubuntu jammy-updates/universe amd64 c-n-f Metadata [14.0 kB]
Get:12 http://archive.ubuntu.com/ubuntu jammy-updates/multiverse i386 Packages [1,712 B]
Get:13 http://archive.ubuntu.com/ubuntu jammy-updates/multiverse amd64 Packages [7,300 B]
Get:14 http://archive.ubuntu.com/ubuntu jammy-updates/restricted i386 Packages [24.7 kB]
Get:15 http://archive.ubuntu.com/ubuntu jammy-updates/restricted amd64 Packages [484 kB]
Get:16 http://archive.ubuntu.com/ubuntu jammy-security/main i386 Packages [221 kB]
Get:17 http://archive.ubuntu.com/ubuntu jammy-security/main amd64 Packages [522 kB]
Get:18 http://archive.ubuntu.com/ubuntu jammy-security/main Translation-en [112 kB]
[...]

Issue #3699, opened 2022-12-06
(Few)debianutils programs not installed on jammy64.


lakshayrohila:
I encountered with this error today when I tried to install `ksh` through `apt`.

It said that a program named `add-shell` was not found. So the post-configuration was not completed.

Again, when I tried to uninstall `ksh`, it reported error that `remove-shell` was not found.

Both are part of `debianutils`.

According to https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-distro/x86_64/ubuntu/jammy64/DISTRO_PKGS_SPECS-ubuntu-jammy#L98, `debianutils` **is** going to be installed, which it reports to be as well.

Running `apt install --reinstall debianutils` resolves the problem, with some other configurations so that apt does not continue with uninstalling/installing  `ksh` (or other shells).

lakshayrohila:
Investigating a bit more into the problem, I found this message when `apt install --reinstall debianutils` is run without running `apt update` at first. Here is the message:-
```
$ apt install --reinstall debianutils
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Reinstallation of debianutils is not possible, it cannot be downloaded.
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
```

lakshayrohila:
I ran a build yesterday (woof-CE was at commit 54495087bac668daa0abdb7a1308e0724b76f359 " add hardinfo petbuild"): [link to the github build](https://github.com/lakshayrohila/woof-CE/actions/runs/3929788799/jobs/6719016420).

I downloaded the log, wrote a script which prints grepped log messages in tabular form, and ran that script on the log. Here is the result by grepping `debianutils`:
```
----------------------------------------------------------------------------------------------------------------------------------
| ./0setup                                                                                                                       |
----------------------------------------------------------------------------------------------------------------------------------
| 6722 | abiword-deps acl adduser alsa-lib libasound2-plugins-pulseonly alsa-utils atk at-spi2-atk attr audit autoconf automake  |
|      | autopoint base-files bash bash-completion bbe bc binutils bison blueman bluez bzip2 breeze-cursor-theme brightnessctl c |
|      | a-certificates cairo ccache connman coreutils cmake cpio cpp crda curl cyrus-sasl2 dash dbus dbus-user-session d-conf d |
[...]

PR #3702, opened 2022-12-06
Possibly fix https://github.com/puppylinux-woof-CE/woof-CE/issues/3701.

lakshayrohila:


dimkr:
Why not just let the user install these extra packages if needed?

lakshayrohila:
I need a bit more time to investigate into this problem and https://github.com/puppylinux-woof-CE/woof-CE/issues/3699, which seem to bee related. Will tell you about this thing later on. Currently converting this PR as a draft.

lakshayrohila:
Closing the PR for now until I get any idea of what needs to be done. Haven't actually tested this stuff yet.

PR #3589, opened 2022-11-07
Add proper support for Devuan -updates and -security repos

dimkr:
Requested by josejp2424

PR #3704, opened 2022-12-06
Fix #3699

### LINKED ISSUE #3699: (Few)debianutils programs not installed on jammy64.

lakshayrohila:
This PR tries to fix https://github.com/puppylinux-woof-CE/woof-CE/issues/3699. I am currently running a github action to see if this works or not. Till now, 1download has actually downloaded `debianutils`. So maybe this will resolve the error.

### DIFF
--- a/woof-distro/x86_64/ubuntu/jammy64/DISTRO_COMPAT_REPOS-ubuntu-jammy
+++ b/woof-distro/x86_64/ubuntu/jammy64/DISTRO_COMPAT_REPOS-ubuntu-jammy
@@ -28,12 +28,12 @@ PKG_DOCS_DISTRO_COMPAT="
 z|http://archive.ubuntu.com/ubuntu/dists/${DISTRO_COMPAT_VERSION}/main/binary-${DBIN_ARCH}/Packages.${DDB_COMP}|Packages-${DISTRO_BINARY_COMPAT}-${DISTRO_COMPAT_VERSION}-main
 z|http://archive.ubuntu.com/ubuntu/dists/${DISTRO_COMPAT_VERSION}/universe/binary-${DBIN_ARCH}/Packages.${DDB_COMP}|Packages-${DISTRO_BINARY_COMPAT}-${DISTRO_COMPAT_VERSION}-universe
 z|http://archive.ubuntu.com/ubuntu/dists/${DISTRO_COMPAT_VERSION}/multiverse/binary-${DBIN_ARCH}/Packages.${DDB_COMP}|Packages-${DISTRO_BINARY_COMPAT}-${DISTRO_COMPAT_VERSION}-multiverse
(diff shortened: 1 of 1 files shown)


dimkr:
Maybe automate this by changing https://github.com/puppylinux-woof-CE/woof-CE/blob/262b4dae5b46f193753a57d8767fa6483a4748f9/woof-code/0setup#L166?

(#3589 is a good example)

lakshayrohila:
> Maybe automate this by changing
> 
> https://github.com/puppylinux-woof-CE/woof-CE/blob/262b4dae5b46f193753a57d8767fa6483a4748f9/woof-code/0setup#L166
> ?
> 
> (#3589 is a good example)

I'll see to it. First testing if this works. Maybe doing the automation in `woof-distro/x86_64/ubuntu/jammy64/DISTRO_COMPAT_REPOS-ubuntu-jammy` is better. I mean, do all ubuntu based distros' packages follow the same rule (we need to be sure about that)?

dimkr:
> First testing if this works

It won't work, woof-CE doesn't know how to deal with a package that exists in repo A and in repo B (hence the stable version sort+uniq in https://github.com/puppylinux-woof-CE/woof-CE/blob/262b4dae5b46f193753a57d8767fa6483a4748f9/woof-code/0setup#L319).

lakshayrohila:
> It won't work, woof-CE doesn't know how to deal with a package that exists in repo A and in repo B (hence the stable version sort+uniq in
> 
> https://github.com/puppylinux-woof-CE/woof-CE/blob/262b4dae5b46f193753a57d8767fa6483a4748f9/woof-code/0setup#L319
> ).

Remove updates repo then?

dimkr:
woof-CE doesn't know how to deal with the -updates and -security repos, unless it's done the way it's done by 0setup. If package A is version 1.2.3 and updates has 1.2.3-1 and security has 1.2.3-2, 0setup won't pick the right package unless the three repos are combined into one package list, with one version (= the highest) for each package.

lakshayrohila:
@dimkr, I think this currently goes over my head, so closing this draft. + is this the real issue for debianutils not being installed, or do you think there must be something else behind it?

dimkr:
> do you think there must be something else behind it?

I don't have a quick answer, but I do know that we don't support /etc/shells and many things in Puppy assume that /bin/sh is bash.

PR #3708, opened 2022-12-06
Fix ptheme_gtk not showing up after one theme change

dimkr:
Very fun bug that left me puzzled because it seemed so random. Now I realize it worked _sometimes_ because sometimes I started ptheme_gtk via the JWM menu.

To reproduce the bug:

1. Run ptheme_gtk **through the pinboard icon** (setup -> Desktop -> Desktop Settings -> Theming)
2. Change the GTK+ theme
3. Repeat (1)

Nothing will happen, because ptheme_gtk re-runs ROX-Filer while the THEME environment variable is set. Then, when ROX-Filer runs ptheme_gtk (via pdesktop, etc'), this environment variable is inherited multiple times until it reaches ptheme_gtk, and ptheme_gtk sets the theme instead of showing a dialog when THEME is set.

If I understand correctly, this bug is present since 6f3e8b7 (2015) 🎂

### DIFF
--- a/woof-code/rootfs-packages/ptheme/usr/sbin/ptheme_gtk
+++ b/woof-code/rootfs-packages/ptheme/usr/sbin/ptheme_gtk
@@ -200,6 +200,7 @@ done
 
 if [ "$THEME" ]; then
 	set_theme_gtk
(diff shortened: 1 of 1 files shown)


lakshayrohila (approved):
I come across the same bug sometimes. I think this would fix the issue.

PR #3715, opened 2022-12-07
Silence makepkgdb.sh errors

dimkr:


peabee (changes requested):
I don't think its just a question of silencing - not sure the new /var/db/xbps/pkgdb-0.38.plist is being generated.....

dimkr:
This PR won't make the problem worse, if it doesn't solve it.

peabee:
Maybe better to keep the errors until the cause can be determined.........

makepkgdb runs ok, if slowly, locally - how to diagnose when it is running on github???

dimkr:
> makepkgdb runs ok, if slowly, locally - how to diagnose when it is running on github???


Run it inside an Ubuntu container or something. I'd add `set -x`, to see which lines produce these errors.

If you choose to debug on GitHub Actions, please do this in a personal fork of woof-CE, because every build eats a portion of allowed build minutes and storage.

peabee:
> Run it inside an Ubuntu container or something.

No idea how to do this :-((

dimkr:
https://podman.io/getting-started/ is a good place to start if you've never used containers before.

peabee:
Have experimented running makepkgdb locally on other upups - fossapup64 and bionicpup64 - and in both cases no errors appear just like with Slackware...... beginning to suspect its some sort of resource limitation on github??

Strangely, the void pkg database seems to be generated ok despite the error indications.

Google shows a few hits for "tail: error writing 'standard output': Broken pipe" but nothing of significance.

PR #3716, opened 2022-12-08
Copy extra petbuild files after the _DOC split

dimkr:


lakshayrohila (approved):

dimkr:
Closing, this needs to be re-implemented. The copying to `packages-${DISTRO_FILE_PREFIX}` needs to happen even if the build output is cached.

Only this block needs to be moved:

```
    for EXTRAFILE in ../rootfs-petbuilds/${NAME}/*; do
        case "${EXTRAFILE##*/}" in
        petbuild|*.patch|sha256.sum|*-*|DOTconfig|*.c|*.h) ;;
        *) cp -a $EXTRAFILE ../petbuild-output/${NAME}/
        esac
    done
```

PR #3718, opened 2022-12-08
Drop evince and use qpdfview instead in jammy64.

lakshayrohila:
Few icons are not found for evince to work + qpdfview is smaller in size (by few MBs). Hence I want to change evince with qpdfview.


dimkr:
qpdfview itself may be smaller but it pulls in Qt. Are you sure you want to do this?

lakshayrohila:
> qpdfview itself may be smaller but it pulls in Qt. Are you sure you want to do this?

Yes, unless we get the icons working in evince. I'll search more about it today.

lakshayrohila:
@dimkr, I think it is great to do so. It adds just +5Mbs (total of 483MBs) in iso, but as you said, also brings in Qt. And I think that at some point of time the user will need to install an application which requires Qt.

lakshayrohila:
I think that it is great to do so. It adds just +5Mbs (total of 483MBs) in iso, but as you said, also brings in Qt. And I think that at some point of time the user will need to install an application which requires Qt.

dimkr:
> And I think that at some point of time the user will need to install an application which requires Qt.

I think the size of Qt is a big price to pay for a PDF viewer. I've been Qt-free for years, and most modern browsers have a built-in PDF viewer.

But, although I wouldn't do this, I have no objections to this PR. Let me know if you want me to merge it.

lakshayrohila:
> > And I think that at some point of time the user will need to install an application which requires Qt.
> 
> I think the size of Qt is a big price to pay for a PDF viewer. I've been Qt-free for years, and most modern browsers have a built-in PDF viewer.
> 
> But, although I wouldn't do this, I have no objections to this PR. Let me know if you want me to merge it.

Well, that's a great option as well. If palemoon supports viewing pdf, I can even set defaultpdfviewer to that! I'll take a look if it does or not, while most browsers do.

lakshayrohila:
@dimkr, I see that the not found icons on evince was just a problem of set gtk icon theme. Earlier, when it was `Pmaterial`, the icons would not be found. But now as I installed `Breeze` icon theme (using `apt`) the icons are found. What do you think we should do in this case? Switch to a non-puppy theme?

dimkr:
Icon themes are huge. Does the problem reproduce with Pmaterial after https://github.com/puppylinux-woof-CE/Pmaterial/commit/03426bff98e7ccb8fe4d431dcad139fdc43c9624?

lakshayrohila:
> Icon themes are huge. Does the problem reproduce with Pmaterial after [puppylinux-woof-CE/Pmaterial@03426bf](https://github.com/puppylinux-woof-CE/Pmaterial/commit/03426bff98e7ccb8fe4d431dcad139fdc43c9624)?

I'll check and tell. I have a very latest build (built today) sitting at my desk 👍.

dimkr:
Make sure you delete petbuild-sources/pmaterial_icons/pmaterial_icons.tar.xz before 3builddistro if you want to ensure you have the fix, or just change /usr/share/icons/PMaterial/index.theme manually.

PR #3726, opened 2022-12-10
Show message in PPM to notify user to use apt when bdrv is present.

lakshayrohila:
@dimkr, this is not the complete doc, though, it is a useful thing to be added in PPM.

### DIFF
--- a/woof-code/support/bdrv.sh
+++ b/woof-code/support/bdrv.sh
@@ -221,7 +221,7 @@ if [ -f "$PPM_SRC_LOCATION" ]; then
 	cat > $PPM_TARGET_LOCATION <<< $(echo -n "$PPM_FIRST_LINES"'
 
 # Support for apt is built. Explain the user to use apt instead of PPM (PPM is outdated).
-. /usr/lib/gtkdialog/box_ok "APT FOUND!" info "Note that apt is found on this system. So to install any program, use apt instead."
+. /usr/lib/gtkdialog/box_ok "APT FOUND!" info "Note that apt is found on this system. So to install any program, except for puppy specific programs, use apt instead."
 #----------------------------------------------------------------------------------------
 ' && tail -n+5 $PPM_SRC_LOCATION)
 
--- a/woof-code/support/bdrv.sh
+++ b/woof-code/support/bdrv.sh
@@ -208,3 +208,24 @@ if [ -d bdrv/usr/share/gnome/help ]; then
 	mv bdrv/usr/share/gnome/help bdrv_DOC/usr/share/gnome/
 	rmdir bdrv/usr/share/gnome 2>/dev/null
 fi
+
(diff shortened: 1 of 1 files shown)


woof-code/support/bdrv.sh
```diff
@@ -208,3 +208,24 @@ if [ -d bdrv/usr/share/gnome/help ]; then
...
+	PPM_FIRST_LINES=`head -n5 $PPM_SRC_LOCATION` # to preserve 'comments + #!/...' before adding this doc.
+
+	cat > $PPM_TARGET_LOCATION <<< $(echo -n "$PPM_FIRST_LINES"'
+
+# Support for apt is built. Explain the user to use apt instead of PPM (PPM is outdated).
+. /usr/lib/gtkdialog/box_ok "APT FOUND!" info "Note that apt is found on this system. So to install any program, use apt instead."
+#----------------------------------------------------------------------------------------
+' && tail -n+5 $PPM_SRC_LOCATION)
+
+	chmod +x $PPM_TARGET_LOCATION
```
lakshayrohila, line 228: @dimkr, just wanted to know if I needed to remove source ppm file after this line.

dimkr (changes requested):
Don't do this. Support for .deb packages in PPM is disabled when apt is present, and PPM only deals with .pet packages. No reason to prevent the user from using it.

lakshayrohila:
> Don't do this. Support for .deb packages in PPM is disabled when apt is present, and PPM only deals with .pet packages. No reason to prevent the user from using it.

But they may not know whether PPM is there for show only or is it able to actually install something. I am updating the content a bit and hope you think it is be better now.

dimkr:
I don't like this PR. I prefer things like #3606 and #3575 to a popup. I prefer to lead users to Synaptic, so users who don't use .pet packages don't ever use PPM and don't need to wonder why PPM is so limited.

lakshayrohila:
> I don't like this PR. I prefer things like #3606 and #3575 to a popup. I prefer to lead users to Synaptic, so users who don't use .pet packages don't ever use PPM and don't need to wonder why PPM is so limited.

Let's think about someone like me. If I didn't know PPM was there, I would rather not focus on the restrictions. But PPM is not completely removed, and few applications all link to PPM, not apt or synaptic. So, what I'd think is that PPM is a package manager built especially for this distro only, atleast PPM means **Puppy** Package Manager. So, I will try running it. But, finding no package to be installed which I know about, I'd think it is a bug. This popup will only allow the users to know that apt is to be used but does not forbid them from using PPM. I'll also add a doc file for this, which will explain to the user everything. So, I am currently closing this until I create that doc, but want to know if your mind has changed?

Thanks.

dimkr:
> few applications all link to PPM

Let's fix this first, then re-evaluate. I don't want a popup to be a substitute for fixing the things that tell users they should use PPM.

PR #3729, opened 2022-12-11
Change to xlockmore petbuild

peabee:


dimkr:
Next time you should squash merge PRs like this, so they become one commit after the merge 💡

peabee:
> Next time you should squash merge PRs like this, so they become one commit after the merge bulb

I can't see how to do this on the web interface to Github?? Google search says click "History" but I see no such.

dimkr:
image

PR #3733, opened 2022-12-12
Make missingdefaultapp to point the user for using Synaptic instead of PPM if bdrv is present.

lakshayrohila:
Currently `woof-code/rootfs-skeleton/usr/local/bin/missingdefaultapp` points the user to use PPM even if bdrv is present. This PR solves the problem.

### DIFF
--- a/woof-code/rootfs-skeleton/usr/local/bin/missingdefaultapp
+++ b/woof-code/rootfs-skeleton/usr/local/bin/missingdefaultapp
@@ -1,5 +1,8 @@
 #!/bin/sh
 
+# remember to add a .patch file at woof-code/support/bdrv_support/missingdefaultapp.patch when making any changes to this file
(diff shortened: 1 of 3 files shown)


dimkr:
Duplicate of #3575?

peabee:
Why only synaptic/bdrv? What about xbps/vpm if that is present? or any other distro's pm mechanism.....

dimkr:
It should be super easy to add a second `if` to #3575 😸

peabee:
We probably need some symbols defined in build.conf to say what is the "preferred" PM and what is the "alternative" PM?

dimkr:
PPM is never the correct answer, so I don't see a reason to make this configurable.

lakshayrohila:
> Duplicate of #3575?

Ah! Didn't see that. Sorry 😅!

PR #3735, opened 2022-12-13
.xinitrc: Properly set XDG Desktop variables

rizalmart:


woof-code/rootfs-skeleton/root/.xinitrc
```diff
@@ -72,6 +72,24 @@ elif [ -e /usr/bin/pulseaudio ]; then
...
+*gnome*) DE2=GNOME; DE=gnome ;;
+*lxqt*) DE2=LXQT; DE=lxqt ;;
+*kde*) DE2=KDE; DE=kde ;;
+*mate*) DE2=MATE; DE=mate ;;
+*cinnamon*) DE2=Cinnamon; DE=cinnamon ;;
+*deepin*) DE2=deepin; DE=deepin ;;
+*) DE2="X-Generic"; DE=generic ;;
+esac
+
+export XDG_CURRENT_DESKTOP="$DE2"
```
dimkr: I think it would be a better idea to do this via /etc/environment or /etc/profile, because some graphical applications are not children (or children of children of ...) of xinit. (The dwl petbuild sets XDG_CURRENT_DESKTOP via /etc/environment.)

rizalmart: I just make it adaptive by the way

Issue #3736, opened 2022-12-13
Make ptheme_gtk show example window instantaneously.


lakshayrohila:
I request to add, if possible, a feature which allows ptheme_gtk to show an example window instantaneously, maybe after just a change in selected item or maybe after click of some button. It is frustrating to test all the given themes because we have to open it again and again for testing them.

lakshayrohila:
I am currently working on the feature request, so no one else needs to.

PR #3737, opened 2022-12-13
Make ptheme_gtk window to appear even after -t is used (add option -T)

lakshayrohila:
This completes https://github.com/puppylinux-woof-CE/woof-CE/issues/3736.

When using this application the user might find it frustrating to re-enter the command again and again even if they just want to test the themes. This PR allows the window to appear, as if window getting reloading (but a bit of lag in between :grin:; of course the window has to close and then reopen). This might remove their source of frustration (at least mine).

### DIFF
--- a/woof-code/rootfs-packages/ptheme/usr/sbin/ptheme_gtk
+++ b/woof-code/rootfs-packages/ptheme/usr/sbin/ptheme_gtk
@@ -181,11 +181,11 @@ while [ $# != 0 ]; do
 			-t) export THEME="$2"
 				[ ! "$THEME" ] && export THEME="$ACTIVE_THEME"
 				shift
-				export EXIT_WITHOUT_GUI=1;;
+				EXIT_WITHOUT_GUI=1;;
 			-T) export THEME="$2"
 				[ ! "$THEME" ] && export THEME="$ACTIVE_THEME"
 				shift
-				export EXIT_WITHOUT_GUI=0;;
+				EXIT_WITHOUT_GUI=0;;
 			-f) export FONT="$2"; shift;;
 			-i) export ICON_THEME="$2"
 				shift;;
@@ -219,11 +219,8 @@ if [ "$THEME" ]; then
 	fi
 
 	if [ $EXIT_WITHOUT_GUI -eq 1 ]; then
-		unset EXIT_WITHOUT_GUI
 		exit
 	fi
-
-	unset EXIT_WITHOUT_GUI
 fi
 
 #build list
--- a/woof-code/rootfs-packages/ptheme/usr/sbin/ptheme_gtk
+++ b/woof-code/rootfs-packages/ptheme/usr/sbin/ptheme_gtk
@@ -173,6 +173,7 @@ set_font (){
 
 export -f set_theme_gtk set_font set_gtk3_theme set_gtk4_theme
 
+EXIT_WITHOUT_GUI=0
 #parameters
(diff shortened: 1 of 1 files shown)


woof-code/rootfs-packages/ptheme/usr/sbin/ptheme_gtk
```diff
@@ -180,7 +180,12 @@ while [ $# != 0 ]; do
 		case $1 in
 			-t) export THEME="$2"
 				[ ! "$THEME" ] && export THEME="$ACTIVE_THEME"
-				shift;;
+				shift
+				export EXIT_WITHOUT_GUI=1;;
```
dimkr: No need for `export`, if I understand correctly a variable (= not an environment variable) should work just fine

lakshayrohila:
@peabee, your thoughts on this PR?

woof-code/rootfs-packages/ptheme/usr/sbin/ptheme_gtk
```diff
@@ -180,7 +180,12 @@ while [ $# != 0 ]; do
 		case $1 in
 			-t) export THEME="$2"
 				[ ! "$THEME" ] && export THEME="$ACTIVE_THEME"
-				shift;;
+				shift
+				EXIT_WITHOUT_GUI=1;;
+			-T) export THEME="$2"
+				[ ! "$THEME" ] && export THEME="$ACTIVE_THEME"
+				shift
+				EXIT_WITHOUT_GUI=0;;
```
dimkr, line 188: I think you should move this to line 176, so `EXIT_WITHOUT_GUI=0` even if neither -t nor -T is specified

woof-code/rootfs-packages/ptheme/usr/sbin/ptheme_gtk
```diff
@@ -180,7 +180,12 @@ while [ $# != 0 ]; do
 		case $1 in
 			-t) export THEME="$2"
 				[ ! "$THEME" ] && export THEME="$ACTIVE_THEME"
-				shift;;
+				shift
+				EXIT_WITHOUT_GUI=1;;
```
dimkr, line 185: I think you should move this to line 176, so `EXIT_WITHOUT_GUI` is defined even if neither -t nor -T is specified

lakshayrohila: It would never be needed then, since `THEME` would not be set, I suppose. Still correcting this, it does not hurt to do so 👍.

dimkr (approved):
If it works, IMO this makes ptheme_gtk less annoying to use

lakshayrohila:
> If it works, IMO this makes ptheme_gtk less annoying to use

I think it must work. Out of nearly 20 times, it worked once on an already jammy64. Maybe the reason is that I am using an old jammy64 (built nearly a fortnight ago), in which ptheme_gtk had a problem of failing to start 70% (just a rough figure) of times. So maybe the window doesn't popup on my device because the application has to successfully start twice.

If you have an OS with ptheme built-in, I think you could then test it more easily. I am trying to confirm if this works from my side as well.

**Edit:** Just got to remember that I had some newer build of jammy64 available with me. Tried using it on that OS. Works 100% of the times, except when I ran it the first time (the window didn't pop-up, nor did it set any theme. I ran `ptheme_gtk`, with no -T option.). Don't know the reason, but after it happened, it worked every time.

PR #3748, opened 2022-12-15
Add error message about cups not being installed to cups_shell.

lakshayrohila:
If `cups_shell` is started through shell, it will try running `/etc/init.d/cups` even if it is not there.


lakshayrohila:
@dimkr, please review.

PR #3752, opened 2022-12-16
Re-run ldconfig after ld.so.conf is changed (fixes #3751)

### LINKED ISSUE #3751: VoidPup kernel panics
Both VoidPup64 and VoidPup32 built from testing branch fail to boot - kernel panic.

Can chroot into a writeable copy of puppy_vpup64_22.12.sfs but leafpad gives multiple gdk-pixbuf errors.

initrd.gz is much bigger than local build - 2.3MB vs 1.8MB

dimkr:


peabee:
Many thanks.

PR #3765, opened 2022-12-21
Default to COPY2RAM=yes based on free RAM, not total RAM

dimkr:
Puppy used to be much smaller, and a modern Puppy no longer runs comfortably with 1 GB of RAM unless `pfix=nocopy`. It's just too big to fit in RAM and leave enough space for things like installing a browser in a live session, without constant swapping.

Right now, a SFS is copied to RAM if a computer has more than 1 GB of RAM **and** the amount of RAM is at least ~twice the SFS size. This condition is still true with 1 GB of RAM, although today's Puppy is so much bigger than it used to be around 2010, leaving the user with very little usable RAM by today's standards.

I'm proposing a change to the current logic: do the same check but with free ramdisk space instead. The amount of free RAM, not total RAM, is the way to predict bad user experience with the SFSs copied to RAM. Users who insist on copying to RAM despite the low amount of free RAM can always use `pfix=copy`, as today.

### DIFF
--- a/initrd-progs/0initrd/init
+++ b/initrd-progs/0initrd/init
@@ -394,7 +394,8 @@ setup_onepupdrv() {
   MINRAM2CPY=$(($SIZESFSK * 2)) #100222 technosaurus: in case of very big puppies.
   #decide whether to copy .sfs's to ram
   if [ $RAMSIZE -gt 1048576 ] ; then
-   [ $RAMSIZE -gt $MINRAM2CPY ] && COPY2RAM="yes"
+   TFREEK=$(df | grep -m1 ' /mnt/tmpfs' | tr -s ' ' | cut -f 4 -d ' ')
+   [ $TFREEK -gt $MINRAM2CPY ] && COPY2RAM="yes"
   fi
  fi
  ONE_LOOP="$(losetup -f)"


peabee (commented):
I don't have this problem - nor do I see any reports of it being a widespread problem amongst Puppy users?
I think a change like this should be tested by providing a test build on the forum so that people with resource limited pc's can test and assess the impact.
No objections in principle though.

dimkr:
Have you tried to use any modern Puppy with 1 GB of RAM?

peabee:
No! Thats why other testers are needed.

dimkr:
Why don't you try VoidPup or S15Pup on a machine with 1 GB of RAM and see for yourself?

peabee:
Don't have such a pc............

dimkr:
You can try a virtual machine with 1 GB of RAM. It's painfully slow even with plenty of CPU cores, and things start to work reasonably well with `pfix=copy` at 1.5-2 GB of RAM.

bigpup1:
What point swap is set to start being used is something to consider.
https://oldforum.puppylinux.com/viewtopic.php?t=100219
That had an affect when most computers were still using 1GB or less RAM.
Swap was set to kick in too soon.

The Puppy boot process, in the past, did consider RAM size and auto-adjust how much got loaded into RAM, based on what Ram is in the computer.
Not sure if that is still in the boot process code.

dimkr:
> based on what Ram is in the computer.

This is exactly the piece of code changed by this PR. The code uses available RAM and not free RAM to determine if it's a good idea to copy SFSs. It's a fallacy to copy the main SFS to RAM if you have 1 GB of RAM while the SFS is 900 MB, leaving only 100 MB of free RAM.

mavrothal:
I would say that the change is not going far enough!
The original code was from the time that puppy had one sfs and all the apps already in it.
Today is not the case anymore. So not only I would go with the propose change but I would also increase the MINRAM2CPY above, to 3 or even 4 ie ```MINRAM2CPY=$(($SIZESFSK * 3))``` to allow some more RAM space for the apps loaded from the savefile.
BTW I do have an old machine with 768 MB RAM and for quite some time now whenever I run puppies I do it with ```nocopy```, even if the SFSs fit in RAM,  because they behave better.
Happy New Year everyone.

01micko:
I tend to agree with @mavrothal 

Case in point - [spup-100](https://oldforum.puppylinux.com/viewtopic.php?t=62750) from almost __12__ years ago was 117MB. Any slacko/spup/scpup now days - with the same apps included - will be at least 4X that size.

dimkr:
> I would say that the change is not going far enough!

I agree, the right thing to do is to check free space and to do that after each SFS, increase that safety margin and so on.

However, I'm also fine with a minimal change like the one proposed here because the main SFS size is multiplied by 2 and the main SFS is more than twice the size of all other SFSs combined. The old formula still works in 80% of cases when it looks at the right number.

01micko:
Ooh - posted almost exactly the same time @dimkr :four_leaf_clover:

dimkr:
It's a miracle!

01micko:
:rofl:

01micko (approved):

dimkr:
Opened #3777, a counter-proposal that makes more sense to me compared to the `size * 2 + 1000` heuristic we have today. Bigger change, though.

PR #3769, opened 2022-12-22
Enable more petbuilds in jammy64

dimkr:
More use of petbuilds = less duplicate effort and old, buggy packages 😸 

@lakshayrohila

### DIFF
--- a/woof-distro/x86_64/ubuntu/jammy64/_00build.conf
+++ b/woof-distro/x86_64/ubuntu/jammy64/_00build.conf
@@ -41,7 +41,7 @@ USR_SYMLINKS=yes
 BUILD_BDRV=yes
 
 ## packages to build from source
-PETBUILDS="busybox aaa_pup_c disktype dmz-cursor-theme firewallstatus freememapplet geany gexec gpicview gtkdialog gtk_theme_flat_grey_rounded gtk_theme_polished_blue gtk_theme_gradient_grey gtk_theme_buntoo_ambience gtk_theme_stark_blueish gxmessage jwm l3afpad lxtask lxterminal mtpaint pa-applet powerapplet_tray rox-filer claws-mail transmission xarchiver xcur2png xdelta xdg-puppy-jwm Xdialog xlockmore yad pmaterial_icons puppy_standard_icons puppy_flat_icons ram-saver connman-puppy connman-ui connman-gtk fixmenusd spot-pkexec notification-daemon-stub palemoon weechat"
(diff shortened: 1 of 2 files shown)


woof-distro/x86_64/ubuntu/jammy64/_00build.conf
```diff
@@ -41,7 +41,7 @@ USR_SYMLINKS=yes
...
+PETBUILDS="busybox aaa_pup_c disktype dmz-cursor-theme firewallstatus freememapplet geany gexec gpicview gtkdialog gtk_theme_flat_grey_rounded gtk_theme_polished_blue gtk_theme_gradient_grey gtk_theme_buntoo_ambience gtk_theme_stark_blueish gxmessage jwm l3afpad lxtask lxterminal mtpaint pa-applet powerapplet_tray rox-filer claws-mail transmission xarchiver xcur2png xdelta xdg-puppy-jwm Xdialog xlockmore yad pmaterial_icons puppy_standard_icons puppy_flat_icons ram-saver connman-puppy connman-ui connman-gtk fixmenusd spot-pkexec notification-daemon-stub palemoon weechat osmo abiword gnumeric xournalpp mpv efilinux"
```
lakshayrohila, line 44: Does this osmo not yield segfault (the earlier one caused, so please test it)?

dimkr: Started just fine

dimkr: image

lakshayrohila:
I also was thinking about abiword and gnumeric and other lightweight document viewer stuff. I mean, is this really needed? Not compelling, but just putting my ideas forward. I personally don't use gnumeric but use libreoffice, which we can allow the user to install themselves (in case they ever need). I feel like these document viewers sometimes are quite big and hence might be a burden for some users. I don't know particularly how much space do these use, so if small in size then it can be great as well.

woof-distro/x86_64/ubuntu/jammy64/_00build.conf
lakshayrohila: Its great to have it then!

dimkr:
I don't use them myself either, but many Puppy users expect them to be there. They're pretty small, because they're built without optional dependencies. In particular, mpv is very small compared to the Ubuntu package plus dependencies. The ISO size is increased by just 5 MB, that's very little for so much functionality.

lakshayrohila:
> I don't use them myself either, but many Puppy users expect them to be there. They're pretty small, because they're built without optional dependencies. In particular, mpv is very small compared to the Ubuntu package plus dependencies. The ISO size is increased by just 5 MB, that's very little for so much functionality.

That's great! Just if all of it works then this PR is good to be merged. Thanks!

lakshayrohila (approved):

PR #3771, opened 2022-12-24
Add gparted petbuild to drop the polkit dependency

dimkr:
@lakshayrohila Objections? This PR reduces size by dropping optional dependencies.


lakshayrohila:
I have very little information about polkit, so have no say on this. I'd say, if it works then just go for it!

PR #3598, opened 2022-11-10
Update documentation pages

lakshayrohila:
This PR tries to update few documentation pages, according to [this](https://forum.puppylinux.com/viewtopic.php?t=7273) puppy forum post. Currently only these pages are updated (though there are few links to oldforums, still as those links are working, and I cannot find any similar content in newer forum, I am leaving those links as is):-

1. `devx.htm`
2. `HOWTO-microsoft.htm`
3.  `HOWTO-regexps.htm` (Deleted, not of use to puppy)
4.  `forums.htm`
5.  `forums.css` (Deleted, currently of no use)
6.  `gtkdialog.htm`
7.  `HOWTO_Internet.htm` (More links added)
8.  `release-skeleton.bottom.htm`
9.  `HOWTO_dialup.htm` (Renamed from `HOWTO_modem.htm` and links set to proper state, no content updated)
10.  `installation.htm` (Latest one from **bigpup**)
11. `root.htm`
12.  `index.html.top` (More links added)
13.  `home.htm` (By **bigpup**)
14.  `cups_shell.htm` (By **bigpup**)

`manual.css` is also updated to change link colors. `pmusic.jpg` is added for use in `gtkdialog.htm`. `index.html.top` is changed a bit. `sns_imgs` directory is also added, containing images for `HOWTO_sns.htm`. Also, a **scroll-to-top** button is added to each file.

Thanks to [williwaw](https://forum.puppylinux.com/memberlist.php?mode=viewprofile&u=1096) and [bigpup](https://forum.puppylinux.com/memberlist.php?mode=viewprofile&u=1095) for their earnest efforts made to guide me making these changes available.

Documentation for `sns` and `connman-gtk` petbuild is also added.

[This](github.com/lakshayrohila/woof-CE/releases/tag/jammypup64-10-docs-test1) is a jammy64 built from the aforementioned changes.

PR #3777, opened 2022-12-26
Default to COPY2RAM=yes when the ramdisk has 512 MB of free space

dimkr:
A more aggressive implementation of what #3765 tries to achieve.

(Not tested yet; I want to get some feedback first.)

### DIFF
--- a/initrd-progs/0initrd/README.txt
+++ b/initrd-progs/0initrd/README.txt
@@ -231,7 +231,7 @@ pfix=<ram, nox, nocopy, fsck, fsckp, rdsh, <number>>
    ram:      run in ram only (do not load ${DISTRO_FILE_PREFIX}save).
    nox:      do not start X.
    copy:     copy .sfs files into ram
-   nocopy:   do not copy .sfs files into ram (default is copy if ram <= 1024 MB).
+   nocopy:   do not copy .sfs files into ram (default is copy while free ramdisk space is >= 512 MB).
    fsck:     do fsck of ${DISTRO_FILE_PREFIX}save.?fs file.
    fsckp:    do fsck before first mount of supported partitions.
    rdsh:     exit to shell in initial ramdisk.
--- a/initrd-progs/0initrd/init
+++ b/initrd-progs/0initrd/init
@@ -386,20 +386,10 @@ setup_onepupdrv() {
    COPY2RAM="no"
  fi
  #-
- if [ "$COPY2RAM" = "" ];then
-  COPY2RAM="no"
-  #if there's heaps of ram, copy puppy.sfs to a tmpfs...
-  SIZESFSK=$(du -k $ONE_FN | cut -f 1)
-  SIZESFSK=$(($SIZESFSK + 1000)) #some slack.
-  MINRAM2CPY=$(($SIZESFSK * 2)) #100222 technosaurus: in case of very big puppies.
-  #decide whether to copy .sfs's to ram
-  if [ $RAMSIZE -gt 1048576 ] ; then
-   [ $RAMSIZE -gt $MINRAM2CPY ] && COPY2RAM="yes"
-  fi
- fi
  ONE_LOOP="$(losetup -f)"
- if [ "$COPY2RAM" = "yes" ];then
+ if [ "$COPY2RAM" = "" -o "$COPY2RAM" = "yes" ];then
   SIZEZK=$(du -k $ONE_FN | cut -f 1)
+  [ "$COPY2RAM" = "" ] && SIZEZK=$(($SIZEZK + 524287)) #some slack.
   TFREEK=$(df | grep -m1 ' /mnt/tmpfs' | tr -s ' ' | cut -f 4 -d ' ')
   if [ $TFREEK -gt $SIZEZK ];then
    if [ "$ONE_MP" = "" ];then #101101 humongous initrd.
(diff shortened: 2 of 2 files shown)


01micko (commented):
Looks kike `COPY2RAM="no"`

dimkr:
@01micko Not exactly.

Before this PR:

* pfix=copy means "copy if ramdisk has enough space for this SFS"
* pfix=nocopy means "don't copy any SFS"
* Neither pfix=copy nor pfix=nocopy means "copy if we have 1 GB of RAM and `RAM > main SFS size * 2 + 1000K` and the ramdisk has enough space for this SFS"

With today's main SFS size, the third case is broken because 1 GB  of RAM is still enough to satisfy this condition, although you can't install things and run a browser.

With this PR:

* pfix=copy means "copy if ramdisk has enough space for this SFS"
* pfix=nocopy means "don't copy any SFS"
* Neither pfix=copy nor pfix=nocopy means "copy if ramdisk has enough space for this SFS **and extra 512 MB**" (softer version of pfix=copy really)

Assuming, 512 MB of free ramdisk space is enough for basic tasks.

dimkr:
(changed to `+ 524287` instead of `+ 524288`, because it's `-gt` and not `-ge`)

lakshayrohila:
I think 512MB free ram after copying the files is still great. At least we would notify the user about the `pfix=nocopy` once I complete [this](https://github.com/puppylinux-woof-CE/woof-CE/pull/3598) PR and it is merged.

01micko (approved):

lakshayrohila (approved):

dimkr:
> I think 512MB free ram after copying the files is still great.

Maybe it's too much?

If I remember correctly, the default tmpfs size is 50% of available RAM. This means that this PR lifts the minimum RAM for automatic copying of a 512 MB-sized Puppy from 1 GB to 2 GB.

I think this is reasonable, because 1 GB is way too little for pfix=copy nowadays and 1.25 or 1.5 GB of RAM are not common. Reducing to 256 MB won't do anything for most people, if I understand correctly.

dimkr:
Some test results with fresh builds:

1. dpup

```
12M	bdrv_dpup_9.3.sfs
20M	kbuild-5.10.149.sfs
52M	docx_dpup_9.3.sfs
56M	zdrv_dpup_9.3.sfs
74M	fdrv_dpup_9.3.sfs
174M	nlsx_dpup_9.3.sfs
357M	puppy_dpup_9.3.sfs
740M	total
```

1 GB: nothing is copied
1.25 GB: only fdrv and bdrv are copied
1.5 GB: everything except the main SFS and nlsx is copied
2 GB: everything except zdrv, docx and nlsx is copied
2.53 GB (a tiny bit more than the total SFS size plus 512 MB, doubled): everything is copied

2. vpup

```
26M	adrv_vpup64_1.2.3.sfs
47M	zdrv_vpup64_1.2.3.sfs
78M	fdrv_vpup64_1.2.3.sfs
249M	puppy_vpup64_1.2.3.sfs
398M	total
```

1 GB: nothing is copied
1.25 GB: only fdrv is copied
1.5 GB: everything except the main SFS is copied
2 GB: everything is copied

---

Yes, the ramdisk at / uses half the amount of installed RAM, so these numbers make sense: all SFSs are copied to the ramdisk only when `RAM >= 2 * total SFS size + 1 GB`, leaving >= 512 MB of free space in pup_rw after the copying.

dimkr:
If the main SFS is too big to copy but fdrv is small enough, it makes no sense to copy only fdrv to RAM. fdrv is used only rarely and it's a waste to copy it to RAM in a computer with little RAM (and not much space in the ramdisk).

I pushed d4b8e7d, which disables automatic copying of the next SFSs when the current one cannot be copied.

Now, only the main SFS and fdrv are copied to RAM when this dpup ISO is booted with 2 GB of RAM (357 MB + 74 MB + 512 MB of free space) and nothing is copied with 1.5 GB of RAM. (Behavior in Void should be similar: copying starts around 1.5-2 GB RAM, depending on the main SFS size.)

01micko (approved):

mavrothal:
If you are going for  rewrite, first I think one needs to know the total size of the SFSs to be loaded to decide for copying or not.
If not enough should prioritise. Fdrv and Docx are not used often enough to justify RAM loading for example
Second should decide on a sensible tempsf space rather that just "half" (another relic of the yesteryears that half was 500MG to 1GB at best).
Tempfs size should also take in account the PUPMODE as pumode 13 usually is happier with more tempfs space. 
In any case 2GB should be enough space for any puppy to run from RAM today specially with zram implementation across the board

dimkr:
> If not enough should prioritise. 

Working on that, it will be a separate PR. It splits setup_onepupdrv() into two functions: one that decides whether or not to copy and copies, and another one which mounts the SFS and adds it to the stack.

The copying is done in order of priority (main SFS, adrv, ydrv ... fdrv, zdrv, kbuild) while the stacking of layers is done in the same order as today.

> pumode 13 usually is happier with more tempfs space.

The init script already takes care of this and increases the tmpfs size to 75% when using PUPMODE 13 with overlay. The second init script (rc.sysinit) resizes pup_rw again, after zram is configured.

EDIT: done, see #3783

peabee (commented):
see #3765 comment

PR #3782, opened 2022-12-27
Add spot to sudoers.

lakshayrohila:
Before this PR:-
```
$ sudo cmd 
spot is not in the sudoers file.  This incident will be reported.
```

After this PR, everything works as expected.

### DIFF
--- a/woof-code/rootfs-skeleton/etc/sudoers
+++ b/woof-code/rootfs-skeleton/etc/sudoers
@@ -45,4 +45,4 @@ root	ALL=(ALL) ALL
 
 # want to run certain apps without asking for password...
 # %users ALL=(SHUTDOWN) NOPASSWD: ALL
-%users	ALL=(ALL) ALL, NOPASSWD: SHUTDOWN, SYSTEM
+%users,spot	ALL=(ALL) ALL, NOPASSWD: SHUTDOWN, SYSTEM
--- a/woof-code/rootfs-skeleton/pinstall.sh
+++ b/woof-code/rootfs-skeleton/pinstall.sh
@@ -91,4 +91,8 @@ exec '${DEFBROWSER}' "$@"
 	fi
 fi
 
+# set spot password
+chroot ./ /usr/bin/passwd spot << "woofwoof
+woofwoof" # passwd asks for entering password twice
+
 ### END ###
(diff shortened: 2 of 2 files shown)


dimkr:
What's the point if having the spot user if it can `sudo x` without a password to run `x` as root?

lakshayrohila:
> What's the point if having the spot user if it can `sudo x` without a password to run `x` as root?

Doesn't it have one? Okay, I remember now. I used `passwd spot` to make it have one. Any solution for that as well?

lakshayrohila:
@dimkr, please check if it would work now.

**Edit:** It won't. I'll take a look into it later on.

dimkr:
I think you misunderstood my question. This PR allows spot (or at least, that's the idea) to run a process as root without entering a password (NOPASSWD). This kind of ruins the idea of having an unprivileged user, because that user can do pretty much anything (= privileged).

lakshayrohila:
> I think you misunderstood my question. This PR allows spot (or at least, that's the idea) to run a process as root without entering a password (NOPASSWD). This kind of ruins the idea of having an unprivileged user, because that user can do pretty much anything (= privileged).

Not without a password, but allowing it to use sudo. Currently sudo won't allow spot to run anything.

Maybe I am misunderstanding the syntax of sudoers file?

dimkr:
If this PR works, it allows spot to run some commands as root without having to enter a password.

If you want to make spot capable of running any command as root, the spot user is useless from a security perspective.

If you want to make spot capable of running *some* commands (like some shell scripts from rootfs-skeleton) as root, they need a deep security review to ensure they can't be abused to run some other command or code from a location spot can write to (like /tmp). That would be a classic privilege escalation vulnerability.

lakshayrohila:
> If you want to make spot capable of running _some_ commands (like some shell scripts from rootfs-skeleton) as root, they need a deep security review to ensure they can't be abused to run some other command or code from a location spot can write to (like /tmp).

I actually want that spot runs all commands as user spot, unless sudo is used. Currently, you cannot use sudo as user spot because of the error I posted above. This adds spot to the sudoers file, giving it same permissions as any other user in the `users` group will have. The password for spot will still be asked, only when using sudo.

dimkr:
The idea behind the spot user is having a user who can't run commands as root - the inability to run sudo is a feature, not a bug 💡 

(And again, look closely - that NOPASSWD means that no password is needed)

PR #3783, opened 2022-12-27
If RAM is limited, copy only frequently-used SFSs to RAM

dimkr:
This is a step beyond #3777.

See https://github.com/puppylinux-woof-CE/woof-CE/pull/3777#issuecomment-1365275568.

This PR splits setup_onepupdrv() into two functions: copy_onepupdrv() and stack_oneppdrv(). The SFSs are copied to RAM, then stacked.

The decision whether or not to copy is smarter because it's done in order of importance: if RAM is limited, copying adrv to RAM will improve startup times of applications while copying fdrv to RAM is a waste of RAM.

Copying stops once one SFS cannot be copied to RAM, to avoid waste of RAM. For example, bdrv is tiny, but if RAM is so low the main SFS cannot be copied to RAM, no SFS should be copied to RAM, and definitely not rarely-accessed ones.

Example with dpup, running on a VM with **exactly** the amount of RAM needed for the main SFS, zdrv and 512 MB of free space in the ramdisk. bdrv is copied to RAM instead of zdrv, although zdrv is added to the stack earlier and although zdrv is bigger, because having bdrv in RAM results in increased responsiveness.

copy

(ignore the typo in the screenshot)

EDIT: reduced the requirement to 64 MB free

### DIFF
--- a/initrd-progs/0initrd/init
+++ b/initrd-progs/0initrd/init
@@ -32,7 +32,7 @@ L_ADDING_SAVE_LAYER_FAILED="adding %s to aufs stack failed." #printf
 L_WAITING_FOR_DEV='Waiting %s seconds for slow storage devices.' #printf
 L_WAITING_FOR_PART='Waiting for partition [%s]' #printf
 L_LOADING_KEYBOARD_LAYOUT="Loading '%s' keyboard layout..." #printf
-L_COPY_MESSAGE="copying to ram"
+L_COPY_MESSAGE="Copying '%s' to ram..." #printf
 L_LOAD_MESSAGE="main"
 L_PRE_MESSAGE="Loading puppy main sfs file."
 L_ERR_PDRV_INCOMPLETE="%s information is incomplete." #printf
@@ -354,7 +354,7 @@ find_onepupdrv() {
  [ "$ONE_FN" = "" -a "${2}" ] && echo "$ONE_PART, $ONE_TRY_FN file not found."
 }
 
-setup_onepupdrv() {
+stack_onepupdrv() {
 # "$1" - sfs spec - ex: sdb2,ext4,/pup/mydrv-1.2.3.sfs
 # "$2" - suffix for "pup_" branch directory - ex: a
 # "$3" - prepend after rw layer indicator - ex: p
@@ -367,7 +367,6 @@ setup_onepupdrv() {
  ONE_PREP="${3}"
  if [ "$ONE_PART" = "rootfs" ];then #humongous initrd.
   ONE_MP="" #actually it's '/'.
-  COPY2RAM='yes' #actually it is already in ram, but load_sfs_file code puts it in a tmpfs.
  else
   ONE_MP="$(mount | grep -m1 "/dev/$ONE_PART " | cut -f 3 -d ' ')"
   [ "$ONE_MP" ] || return 2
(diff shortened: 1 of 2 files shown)


peabee (commented):
see #3765 comment

dimkr:
@peabee Which comment?

mavrothal:
With this implementation dpup 10.0.131 needs 2908MB of RAM to load all 7 SFSs in RAM.
Looks excessive for an ~850MB ISO. 
At the same time it ends up with 2.1GB out of 2.9GB free in tmpfs and 1.4GB out of 1.4GB free in devtmpfs 
Might be better to check 'available' or 'free' RAM in the ```free``` output rather than /mnt/tmpfs

dimkr:
> Might be better to check 'available' or 'free' RAM in the free output rather than /mnt/tmpfs

I think the ramdisk size should stay at 50%, as today. If I increase the ramdisk size, this can enable copying to RAM in old computers, when they boot a barebones Puppy, and I prefer not to breaks things for some users: disable copying for users who are probably better off without it and improve the decision what SFSs get copied to RAM when it's limited, but without enabling copying for users who rely on the current behavior of no copying.

dpup is a bad example because it's the only Puppy with extras like docx and nlsx in the ISO. Without extras like docx, nlsx and kbuild, the size is not too far from the size of S15Pup and VoidPup. It's OK not to have the extra SFSs in RAM - English-speaking users might want to drop nlsx and most Puppy users don't boot it from a CD they want to eject.

I can decrease the requirement from 512 MB of free ramdisk space to 128 MB, so computers with 2 GB of RAM will copy all SFSs in a 850 MB ISO to RAM. However, I think that's too little, and users should decide if they want to explicitly enable copying when remaining ramdisk space is so limited. (128 MB should be OK thanks to zram swap, but PUPMODE 13 when pup_rw is compressed is not fun at all, on some computers.)

mavrothal:
>>Might be better to check 'available' or 'free' RAM in the free output rather than /mnt/tmpfs
>
>I think the ramdisk size should stay at 50%, as today. If I increase the ramdisk size, this can enable copying to RAM in old computers, when they boot a barebones Puppy, and I prefer not to breaks things for some users: 

The original init was looking at the total RAM
This init is looking at the available ramdisk space plus 512 MB
What I suggest is to look at the available RAM (as indicated by the ```free``` utility), plus 512MB to decide loading in RAM or not

dimkr:
> What I suggest is to look at the available RAM (as indicated by the free utility), plus 512MB to decide loading in RAM or not

Even if this means the ramdisk size needs to be increased from 50%?

mavrothal:
What is a scenario that this would happen?
Besides devtmpfs defaults to 50% of RAM if I understand correctly.

dimkr:
That 50% means "up to 50%", not "pre-allocate 50%".

I pushed 0c40410, which increases the ramdisk size limit to 100% of RAM, and copy_onepupdrv() takes care of leaving 512 MB free. Let's play with this and see how things behave.

01micko (approved):

dimkr:
@mavrothal 0c40410 is equivalent to checking free RAM instead of free ramdisk space. After quick testing I can say it **will** make things much worse for many users, I've reverted it. I restored the restriction of total SFS size to 50% of available RAM (as today) but dropped the free space requirement to 64 MB, because rc.sysinit increases ramdisk space.

mavrothal:
> @mavrothal https://github.com/puppylinux-woof-CE/woof-CE/commit/0c40410e0f1fb40a175d91416966f6ece1d297f5 is equivalent to checking free RAM instead of free ramdisk space

equivalent is not the same, but it doesn't matter. The 64MB slack works fine and loads everything with 2GB RAM.

mavrothal (approved):

dimkr:
The big improvement (and the main motivation behind this PR) is with 1 to 2 GB of RAM. With 1.25 GB, only the main SFS and the tiny bdrv are copied to RAM, while previously, copying of fdrv and zdrv was a big waste.

initrd-progs/0initrd/init
```diff
@@ -1030,42 +1044,55 @@ PUP_LAYER="$SFS_MP"
...
+#if possible, copy SFSs to RAM and prioritize frequently-accessed ones
+[ "$ADRV" ] && copy_onepupdrv "$ADRV"
+[ "$YDRV" ] && copy_onepupdrv "$YDRV"
+[ "$BDRV" ] && copy_onepupdrv "$BDRV"
+[ "$DEVX" ] && copy_onepupdrv "$DEVX"
+[ "$NLSX" ] && copy_onepupdrv "$NLSX"
+[ "$DOCX" ] && copy_onepupdrv "$DOCX"
+[ "$ZDRV" ] && copy_onepupdrv "$ZDRV"
+[ "$FDRV" ] && copy_onepupdrv "$FDRV"
+[ "$KBUILD" ] && copy_onepupdrv "$KBUILD"
```
dimkr, line 1083: @01micko @mavrothal Is this the right priority? adrv is top-priority, kbuild is used only rarely, zdrv and fdrv are mostly needed at boot time so they're unimporant.

I'm not sure if the order of ydrv, bdrv, devx, nlsx and docx makes sense:
- Locales on disk can slow down pretty much any application, including things like `cp`, but nlsx is very big.
- bdrv is tiny, but it may contain libraries or configuration files used by many applications, so it's a good investment of RAM.
- Many puppies are built with commonly used libraries in devx (gmp and mpfr are good examples), so if you install an application that depends on these libraries, you want devx in RAM.
- docx is relatively small but some users use it a lot: for example, I use man pages very often.
- ydrv is a wildcard, it can be big or small, depending on the user, but it probably should be in RAM because sometimes it's used to override parts of the main SFS.

dimkr:
(squashed into a single commit)

mavrothal:
The order looks pretty sensible to me
I would demote DOCX at the end as not so many read (...)
Maybe also put DEVX before BDRV but with BDRV being so small would rarely make a difference.

PR #3784, opened 2022-12-28
Keep symbolic icons used by Evince when minimizing Adwaita

dimkr:


lakshayrohila:
If jammy doesn't use adwaita theme, then will missing icons be inherited from it? (Question out of curiosity, I have never tampered with icon themes before this :sweat_smile:.)

dimkr:
Looks like the answer is positive - `strace evince 2>&1  grep Adwaita` shows that Evince takes the symbolic icons from Adwaita, even if you use a different theme.

woof-code/packages-templates/gtk3_FIXUPHACK
```diff
@@ -13,7 +13,7 @@ if [ -d usr/share/icons/Adwaita ]; then
 	rm -rf [0-9]*x* cursor* scalable-*
 	find scalable -type f | while read ICON; do
 		case $ICON in
-		*/pan-*) ;;
+		*/pan-*|*/send-to-symbolic.svg|*/open-menu-symbolic.svg|*/document-edit-symbolic.svg) ;;
```
lakshayrohila, line 16: Aren't you just preserving the symbolic links and not the icons which the symbolic links point to?

dimkr: Maybe, we'll know soon :)

dimkr: Works great!

image

lakshayrohila: That theme is the default theme? Seems completely changed!

dimkr: Puppy Standard

image

Some GTK+ 3/4 applications use mostly symbolic icons, so they don't follow the chosen icon theme and use those flat icons instead. Blueman is a good example.

dimkr: And only a tiny fraction of Adwaita is present,

image

lakshayrohila: Should be a great PR, but since we have developed icons (2 are in progress as you know, 1 is resolved), doing this, if I understand correctly, may allow evince to use Adwaita icons instead of puppy icons. Don't know if it will really use Adwaita icons even if those are present in default theme used by user.

dimkr: I prefer the solution proposed by @01micko and won't be offended if we close this PR.

lakshayrohila: I have tested this on my jammy64 by reinstalling whole adwaita theme. Evince does not seem to be respecting PMaterial symlinks. Rather, when I moved those icons from `icon-name.svg` to `icon-name-symbolic.svg`, it seemed to be working. We could actually merge this, just that we'll have to move symlinks to `*-symbolic.svg` for only those two symbolic icons. Still don't know if this solution will work for `document-edit.svg` as well.

PR #3786, opened 2022-12-29
More fixes for newer 'grep'

01micko:
Also removed a dead line in `latest_microcode.sh`


01micko:
Did a build and was successful with zero grep warnings. Merging.

PR #3789, opened 2022-12-29
hicolor: add changes-prevent.svg and changes-allow.svg

### LINKED ISSUE #3787: Jammy64: pavucontrol: icon not found.
Completely like #3781, except that with pavucontrol this time. Here are the screenshots:-

And with breeze:-
breeze1

The name I found for the icon to be added:-
sol

There was an image named `lock.svg` as well, but changing that was of no use.

I don't think PMaterial has same icon.

01micko:
This fixes a missing icon in pavucontrol
closes #3787


01micko:
pavu

dimkr:
Works great!

screenshot

PR #3795, opened 2022-12-31
Add error message if xsane is not present in xsaneshell.

lakshayrohila:


woof-code/rootfs-skeleton/usr/bin/xsaneshell
```diff
@@ -1,6 +1,13 @@
 #!/bin/sh
 #8.06.11 rodin.s: i18n
 
+xsane -v >/dev/null 2>/dev/null
```
01micko: I'd prefer something like
```
command -v xsane >/dev/null 2>&1 # or replace 'command -v' with 'type' - both are builtin - 0 for true 1 for false

File: woof-code/rootfs-skeleton/usr/bin/xsaneshell [hunk 2]
```diff
@@ -1,6 +1,13 @@
...
+if [ $? -eq 127 ]; then
```
01micko: `if [ $? -gt 0]; then`

woof-code/rootfs-skeleton/usr/bin/xsaneshell
```diff
@@ -1,6 +1,13 @@
...
+  /usr/lib/gtkdialog/box_splash -bg red -fg white -timeout 5 -text "ERROR: xsane is not installed. Install xsane first."
```
01micko: maybe add '-box close' else a spurious mouse over may close the message before read.

woof-code/rootfs-skeleton/usr/bin/xsaneshell
lakshayrohila: Why need to check it for greater than 0 when the shell will return error code 127 only when program is not found? Note that if the program was found but in case it returned 1 or some other error code, this script to test whether the program existed will fail. So why not just check if it returned 127?

01micko (approved):

woof-code/rootfs-skeleton/usr/bin/xsaneshell
```diff
@@ -1,9 +1,9 @@
 #!/bin/sh
 #8.06.11 rodin.s: i18n
 
-xsane -v >/dev/null 2>/dev/null
-if [ $? -eq 127 ]; then
-  /usr/lib/gtkdialog/box_splash -bg red -fg white -timeout 5 -text "ERROR: xsane is not installed. Install xsane first."
+command -v xsane >/dev/null 2>&1
+if [ $? -gt 0 ]; then
+  /usr/lib/gtkdialog/box_splash -bg red -fg white -timeout 5 -box close -text "ERROR: xsane is not installed. Install xsane first."
```
01micko: my bad, should be `-close box`

lakshayrohila:
@01micko, please review.

01micko (approved):
Sorry about that. Much better!

lakshayrohila:
Few days back, I did the same solution for cups_shell as well. Need to update the method there as well.

lakshayrohila:
@dimkr, please review.

dimkr (approved):

PR #3801, opened 2023-01-02
Change the defaults of release.yml to what most users use

dimkr:


lakshayrohila:
Ah, that's a relief! Thanks!

PR #3803, opened 2023-01-02
Add pmusic to jammy64.

lakshayrohila:


woof-distro/x86_64/ubuntu/jammy64/_00build.conf
```diff
@@ -41,7 +41,7 @@ USR_SYMLINKS=yes
...
+PETBUILDS="busybox aaa_pup_c disktype dmz-cursor-theme firewallstatus freememapplet geany gexec gpicview gtkdialog gtk_theme_flat_grey_rounded gtk_theme_polished_blue gtk_theme_gradient_grey gtk_theme_buntoo_ambience gtk_theme_stark_blueish gxmessage jwm l3afpad lxtask lxterminal mtpaint pa-applet pmusic powerapplet_tray rox-filer claws-mail transmission xarchiver xcur2png xdelta xdg-puppy-jwm Xdialog xlockmore yad pmaterial_icons puppy_standard_icons puppy_flat_icons ram-saver connman-puppy connman-ui connman-gtk fixmenusd spot-pkexec notification-daemon-stub palemoon weechat osmo abiword gnumeric xournalpp mpv efilinux gparted"
```
dimkr, line 44: pmusic is not a petbuild, it's in rootfs-packages. It will get auto-enabled if you add rootfs-packages.conf or add ffmpeg do DISTRO_PKGS_SPECS (see https://github.com/puppylinux-woof-CE/woof-CE/blob/e3963fcc13c9073da8cdf80422f57d5d1a233931/woof-code/3builddistro#L128).

lakshayrohila: BTW, is deadbeef installed in latest builds? It is set as a default, so if it is present, then this PR may not be needed.

PR #3804, opened 2023-01-03
Add a small deadbeef petbuild

dimkr:
It's a slightly older version that doesn't depend on clang at build-time, and built without extra features (like artwork display and streaming) and extra dependencies (like ffmpeg). It's a small and lightweight solution for audio playback, especially for computers too slow for Spotify.

Some optional features are auto-enabled if the dependencies are present.

If the mpv petbuild is enabled, ffmpeg is already included in the build, so deabeef can use libflac, libmpg123, etc' (ffmpeg's dependencies) and needs only one extra dependency (libjansson). Very compact!

image

@lakshayrohila Waiting for your approval


lakshayrohila:
Go on! I personally don't use deadbeef, or any music application at all. For me, youtube is the best :grinning:. Just that I think its a norm to keep at least one application for each multimedia type.

BTW, will deadbeef also support audio from CDs (never used one on Linux, but updating HOWTO-multimedia.htm, I understood that audio CDs require different apps for listening to them)?

dimkr:
It should support CD playback if the required libraries are present.

lakshayrohila:
Let's go on with these changes and see what happens.

PR #3806, opened 2023-01-04
Put fdrv and zdrv below pdrv when using overlay

dimkr:
This inconsistency with aufs can cause issues for users who assume the same stacking order.

~~(Not tested yet)~~

image


dimkr:
Required by #3808, merging.

PR #3814, opened 2023-01-07
Add xpad petbuild

dimkr:
This is an old version, to avoid the GtkSourceView dependency of later versions, and built without session management. It's much smaller than notecase (~150 KB), uses GTK+ 3 and very easy to use. It also seems to work fine under Wayland, without the tray icon.

xpad

@lakshayrohila


lakshayrohila:
Go on!

PR #3827, opened 2023-01-13
Add viewnior petbuild

dimkr:
A patch makes exiv2 optional to avoid extra dependencies compared to gpicview.

viewnior


lakshayrohila (approved):
I was always in favour of viewnior, more than gpicview.

PR #3838, opened 2023-01-15
Add hardinfo petbuild

dimkr:
libsoup is an optional dependency and pavucontrol already depends on libjson-glib (it's only the development headers that are missing), so hardinfo doesn't have any extra dependencies. Not a tiny application but requested often and useful in a live distro.

@lakshayrohila Prefer this one or the old version?

hardinfo


lakshayrohila:
As you wish to :smiley:. I have no knowledge about these programs, and usually do not use them.

PR #3839, opened 2023-01-15
Add Settings menu item to main menu.

lakshayrohila:
This PR adds a "Settings" menu item to main menu, as you will usually find in any desktop environment (including windows and KDE). Here is a screenshot:
screenshot1

Earlier, it would have been difficult to reach the "Puppy Setup" menu item (MENU > Setup > Puppy Setup) + many users will confuse between "Puppy Setup" and "Desktop settings" ("Puppy setup" is surrounded by many menu items, but "Desktop settings" is the topmost menu item in the top-most menu category).

### DIFF
--- a/woof-code/rootfs-packages/jwm_config/etc/xdg/templates/_root_.jwmrc
+++ b/woof-code/rootfs-packages/jwm_config/etc/xdg/templates/_root_.jwmrc
@@ -40,6 +40,7 @@
 	PUPPYMENU jwm-xdgmenu /etc/xdg/menus/puppy-multimedia.menu
 	PUPPYMENU jwm-xdgmenu /etc/xdg/menus/puppy-fun.menu
 	<Separator/>
+	<Program label="Settings" icon="/usr/share/pixmaps/puppy/puppy_config.svg">wizardwizard</Program>
 	<Program label="Help" icon="help48.png">/usr/sbin/puppyhelp</Program>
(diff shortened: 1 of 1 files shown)


lakshayrohila:
@dimkr, @01micko, @peabee please review.

dimkr:
> "Puppy setup" is surrounded by many menu items

What do you think about moving it to the top of its submenu instead of making the main menu taller?

EDIT: if I understand correctly, `Categories=X-Setup` in /usr/share/applications/Wizard-Wizard.desktop should work

lakshayrohila:
> What do you think about moving it to the top of its submenu instead of making the main menu taller?

I'll test with that as well.

What I thought was that there must be a seperate menu entry for settings on the main menu, which is what most users new to puppy coming from windows or other DEs might want to find - something that explicitly says "settings".

**EDIT:** @dimkr, Setting `Categories=X-Setup` works fine, but I still want you to reconsider adding this menu item. It even though makes the main menu one entry longer, still keeping the "Settings" menu entry that way becomes easier to find even for a newbie to puppy.

peabee:
I don't feel the need for this change - and it could open a "can of worms" for consequential changes....
Puppy does not have to be identical to other distros - it is unique!

dimkr:
I think the best approach would be unifying wizardwizard and pdesktop into one **tabbed** wizard that has absolutely everything, then drop confusing menu entries for things like xorgwizard. The unique mess in Puppy is not a feature, in my mind.

PR #3840, opened 2023-01-15
Add "Default Apps" (puppyapps) to wizardwizard.

lakshayrohila:
"Default Apps" (or `puppyapps`) is not in either of `pdesktop` or `wizardwizard`. This PR adds it to `wizardwizard`.

screenshot2


lakshayrohila:
@dimkr, I wanted that "Default Apps" button's width was equal to other buttons' widths, but could not find any tag that may do this (`<width>` sets image width). Please help me doing that.

Thank you

lakshayrohila:
@dimkr, @01micko, @peabee please review.

dimkr:
It's a slippery slope, there are so many things that can be added here and don't 100% belong.

PR #3850, opened 2023-01-20
Restore removed menu entries in jammy64

dimkr:
@lakshayrohila Originally, I removed them because pdesktop and wizardwizard lead to the various configuration tools.

### DIFF
--- a/woof-distro/x86_64/ubuntu/jammy64/_00build.conf
+++ b/woof-distro/x86_64/ubuntu/jammy64/_00build.conf
@@ -159,7 +159,6 @@ mv -f root/.spot-status.orig root/.spot-status
 mv root/Downloads home/spot/
 chroot . chown -R spot:spot /home/spot/Downloads
 ln -s ../home/spot/Downloads root/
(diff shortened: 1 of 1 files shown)


lakshayrohila:
Do these menu items provide any more funtionality that `pdesktop` and `wizardwizard` do not provide?

dimkr:
Nope. I removed them because they duplicate what's inside pdesktop and wizardwizard. I think longer menus are harder to navigate and inconvenient with with wide screens most of us have.

lakshayrohila:
> Nope. I removed them because they duplicate what's inside pdesktop and wizardwizard. I think longer menus are harder to navigate and inconvenient with with wide screens most of us have.

Then... why, this PR?

Still if you think the change is needed, I have no issue with the PR.

dimkr:
> Then... why, this PR?

In case you're a fan of the Puppy tradition of long menus 😄

lakshayrohila:
> In case you're a fan of the Puppy tradition of long menus smile

Well, I am a fan of laziness - the lesser you have to navigate, the better :smile:.

PR #3852, opened 2023-01-20
Mount overlay with metacopy=on to save RAM and speed up save2flash

dimkr:


dimkr:
This doesn't work: overlay doesn't support offline changes to pup_ro1 and snapmergepuppy.overlay works only because the file in pup_rw masks the file in pup_ro1, so the changes never need to propagate upwards. With `metacopy=on`, the metadata sits in pup_rw while the contents sit in pup_ro1, breaking this guarantee.

PR #3861, opened 2023-01-21
Drop the debianutils template (fixes #3699)

### LINKED ISSUE #3699: (Few)debianutils programs not installed on jammy64.

dimkr:
@lakshayrohila Please let me know if this fixes #3699 for you.

### DIFF
--- a/woof-code/packages-templates/debianutils_FIXUPHACK
+++ b/woof-code/packages-templates/debianutils_FIXUPHACK
@@ -1,5 +1,0 @@
-# only want run-parts and tempfile
-# both are in /bin ...
-
-rm -rf sbin usr
-rm -f bin/which #shell script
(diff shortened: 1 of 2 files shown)


lakshayrohila:
I have run GitHub action with the changes. I'll tell you the result.

lakshayrohila (approved):
I have done some testing installing various shells - ksh, csh, tcsh, zsh, ...

apt doesn't give any problem. The shells also seem to be working as intended.

I think it is just the required fix for the issue.

lakshayrohila:
Thank you for the PR :+1: !

PR #3874, opened 2023-01-28
Align spup UIDs and GIDs with Slackware (fixes #3872)

### LINKED ISSUE #3872: /var/spool/cups has wrong ownership
As reported by @Marv....
Both S15Pup64 and S15Pup64 +2 built on Github
/var/spool/cups has ownership root:7 whereas it should be root:lp

Screenshot

Locally built +1 had the correct ownership

dimkr:
This PR changes UIDs and GIDs so they match those in etc-15.0-x86_64-17.txz, without adding all the predefined users and groups in Slackware to achieve maximum compatibility.

```
-daemon:x:1:1::/:
+daemon:x:2:2::/:
-bin:x:2:2:bin:/bin:
+bin:x:1:1:bin:/bin:
-ftp:x:1000:1000:Linux User,,,:/root/ftpd:/bin/sh
+ftp:x:14:50:Linux User,,,:/root/ftpd:/bin/sh
-haldaemon:x:108:119:Hardware abstraction layer,,,:/var/run/hald:/bin/false
+haldaemon:x:82:82:Hardware abstraction layer,,,:/var/run/hald:/bin/false
-messagebus:x:503:503:Linux User,,,:/tmp:/bin/sh
+messagebus:x:81:81:Linux User,,,:/tmp:/bin/sh
-uucp:x:10:10:uucp:/var/spool/uucp:/bin/sh
+uucp:x:10:14:uucp:/var/spool/uucp:/bin/sh
```

```
-daemon:x:1:
+daemon:x:2:
-users:x:500:
+users:x:100:
-bin::2:root,bin,daemon
+bin::1:root,bin,daemon
-messagebus:x:503:messagebus,root
+messagebus:x:81:messagebus,root
-ftp:x:1000:
+ftp:x:50:
-uucp::10:
+uucp::14:
-netdev::113:
+netdev::86:
-haldaemon::119:
+haldaemon::82:
-disk::100:root,spot
+disk::6:root,spot
-audio::101:root,spot
+audio::17:root,spot
-lp::102:root,daemon,spot
+lp::7:root,daemon,spot
-dialout::103:root,spot
+dialout::16:root,spot
-kmem::104:root,spot
+kmem::9:root,spot
-video::105:root,spot
+video::18:root,spot
-floppy::106:root,spot
+floppy::11:root,spot
-cdrom::107:root,spot
+cdrom::19:root,spot
-tape::108:root
+tape::26:root
-tty::109:root,spot
+tty::5:root,spot
-plugdev::110:root,spot
+plugdev::83:root,spot
-input:x:31:root,spot
+input:x:71:root,spot
```


peabee (commented):
Needs a test with these changes in a ydrv ....

Does cupsd need to be changed? Does anything else have consequential changes?

Do other distros need the same treatment?

@Marv says:

> Chased it a bit more. With the stock +2 main SFS file, the permissions of /var/spool/cups and /var/spool/cups/tmp are changed to root:lp. This is true in a pristine, pristine +ydrv, pristine +ydrv and 6.1.5 kernel, and with the above plus my savefile with /var/spool scrubbed before boot. However, printing fails in all of the above, png attached.
> 
> I make one and only one change. /var/spool/cups and /var/spool/cups/tmp permissions set to root:lp in the main sfs. Reboot in any of the above configurations and printing is sucessfull.
> 
> I can't find what CUPS isn't setting to root:lp. Is it a file created in /tmp? At any rate I am running -happily- with that one change in my main +2 SFS.
> 
>

dimkr:
> Needs a test with these changes in a ydrv ....

This is not enough because overlaying a different /etc/groups on top of an existing spup will not fix the permissions of files `chown`ed during the build process. If `chown x:y` sets the ownership of a file to `root:tty` and now GID `y` belongs to `daemon`, files will now be owned by `root:daemon` instead.

> Does cupsd need to be changed? Does anything else have consequential changes?

First of all, you'll need to do a fresh woof-CE build without cache of any kind. Then, if you use pet packages that assume that `input` is 31, but now it's 71 (and so on) they will need to be re-packaged after a `chown`. You can detect such packages using `find / -user`, etc'. Maybe all .pet packages are OK - I haven't checked.

> Do other distros need the same treatment?

No. Debian, Ubuntu and Devuan already have their special /etc/{group,passwd,shadow}.$distro files in rootfs-skeleton and all IDs match. spup is late to this party.

peabee:
I tested with this small ydrv and cups seemed to work:
[ydrv_s15pup64_22.12.sfs-false.gz](https://github.com/puppylinux-woof-CE/woof-CE/files/10527794/ydrv_s15pup64_22.12.sfs-false.gz)

peabee:
> First of all, you'll need to do a fresh woof-CE build without cache of any kind. Then, if you use pet packages that assume that `input` is 31, but now it's 71 (and so on) they will need to be re-packaged after a `chown`. You can detect such packages using `find / -user`, etc'. Maybe all .pet packages are OK - I haven't checked.
> 
> > Do other distros need the same treatment?
> 
> No. Debian, Ubuntu and Devuan already have their special /etc/{group,passwd,shadow}.$distro files in rootfs-skeleton and all IDs match. spup is late to this party.

That sounds like a major undertaking!!!

I was thinking of Void......

peabee:
Perhaps just a minor change of lp from 102 to 7 would be "safer"?

dimkr:
> Perhaps just a minor change of lp from 102 to 7 would be "safer"?

It's easier to change only one mismatch among many, yes, but you'll come back here every time somebody complains about a broken Slackware package with messed up ownership.

peabee (changes requested):
I prefer to go the simple route and just change lp - this could be done by this mechanism or by additions to EXTRA_COMMANDS in _00build.conf or even a permanent change to rootfs-skeleton/etc/group if it can apply to all builds

Given the long history of using Slackware packages with the "Puppy groups" I think there is less chance of future problems by largely sticking to the status-quo.

PR #3876, opened 2023-01-29
updated some miss information.

Linux-is-king:
corrected some miss information about hard drive install.. hard drive (frugal) install works with uefi, this is extremely confusing and needs to be fixed. Hard drive frugile install dose not work with puppyie swith a BDRV file, ( not all puppies).

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/dotpup
+++ b/woof-code/rootfs-skeleton/usr/sbin/dotpup
@@ -51,12 +51,12 @@ BOOTFLASHBUTTON=$(install_button \
 	"$(gettext "Quickly create a bootable usb flash drive")" \
 	'bootflash &' flashcard_usb.svg )
 
-if [ -d /usr/share/g4dos -o -f /usr/local/frugalpup/grub4dos.tar.xz -o -f /usr/lib/grub4dos/grldr ] ; then
+#if [ -d /usr/share/g4dos -o -f /usr/local/frugalpup/grub4dos.tar.xz -o -f /usr/lib/grub4dos/grldr ] ; then
 	FULLINSTALLBUTTON=$(install_button \
 	"$(gettext "Installer")" \
-	"$(gettext "Currently doesn't support UEFI")" \
+	"$(gettext "Hard Drive Install:")" \
 	'puppyinstaller &' puppy_install.svg)
-fi
+#fi
 
 if [ -e /dev/sr0 ] && [ -x /usr/bin/cdrecord -o -x /usr/bin/xorriso ] ; then
 	BURNISOBUTTON=$(install_button \
--- a/woof-code/rootfs-skeleton/usr/sbin/puppyinstaller
+++ b/woof-code/rootfs-skeleton/usr/sbin/puppyinstaller
@@ -3,6 +3,8 @@
 #(c) Copyright 01micko, shinobar and others
 #2007 Lesser GPL licence v2 (/usr/share/doc/legal/lgpl-2.1.txt)
 #140705 major overhaul of guis and conditions.
+#2023 a few changes made to correct some misinformation. 
+
 
 # to enable full install to hdd (before running this script):
 #   export ENABLE_FULLINSTALL=1 
@@ -10,7 +12,7 @@
 [ "`whoami`" != "root" ] && exec sudo -A ${0} ${@} #110505
 
 if [ -d /sys/firmware/efi ] ; then
-	/usr/lib/gtkdialog/box_yesno --ok-cancel --no-first --yes-label Continue --no-label Close --warning "Puppy Installer" "This installer does not support UEFI. It's only useful if you want to install to a drive that will not be booted on this machine.."
+	/usr/lib/gtkdialog/box_yesno --ok-cancel --no-first --yes-label Continue --no-label Close --warning "Puppy Installer" "This installer is ment to be used for Hard drive installs. It is known to cause errors in vanilla Dpup and other pups with a BDRV file, Check your USB ( or other istall media) to see if one is present."
 	[ $? -ne 0 ] && exit
 fi
 


woof-code/rootfs-skeleton/usr/sbin/dotpup
```diff
@@ -51,12 +51,12 @@ BOOTFLASHBUTTON=$(install_button \
 	"$(gettext "Quickly create a bootable usb flash drive")" \
 	'bootflash &' flashcard_usb.svg )
 
-if [ -d /usr/share/g4dos -o -f /usr/local/frugalpup/grub4dos.tar.xz -o -f /usr/lib/grub4dos/grldr ] ; then
+#if [ -d /usr/share/g4dos -o -f /usr/local/frugalpup/grub4dos.tar.xz -o -f /usr/lib/grub4dos/grldr ] ; then
```
dimkr, line 54: Why is it commented out? dpup doesn't have grub4dos preinstalled and installation **will** fail if you blindly assume it exists.

woof-code/rootfs-skeleton/usr/sbin/puppyinstaller
```diff
@@ -3,14 +3,16 @@
...
+
 
 # to enable full install to hdd (before running this script):
 #   export ENABLE_FULLINSTALL=1 
 
 [ "`whoami`" != "root" ] && exec sudo -A ${0} ${@} #110505
 
 if [ -d /sys/firmware/efi ] ; then
-	/usr/lib/gtkdialog/box_yesno --ok-cancel --no-first --yes-label Continue --no-label Close --warning "Puppy Installer" "This installer does not support UEFI. It's only useful if you want to install to a drive that will not be booted on this machine.."
+	/usr/lib/gtkdialog/box_yesno --ok-cancel --no-first --yes-label Continue --no-label Close --warning "Puppy Installer" "This installer is ment to be used for Hard drive installs. It is known to cause errors in vanilla Dpup and other pups with a BDRV file, Check your USB ( or other istall media) to see if one is present."
```
dimkr, line 15: Maybe just fix the installer and copy *.sfs (including bdrv) instead of adding this excuse here?

dimkr (changes requested):
Let's just fix the installer instead of adding misinformation and excuses.

woof-code/rootfs-skeleton/usr/sbin/dotpup
```diff
@@ -51,12 +51,12 @@ BOOTFLASHBUTTON=$(install_button \
...
 	FULLINSTALLBUTTON=$(install_button \
 	"$(gettext "Installer")" \
-	"$(gettext "Currently doesn't support UEFI")" \
+	"$(gettext "Hard Drive Install:")" \
```
dimkr, line 57: This label was true, `puppyinstaller` doesn't support UEFI.

Linux-is-king: No, the hard drive installer works with uefi, I have uefi and have tested it. Full install (on hard drive) doesn't work with uefi. Also the commit blocking out the grub for dogs was an error ment to leave a commitment that I need to come back and improve, because it still needs improvement, it only checks for grub 4dos. It does not check if grub 2 is already installed by another os ...

woof-code/rootfs-skeleton/usr/sbin/puppyinstaller
Linux-is-king: Yea that would be a better option, but it is currently the bdrv breaking the hard drive install. I would know I have spent 20+ hours checking and testing this...

woof-code/rootfs-skeleton/usr/sbin/dotpup
Linux-is-king: To be more clear the puppy installer works with uefi in frugal mode. It does not work with "full install".. for more context that you probably not aware of ( this is some stuff from WAY. Back.. ) puppy Linux used to have a install method that would un layer the file system and install it like a " normal" Linux distro to a hard drive .. this full install method is the one that doesn't work with uefi. Full install was never recommended, because it takes away all the advantages of puppy. It was only useful in a handful of situations...

Linux-is-king:
Installer is fixed, and tested!  please see pull request and read the comments, I believe I see how you got confused.. also please make sure it's set to exe ... ( I'm new to GitHub and miss that often..  )

woof-code/rootfs-skeleton/usr/sbin/dotpup
dimkr: Look closely at puppyinstaller, it refuses to do anything if you have GPT, and it never mounts the ESP, so it can't put the kernel there. I think your machine is configured to allow BIOS boot and what's why the installer "works" with UEFI.

Linux-is-king: Nope it's uefi. I mean I can send a full boot video if you don't believe me. The only 2 places I have still had issues was with my $200 2 in 1 that only has flash memory, and inside Virtual box ( virtual box only supports uefi with certain os, so I didnt do any more looking into it). I want to say the drive I have fossfa64 installed to is has a gpt table ( but I could be wrong here.) . I have spent  40 + hours working on this, I have probably done 8 plus install on 2 different machines.. one of those installs I tried a traditional full install ( just to see what would happen) it didn't boot I'm presuming because of the uefi.. I don't have the full specs of my laptop on me but my gaming PC specs  as follows: 
Mobo ASRock b550 gaming
32 gbs ram
Ryzen 5600g
Rx6600 gpu
I have 3 storage device 
1tb nvme
4tb spinning disk
500 gb SSD - with fosfa installed here on a 125 gb partition.

So the installer is still a little wonky, it Dosent tuch the boot partition it calls grub4 dos at the end, if you have grub4 dos then it does scan the system and suggest a boot menu lay out ( my laptop has only grub 4 dos) it dosent work as well with grub2, you have to manually set the menu  it see 
https://forum.puppylinux.com/viewtopic.php?t=437 for reference.. 
For a good "how too" but the installer does do a pop up with the relevant info. 
Generally speaking for most Linux distros after install you would go back to your main distro and just update grub, grub would then find you the Linux distro and add it to the menu, grub 2 dosen't  find puppy correctly....  ( I plan on trying to find fixes for this later but it's a grub2 problem I believe. ) 

But to sum it up unless my asrock 550b came factory set to legacy boot, and the full installer is broken elsewhere ( somewhat likely) and the warning is getting tripped even with my efi being disabled... It works with uefi, I can triple check this tomorrow and even send a full boot videos with me choosing efi on my laptops one time boot menu if you don't believe me but I mean all three would have had to happen.. so yes the installer works as is one uefi, does it work well. No but it does work..  like I said I could see where your confusion comes from, but please have a look at my changes and give them a try, it works pending it's not a flash memory internal  storage device, and it's not inside Virtual box ( I did not test other virtual systems .. ).

Linux-is-king: To add the issue with the two in one is the installer finding the drive..

dimkr:
Can you point out where the fix is, please? (red - original code, green - your changes)

diff

Linux-is-king:
I thought I made another patch for the puppyinstaller, and didn't hit send, sorry about that ! I am a GitHub newbie .. lol 😂 . I just sent it as patch-2

Linux-is-king:
As I believe I have stated please make sure it is set to execute. ( I believe I fixed that probablem before I made the lull request. )

PR #3877, opened 2023-01-29
Change lp group to 7 from 102

peabee:
To enable non debian/ubuntu builds to be configured correctly for cups when built on github using ubuntu dev facilities. Debian/ubuntu builds already have different group identities through the etc/group.xxx mechanism.

### DIFF
--- a/woof-code/rootfs-skeleton/etc/group
+++ b/woof-code/rootfs-skeleton/etc/group
@@ -18,7 +18,7 @@ sshd::33:sshd
 webgroup:x:504:
 disk::100:root,spot
 audio::101:root,spot
-lp::102:root,daemon,spot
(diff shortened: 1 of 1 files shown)


dimkr:
Helpful note in case anyone stumbles upon this PR in the future and tries to understand why this change is needed and why it works: Slackware's `lp` is GID 7, while Puppy had 102. Unlike Debian, Ubuntu and Devuan, which have their custom group, passwd and shadow in rootfs-skeleton, that align UIDs and GIDs with the compat distro, Slackware and Void don't have such files. This change is a hack that will solve bad ownership of CUPS files due to GID mismatch of `lp` (and only this group, without aligning the others and without adding missing groups) for Slackware builds, with unknown impact on Void, and zero impact on others.

PR #3879, opened 2023-01-30
sending another pull request

Linux-is-king:
I am submitting another pull request, I believe there is a lot of confusion mostly coming from the variable names.. to be clear puppyinstaller works with efi in frugal mode.. the "fullinstall" button launches the puppyinstaller to do a frugal install.. full install is locked some how .. read the begging of the script. any ways I have fixed the puppyinstaller and can confirm it now works with efi, bdrv, as well as non-bdrv pups .. please read the commits. I also still left the grub for dos check disabled, read comment to understand why I believe that needs to be done, but if nothing else please implement the other changes...

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/dotpup
+++ b/woof-code/rootfs-skeleton/usr/sbin/dotpup
@@ -38,7 +38,7 @@ install_button() { #text1 text2 action icon
   </hbox>
   '
 }
-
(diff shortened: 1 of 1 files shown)


Linux-is-king:
please read ..

dimkr:
I don't see any fixes in puppyinstaller in this PR. Are you sure this is the right branch?

https://github.com/puppylinux-woof-CE/woof-CE/pull/3879/files

PR #3881, opened 2023-01-30
this will fix the puppyinstaller to work with bdrv

Linux-is-king:
hey, I fixed the puppy installer, please read the commits I made in the above file.. I am also submitting another pull request for the dotpup file, I can see how you are confused, but the puppyinstaller does work with efi in " frugal mode" so puppy Linux use to have multiple hard drive install methods, the one you are referring to that dose not work with efi is the "full install" with this install method the file system was un-compressed , so it was never really recommended .. and the dot pup variable names adds to this confusion..  I hope I don't come off as rude, but I have tested this on 3 machines all of which have efi ..

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/puppyinstaller
+++ b/woof-code/rootfs-skeleton/usr/sbin/puppyinstaller
@@ -8,12 +8,11 @@
 #   export ENABLE_FULLINSTALL=1 
 
 [ "`whoami`" != "root" ] && exec sudo -A ${0} ${@} #110505
-
-if [ -d /sys/firmware/efi ] ; then
-	/usr/lib/gtkdialog/box_yesno --ok-cancel --no-first --yes-label Continue --no-label Close --warning "Puppy Installer" "This installer does not support UEFI. It's only useful if you want to install to a drive that will not be booted on this machine.."
-	[ $? -ne 0 ] && exit
-fi
-
+## this is false "puppyinstaller" workes with efi... in FRUGAL MODE only. full install is already behind a lock read above. (full install does NOT work with efi) ... this is all really old code from pupies legacy days... 
+#if [ -d /sys/firmware/efi ] ; then
+#	/usr/lib/gtkdialog/box_yesno --ok-cancel --no-first --yes-label Continue --no-label Close --warning "Puppy Installer" "This installer is ment to be used for Hard drive installs."
+#	[ $? -ne 0 ] && exit
+#fi
 export TEXTDOMAIN=puppyinstaller
 export OUTPUT_CHARSET=UTF-8
 . gettext.sh
@@ -34,6 +33,7 @@ if [ "$PUPSFS" ] ; then #PUPSTATE
 	DISTRO_FDRVSFS=${FDRV##*/}
 	DISTRO_ADRVSFS=${ADRV##*/}
 	DISTRO_YDRVSFS=${YDRV##*/}
+	DISTRO_BDRVSFS=${BDRV##*/}
 fi
 
 #===================================================================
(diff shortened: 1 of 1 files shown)


Linux-is-king:
This is the correct pull request I believe, I thought I had sent this one before the other lol 😂😂😂 ..

dimkr:
If you reduce this PR to these 4 lines (= no unrelated changes), no objection:

lines

woof-code/rootfs-skeleton/usr/sbin/puppyinstaller
```diff
@@ -1071,26 +1071,26 @@ $(gettext 'your PC for all installations of Puppy and other operating systems)')
...
-			msg="$(printf "$(gettext 'Copying files from %s to %s...')" "$SFS" "$DESTPART")"
-			yaf-splash -close never -bg orange -text "$msg" &
-			XPID=$!
-			FREEDEV=$(losetup -f)
-			losetup ${FREEDEV} "${SRCPATH}/${SFS}" #100913
-			mkdir -p ${DESTMNTPT}/srcmntpt
-			mount -r -t squashfs ${FREEDEV} ${DESTMNTPT}/srcmntpt
-			cp -a --remove-destination ${DESTMNTPT}/srcmntpt/* ${DESTMNTPT}/ || {
+		for SFS in ${DISTRO_PUPPYSFS} ${DISTRO_ZDRVSFS} ${DISTRO_FDRVSFS} ${DISTRO_ADRVSFS} ${DISTRO_YDRVSFS} ${DISTRO_BDRVSFS}
+			do
```
lakshayrohila, line 1075: Indentation is messed up from here to L1085.

Linux-is-king:
Can we also remove the uefi warning from dotpup?

dimkr:
That UEFI warning is still true as far as I see. puppyinstaller doesn't install a UEFI boot loader, doesn't copy the kernel to the ESP and even refuses to do GPT partitioning. I see no traces of UEFI support.

Linux-is-king:
If so I will submit a new pull request with only that in and a comment explaining the frugal/ full install confusion in both dotpup and puppyinstaller (this has caused a lot of issues and it would be nice not to repeat it like I said it is confusing to any one who didn't use puppy in 2010 lol 😆😆😆😆) .

Linux-is-king:
Grub4 dos work with uefi..

Linux-is-king:
I will double check that my gaming computer built in 2021 with an up to date Mobo and Mobo firmware wasn't set to legacy boot ( from factory )  when I get home, if needed but see below until then :
https://forum.puppylinux.com/viewtopic.php?t=5275
https://forum.puppylinux.com/viewtopic.php?t=2479
I have spent a lot of hours testing this , so unless I  some how missed that, then it works... I understand it's confusing, Im assuming it probably reads it similar to a boot USB, not really sure on that one.. but it works.. now full install mode as explained dosen't.

dimkr:
> Grub4 dos work with uefi..

Are you sure about this? AFAIK grub4dos is derived from legacy GRUB 0.x, which doesn't support UEFI and doesn't support GPT. If it doesn't support GPT and puppyinstaller doesn't do GPT partitioning, it can't support UEFI.

Linux-is-king:
So update, gaming PC was in "compitablity mode" I turned it off which means uefi only, puppy Linux fossfa booted no problem. The test laptop was a slightly different story, so the first install was done after a wipe, it did show a uefi boot option after, however if selected it would load grub4dos, but none of the puppies would boot.. I remember doing something to get it to work the first time round, so regardless the installer does need more work... That being said though it is booting on my gaming PC with grub 2 ..

Linux-is-king:
Also yes, grub4dos can now be used with msdos partition tables to boot uefi. As stated my laptops bios did detect a uefi, and load grub4dos when selected.

Linux-is-king:
After doing slightly more research it appears most motherboards support a compitablity mode which will allow both uefi and legacy boot... So I'm not sure how much more work needs to be done, it will work on most machines as is. Maybe update the warning to read something along the lines of "only works with mobos that have a compitablity or legacy mode " ... Still not sure why it's loading on my gaming PC though lol 😆.

Linux-is-king:
https://www.partitionwizard.com/partitionmanager/csm-support-bios.html

Linux-is-king:
So the current warning is false on many machines .. because it reads something to the effect of " this is only useful if you want to boot this drive on another machine..." Which definitely isn't true .

PR #3882, opened 2023-01-31
Update links in icon_switcher

lakshayrohila:
@dimkr, @peabee, earlier, pressing "More Themes" button would open up the link: http://oldforum.puppylinux.com/viewtopic.php?t=31059. I added another link: http://smokey01.com/icon-sets to this list.


lakshayrohila:
@dimkr, @peabee please review. I also don't know if it will be good to add link to smokey01's site in case it gets closed in future and we have to change the links again. I am open to change that link, just that I feel the link to get more icons should open up site containing themes only - oldforum's topic contains hundreds of pages with many number of discussions, not many themes. I only wanted that the link should be clear and direct to the point, just like smokey01's site has got.

PR #3885, opened 2023-02-03
Wait for 5s for a flash drive only if not recognized yet

dimkr:
I totally understand why this `sleep 5` is needed if the flash drive where the save is wasn't recognized yet. However, it's a waste of time on some computers, where it doesn't take so much time to recognize a flash drive. I have a 12 years old laptop that works fine (and boots much faster) without this delay.

Maybe, the logic should be changed so the init script waits for 5 seconds only if the partition wasn't found? (And/or, do this search thing every 1s and stop waiting if the partition was found?)

(This line was introduced in 22a6442962.)

### DIFF
--- a/initrd-progs/0initrd/init
+++ b/initrd-progs/0initrd/init
@@ -942,7 +942,7 @@ elif [ "$PDRV" = "" ];then #not specified anywhere
 fi
 
 if [ "$LOOK_PUP" -o "$LOOK_SAVE" ];then #something to search for
- [ "${PMEDIA:0:3}" != "usb" ] && search_func
+ search_func
  NUM=0
  while [ "$LOOK_PUP" -a "$P_PART" = "" ] || [ "$LOOK_SAVE" -a "$SAVEPART" = "" ];do
   [ $NUM -ge $WAITDEV ] && break


peabee (commented):
The description of the issue seems logical however I don't see how the change proposed removes a 5 sec sleep....... just calls search_func unconditionally....?
Comment from @gyrog and @mavrothal as the original authors would be welcome if available.

dimkr:
> just calls search_func unconditionally....?

Originally, the code unconditionally waits for 5s when `pmedia=usbflash`, then calls that function. With this PR, this function is called immediately even if `pmedia=usbflash`. If it succeeds (the device is found), we don't waste 5s.

dimkr:
I've been using this plus #3888 for some time, with an old SSD with I/O errors every once in a while. These two PRs allow me to install Puppy with `pmedia=usbflash`. Puppy runs with PUPMODE 13, everything is in RAM and I get a chance to save a backup of my home directory if the SSD fails during `snapmergepuppy`. If I boot from a flash drive to repair a partition, now Puppy boots 5s faster.

dimkr:
@okwizard @peabee Objections to this PR?

peabee:
> @okwizard @peabee Objections to this PR?

Sorry - no view - unable to test....

Would be best to do a test build and submit to the forum for testing.....

PR #3888, opened 2023-02-04
Poll block devices every 1s instead of waiting for 5s every time

dimkr:
If the flash drive is recognized after 1s but the timeout is 5s, boot is delayed by 1s and not by 5s.

~~In addition, this PR increases the default timeout from 5s to 20s: some users complain that 5s is not enough, and it doesn't matter if the timeout is 4x as long if users who needed 5s delay still wait for 5s, not more. 20s is just an upper limit now.~~ 

This is not the same as #3885: the two PRs complement each other. This PR makes Puppy boot faster if it takes time to recognize a flash drive: the init script waits until the flash drive is recognized instead of waiting for 5 seconds every time. The latter skips the entire waiting thing if the flash drive is recognized immediately and no delay is needed.

EDIT: 026bff1 also relaxes `pmedia=cd` and replaces the 15s delay with polling every 1s.

EDIT 2: 92c173b changes the meaning `pmedia=cd` and `pmedia=` (unspecified or empty) to _"search all partitions for Puppy files and a save file/folder in 1s intervals, up to 5 times by default"_. This should make Puppy boot more reliably when installed to a flash drive using a third-party tool that copies `pmedia=cd` from grub.cfg. This assumes most Puppy users don't use CDs. In addition, this makes it possible to boot Puppy from a CD but save to a flash drive that isn't recognized immediately.

### DIFF
--- a/initrd-progs/0initrd/README.txt
+++ b/initrd-progs/0initrd/README.txt
@@ -124,7 +124,7 @@ Boot parameters:
 
 pmedia=<atahd|ataflash|usbhd|usbflash|cd> 
    Indicates the type of boot device.
-   If it's "cd" then the partitions are searched for a save layer file, the only situation that triggers such a search.
+   If it's "cd" then all partitions are searched for Puppy files and a save layer file.
    If the first 3 characters are "usb", then any searching is restricted to only usb devices.
    If the last 5 characters are "flash" the top layer in the stack remains the tmpfs in memory, otherwise any found save layer becomes the top layer in the stack.
    This boot parameter should always be provided.
--- a/initrd-progs/0initrd/init
+++ b/initrd-progs/0initrd/init
@@ -29,7 +29,7 @@ L_UPGRADE_NO="Backing off, not using personal storage, booting in RAM only, PUPM
 L_LOADING_FROM_CD="Loading folder %s from CD/DVD..." #printf
 L_RAM_DISK_FULL="RAM disk full, copy failed at %s" #printf
 L_ADDING_SAVE_LAYER_FAILED="adding %s to aufs stack failed." #printf
-L_WAITING_FOR_DEV='Waiting %s seconds for slow storage devices.' #printf
+L_WAITING_FOR_DEV='Waiting %s seconds for slow storage devices [%s/%s].' #printf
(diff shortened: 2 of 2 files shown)


dimkr:
Pushed 026bff1. It makes Puppy boot faster and more reliably when `pmedia=cd`, which is important because that's what mk_iso.sh puts in the ISO.

With this change, instead of waiting for an iso9660 file system and giving up if it's not found (something that happens frequently, because users and tools like Rufus copy that `pmedia=cd` to their installations on FAT32 or whatever), the init script searches for Puppy files in all partitions.

Most Puppy users nowadays probably write the ISO image to a flash drive, so there's no reason to wait for sr0 for 15s (!) before boot can start (or fail hard because no such partition was found), including on computers that have no optical drive. Polling all block devices every second if `pmedia=cd` (or only USB block devices if `pmedia=usbflash`), with timeout >= 15s, makes Puppy boot faster for most users without sacrificing reliability or breaking what worked for many years for those who use Puppy CDs. (And it makes sense to read the main SFS from disk and not from a CD, because that should be much faster.)

dimkr:
@okwizard @peabee Objections to this PR?

peabee (approved):
Sorry - no view - unable to test....

Would be best to do a test build and submit to the forum for testing.....

dimkr:
If I understand correctly, this PR is included in F96-CE_3 and fixes multiple real-world problems Puppy suffers from for years (#3962).

#3885 is extra sugar on top, can reduce boot time by 1s for many users.

okwizard:
@dimkr 
@peabee 
You are correct, radky used the init code from 4a23874 in F96-CE_3 and it passes all of the boot case test that were run for USB and CD

Thanks
Gary

dimkr:
Let's wait for several days. If people keep complaining that Puppy doesn't boot in various scenarios (because of the bad default of `pmedia=cd`) or doesn't work with UNetbootin or Rufus after these changes, I'll merge the two PRs.

The missing /EFI directory is something that can be fixed later.

dimkr:
Still quiet on the F96-CE_3 front, no complaints about failed boot. Merging.

PR #3920, opened 2023-02-19
Enable the gnumeric petbuild in s15pup (fixes #3912)

### LINKED ISSUE #3912: gnumeric petbuild has unsatisfiable dependency for Slackware & Gtk2
gnumeric depends on goffice

There is no goffice petbuild

Slackware does not provide goffice

Gtk2 builds of goffice are no longer available

dimkr:


peabee (changes requested):
Have you built and fully tested? Where is the .iso to download?

dimkr:
> Where is the .iso to download?

https://github.com/dimkr/woof-CE/actions/runs/4215956249

peabee:
Current version seems to be 1.12.55 (51 from Salix):
https://voidlinux.pkgs.org/current/voidlinux-main-x86_64/gnumeric-1.12.55_1.x86_64.xbps.html
so 1.12.48 is a bit old - any reason not to use 55?

Both Abiword & Gnumeric have a "hidden" dependency on Adwaita (not included in S15Pup).......
this can be _largely_ satisfied with a link to Puppy Standard
however
there is a missing icon in both for a truncated menu strip dropdown for hidden items

dimkr:
> so 1.12.48 is a bit old - any reason not to use 55?

Yes. Because 1.12.48 is slightly older, a single petbuild works for all woof-CE build configurations that use the gnumeric petbuild. More shared code, less inconsistency between different Puppies, less work.

I think it's safe to assume that most Puppy users that use spreadsheets either need a basic spreadsheet application (so an older gnumeric is good enough) or install something more full featured like LibreOffice Calc.

> Both Abiword & Gnumeric have a "hidden" dependency on Adwaita (not included in S15Pup).......

They work just fine in dpup, without the full Adwaita icon theme, which is very big. The only Adwaita icons included in the build are symbolic arrow icons (https://github.com/puppylinux-woof-CE/woof-CE/blob/b786a827cff425d7a22b807624ed6fd11969c0d2/woof-code/packages-templates/gtk3_FIXUPHACK#L10).

Regarding inconsistency in appearance - that's because s15pup uses GTK+ 2 almost universally, and the GTK+ 2 theme it uses doesn't have a GTK+ 3 equivalent, hence the fallback to Adwaita. Every time you replace an ancient .pet package with something (more) modern, you get a GTK+ 3 application. dpup doesn't have this inconsistency problem because it only ships themes that support both GTK+ 2 and GTK+ 3. (And it uses GTK+ 3 almost universally, demonstrating that Puppy can live on without all those ancient applications.)

It's up to you whether or not to merge this PR: take it or leave it. I'm not interested in maintaining GTK+ 2 compatible petbuilds and dealing with the maintenance burden, because the list of actively developed applications that use GTK+ 2 is very small. Thanks to Flatpak (which allows the developer of an application to bundle libraries with it), distros will probably drop it from their repos soon (https://www.phoronix.com/news/RHEL-10-No-GTK2) and that's when Puppy's journey with GTK+ 2 is going to end.

PR #3928, opened 2023-02-20
Revert "enable the gnumeric petbuild in s15pup (fixes #3912) (#3920)"

### LINKED ISSUE #3912: gnumeric petbuild has unsatisfiable dependency for Slackware & Gtk2

dimkr:


dimkr:
Oops, reverted the wrong commit. Reverted the revert.

PR #3933, opened 2023-02-20
Replace the sfwbar logo with the Puppy logo

dimkr:
puplogo


LBCrion:
Just a heads up. In the future versions sfwbar will render the fallback icon using CSS foreground colour. This is to make it play nice with light and dark themes.

dimkr:
> Just a heads up. In the future versions sfwbar will render the fallback icon using CSS foreground colour. This is to make it play nice with light and dark themes.

Sounds cool and even better!

PR #3936, opened 2023-02-23
Default gtk-3.0 settings

peabee:


woof-code/rootfs-skeleton/root/.config/gtk-3.0/settings.ini
```diff
@@ -0,0 +1,6 @@
+[Settings]
```
dimkr, line 1: Add a copy of this file at /etc/gtk-3.0/settings.ini - applications that run as spot don't obey a configuration file in root's home directory

peabee: done

PR #3940, opened 2023-02-24
Mount a tmpfs on /run, like other distros do

dimkr:
Previously denied in #2080, because at least one package seems to be buggy: it stores files in /run and expects them to persist. That's not why /run exists, it's supposed to store runtime state and not state in general. When it's persistent, it pollutes save files, slows down save2flash and causes extra writing.


gyrog:
Obvious question, what has changed since #2080?

I'm not familiar with other Linux distributions, but it seems reasonable that /run should not be expected to persist.

Maybe there needs to be a dicussion about which files are considered as appropriate to be persistent.
I'm not sure how to bring such a discussion to the attention of significant numbers of Puppy users.

One of the advantages of the "indirect" save mechanisms like pupmode=13, is that it provides an oportunity to screen out the "rubbish" from being written to the "save".
It would be nice to have an agreed list, so all save mechanisms could be consisent.
Could even do a "cleanup" process for pupmode=12 in init, based on the list.

dimkr:
> Obvious question, what has changed since https://github.com/puppylinux-woof-CE/woof-CE/pull/2080?

Time

> Maybe there needs to be a dicussion about which files are considered as appropriate to be persistent.

I don't see any room for discussion, /run is supposed to be non-persistent and anything that stores stuff there and expects it to persist should be fixed, not the other way around. Somebody, I don't know who, when or why, built a popular .pet or .sfs that puts stores persistent stuff in /run, and that's why Puppy is pretty much the only distro to write /run to disk (only under PUPMODE 12, but still a bad idea).

gyrog:
So, the code that expects persistence for files in /run, could still be out there.
This may cause "complaints" that stuff is getting lost.

That's why it would be good to have a discussion about what directories should be kept, and which can be ignored, (not just about /run) . So that the Puppy community in general will not produce code that breaks the agreed lists.

But I do agree that the offending code needs to be fixed.
The difficulty is in identifying it.
Of course implementing this would be one way of finding it.
Unfortunately the user who would get stung, is probably not the person who is responsible for the problem.

dimkr:
> The difficulty is in identifying it.

The easiest way to identify it is to align Puppy with all other distros by making /run a tmpfs, and wait 😸

PR #3952, opened 2023-02-25
Update firewall_ng to allow correct Cancel and Apply button function. 

Overdrive5:
This change will not turn on firewall when user cancels out of the dialog box. My testing shows when "Cancel" is pressed, the original firewall state is preserved and when "Apply" is pressed the new settings are applied and the firewall is restarted.

### DIFF
--- a/woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
+++ b/woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
@@ -1450,6 +1450,7 @@ fi
 
 [ "$1" = "enable" -a "$MAIN" = "false" ] && exit
 
+
 # run it!
 gtkdialog-splash -bg yellow -timeout 2 -text "Starting firewall" &
 urxvt -bg yellow -fg black -geometry 60X10+1+1 --hold -e /etc/init.d/rc.firewall start &
--- a/woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
+++ b/woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
@@ -1448,7 +1448,7 @@ else
 	exit
 fi
 
-[ "$MAIN" = "false" ] && exit
+[ "$1" = "enable" -a "$MAIN" = "false" ] && exit
 
 # run it!
 gtkdialog-splash -bg yellow -timeout 2 -text "Starting firewall" &
--- a/woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
+++ b/woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
@@ -1448,7 +1448,7 @@ else
 	exit
 fi
 
-[ "$1" = "enable" ] && exit
+["$MAIN" = "false" ] && exit
 
 # run it!
 gtkdialog-splash -bg yellow -timeout 2 -text "Starting firewall" &
--- a/woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
+++ b/woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
@@ -1448,7 +1448,7 @@ else
 	exit
 fi
 
-["$MAIN" = "false" ] && exit
+[ "$MAIN" = "false" ] && exit
 
 # run it!
 gtkdialog-splash -bg yellow -timeout 2 -text "Starting firewall" &


woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
```diff
@@ -1448,7 +1448,7 @@ else
 	exit
 fi
 
-[ "$1" = "enable" ] && exit
+["$MAIN" = "false" ] && exit
```
dimkr: Syntax error, missing space after `[`

woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
```diff
@@ -1448,7 +1448,7 @@ else
 	exit
```
dimkr, line 1448: The commit message, "Update firewall_ng", doesn't explain what this does and how it works

woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
```diff
@@ -1448,7 +1448,7 @@ else
...
 fi
 
-[ "$1" = "enable" ] && exit
```
dimkr, line 1451: This PR breaks this line: https://github.com/puppylinux-woof-CE/woof-CE/blob/28f7424eef841fb3591b9c989c7d5bfd25153128/woof-distro/x86_64/debian/bullseye64/_00build.conf#L132

How about `[ "$1" = "enable" -o "$MAIN" = "false" ] && exit` instead?

woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
Overdrive5: I have updated the commit message to indicate: Update firewall_ng to allow correct Cancel and Apply button function. Is this sufficient or what would you advise?

woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
Overdrive5: Yes, I now see that. However this line 1451 statement did in fact work correctly during my testing. However in the quest for proper language syntax, I will correct it.

woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
Overdrive5: Hmm, let me study this further....

Overdrive5: Well...I'm not fluent enough with this build system (or github for that matter) to understand what is going on here. Could it have been from the missing space in line 1451? I have corrected that now in my patch1 fork.

Overdrive5: In regards to using the `[ "$1" = "enable" -o "$MAIN" = "false" ] && exit` statement. It would allow the "enable" test to exit before restarting the firewall. I am thinking we would only want to exit on a "false" test on this line and allow "enable" to continue further down the source code. No?

woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
Overdrive5: Interesting, when I add the extra space on the line it breaks the test function. Will have to study this further,

woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
dimkr: > I am thinking we would only want to exit on a "false" test on this line and allow "enable" to continue further down the source code. No?

One line with two condition with OR between them or two separate conditions in two lines - doesn't matter. But don't break support for `firewall_ng enable`, things use it.

dimkr:
This PR doesn't work, from what I see. The problem is solved only temporarily. When `MAIN=false`, it leaves behind /etc/init.d/firewall with execution permissions. When the system is rebooted, the firewall is enabled again.

EDIT: this seems to work, but I only enabled and disabled the firewall completely - still haven't tested other features

```
--- a/woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
+++ b/woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
@@ -1432,14 +1432,14 @@ if [ -f "$TMPFW" ];then
        [ $state = 1 -a "$1" != "enable" ] && /etc/init.d/rc.firewall stop && sleep 1
        echo "copying firewall"
        cp -af "$TMPFW" /etc/init.d/rc.firewall
-       chmod 755 /etc/init.d/rc.firewall
+       [ "$MAIN" = "true" ] && chmod 755 /etc/init.d/rc.firewall || chmod 644 /etc/init.d/rc.firewall
        rm -f "$TMPFW"
 else
        echo "Something went wrong"
        exit
 fi
 
-[ "$1" = "enable" ] && exit
+[ "$1" = "enable" -o "$MAIN" = "false" ] && exit
 
 # run it!
 gtkdialog-splash -bg yellow -timeout 2 -text "Starting firewall" &
```

EDIT 2: yes, other use cases are broken. Try to enable the firewall while allowing incoming SSH traffic. The firewall is disabled, but if you run rc.firewall you get a firewall that blocks everything except SSH (as you expect). The script is not executable, so the firewall is disabled. This happens because `MAIN=true` only if the checkbox at the top is selected while others aren't. That `chmod 755` needs to happen in any case where the firewall is enabled, not only when the firewall is enabled without exceptions.

dimkr:
This patch seems to work. The firewall is disabled when it should be disabled, and ports can be blacklisted.

```
--- a/woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
+++ b/woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
@@ -1429,10 +1429,10 @@ echo "Load rules for mangle table ..."
 EOF_REST
 
 if [ -f "$TMPFW" ];then
-       [ $state = 1 -a "$1" != "enable" ] && /etc/init.d/rc.firewall stop && sleep 1
+       [ $state = 1 -a "$1" != "enable" ] && sh /etc/init.d/rc.firewall stop && sleep 1
        echo "copying firewall"
        cp -af "$TMPFW" /etc/init.d/rc.firewall
-       chmod 755 /etc/init.d/rc.firewall
+       [ "$MAIN" = "true" -o "$CONFIG_SSH" = "true" -o "$CONFIG_CUPS" = "true" -o "$CONFIG_SAMBA" = "true" -o "$CONFIG_SAMBA_CLIENT" = "true" -o "$CONFIG_DLNA" = "true" -o "$CONFIG_NTP" = "true" -o "$CONFIG_FTP" = "true" -o "$CONFIG_HTTP" = "true" -o "$CONFIG_DNS" = "true" -o "$CONFIG_DHCP" = "true" -o "$CONFIG_NFS" = "true" ] && chmod 755 /etc/init.d/rc.firewall || chmod 644 /etc/init.d/rc.firewall
        rm -f "$TMPFW"
 else
        echo "Something went wrong"
@@ -1440,6 +1440,7 @@ else
 fi
 
 [ "$1" = "enable" ] && exit
+[ "$MAIN" = "false" -a "$CONFIG_SSH" = "false" -a "$CONFIG_CUPS" = "false" -a "$CONFIG_SAMBA" = "false" -a "$CONFIG_SAMBA_CLIENT" = "false" -a "$CONFIG_DLNA" = "false" -a "$CONFIG_NTP" = "false" -a "$CONFIG_FTP" = "false" -a "$CONFIG_HTTP" = "false" -a "$CONFIG_DNS" = "false" -a "$CONFIG_DHCP" = "false" -a "$CONFIG_NFS" = "false" ] && exit
 
 # run it!
 gtkdialog-splash -bg yellow -timeout 2 -text "Starting firewall" &
```

woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
Overdrive5: OK, understood.  `[ "$1" = "enable" -a "$MAIN" = "false" ] && exit` still keeps Cancel and Apply buttons functionality intact. I will modify my patch1 source. See if that builds?

woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
Overdrive5: Revised line 1451 with correct syntax.

dimkr: See https://github.com/puppylinux-woof-CE/woof-CE/pull/3952#issuecomment-1445280171 and https://github.com/puppylinux-woof-CE/woof-CE/pull/3952#issuecomment-1445284024

woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
Overdrive5: I realize I am using a AND instead of a OR here. Using a OR here breaks code continuation to complete firewall restart.

dimkr: But that's what you want to do. `firewall_ng enable` is used to enable the firewall without starting it (hence, must exit at this point), or if `MAIN=false`, the user chose to disable the firewall (hence, must exit at this point).

Overdrive5: I'm not quite sure I understand. So using a AND here is acceptable? Otherwise like I said, using a OR here breaks firewall restart.

woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
Overdrive5: Yeah, for some reason I did not see these when I commented afterwards.

woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
dimkr: With AND, the condition is _"asked to exit without starting the firewall" AND "the firewall is disabled"_ - this condition is always false when you start firewall_ng via the menu entry (because the left part is false).

With OR, the condition is _"asked to exit without starting the firewall" OR "the firewall is disabled"_, which makes sense - you don't want to enable the firewall in either case.

woof-code/rootfs-packages/firewall_ng/usr/sbin/firewall_ng
Overdrive5: I tried those [https://github.com/puppylinux-woof-CE/woof-CE/pull/3952#issuecomment-1445280171 and https://github.com/puppylinux-woof-CE/woof-CE/pull/3952#issuecomment-1445284024](url) and unfortunately it still breaks firewall restart. Maybe on a reboot it works but the original functionality restarted the firewall with no further user interaction. I would think this is the preferred end user experience.

dimkr: How do you restart the firewall?

Overdrive5: > How do you restart the firewall?

`[ "$1" = "enable" -a "$MAIN" = "false" ] && exit` allows the code to proceed to the `urxvt -bg yellow -fg black -geometry 60X10+1+1 --hold -e /etc/init.d/rc.firewall start &` line when Apply button is pressed. If the user presses Cancel, `[ "$1" = "enable" -a "$MAIN" = "false" ] && exit`  short circuit exits and never restarts the firewall. Nor does it save any changes either. So the firewall "existing state" is preserved. On/Off/disabled or otherwise.

Overdrive5: Unfortunately, Alien Bob's  original code still creates a rc.firewall file if the user unchecks the "When this button is checked a generic firewall is generated on no specific interface." and then presses Apply. I would think most users would just press Cancel if they don't want a firewall anyway.

dimkr: If I'm not mistaken (maybe my memory from this morning is no longer correct), `MAIN=true` and not `MAIN=false` if the firewall is enabled, so clicking Apply without changing anything restarts the firewall when the firewall is already enabled. `bash -x firewall_ng` shows this.

Overdrive5: I temporarily inserted `gtkdialog-splash -bg yellow -timeout 2 -text "Main: $MAIN"`  one line above `[ "$1" = "enable" -a "$MAIN" = "false" ] && exit` on my local testing copy and MAIN=false when I press Apply. Have not studied the rest of the code to trace out why this would be. But `[ "$1" = "enable" -a "$MAIN" = "false" ] && exit` does work correctly for me. Using it restarts the firewall when I press Apply. And Cancel does preserve the previously "existing state". So I think this is what we are trying to accomplish here unless I am missing something.

dimkr: Did you run firewall_ng with `firewall_ng enable`? If not, this condition doesn't do anything (because it's false).

Overdrive5: I ran it from the menu. if that is set this way.

Overdrive5: I just checked my local .jwmrc file and it is just using `firewall_ng`

Overdrive5: which file is executing the `firewall_ng enable` command? I checked welcome1stboot.

dimkr: https://github.com/puppylinux-woof-CE/woof-CE/blob/28f7424eef841fb3591b9c989c7d5bfd25153128/woof-distro/x86_64/debian/bullseye64/_00build.conf#L132

dimkr: > I just checked my local .jwmrc file and it is just using `firewall_ng`

Did you replace /usr/sbin/firewall_ng with the modified one from this PR?

Overdrive5: No, but I've been editing that file for testing.

Overdrive5: > https://github.com/puppylinux-woof-CE/woof-CE/blob/28f7424eef841fb3591b9c989c7d5bfd25153128/woof-distro/x86_64/debian/bullseye64/_00build.conf#L132

So there is no local app that is using this command?

dimkr: No.

dimkr: With the changes in this PR, the way it is now (6bb4ddf), I unticked this checkbox and clicked Apply:

apply

As I expect, the firewall is restarted, not disabled like it's supposed to. I'm running `bash -x firewall_ng`:

```
+ cp -af /tmp/firewall_gen395815 /etc/init.d/rc.firewall
+ chmod 755 /etc/init.d/rc.firewall
+ rm -f /tmp/firewall_gen395815
+ '[' '' = enable -a false = false ']'
+ gtkdialog-splash -bg yellow -timeout 2 -text 'Starting firewall'
+ p=395919
```

`'[' '' = enable -a false = false ']'` is `[ "$1" = "enable" -a "$MAIN" = "false" ]`. Yes, unticking the checkbox sets `MAIN=false`. But, as I said, there's no first argument (it's `firewall_ng`, not `firewall_ng enable`), therefore `$1` is `""`, an empty string. `"$1" = "enable"` is false, therefore `"$1" = "enable" -a "$MAIN" = "false"` is false (in English, "the first argument is 'enable' _and_ the checkbox is unticked") . Because the big condition is false, the script doesn't exit like it's supposed to (the `&& exit` part doesn't happen), and the script starts the firewall again.

Overdrive5: So what do you advise? 

I think the more important functionality is having the Cancel and Apply buttons working correctly from the menu I.E. `firewall_ng`

This `firewall_ng enable` is not included in the ISO but `firewall_ng` is. No?

Overdrive5: using 'bash -x firewall_ng' and:

`
if [ -f "$TMPFW" ];then

	[ $state = 1 -a "$1" != "enable" ] && /etc/init.d/rc.firewall stop && sleep 1

	echo "copying firewall"

	cp -af "$TMPFW" /etc/init.d/rc.firewall

	chmod 755 /etc/init.d/rc.firewall

	rm -f "$TMPFW"

else

	echo "Something went wrong"

	exit

fi`


using apply with:
`[ "$1" = "enable" -o "$MAIN" = "false" ] && exit`

Does not restart the firewall.

using:

`[ "$1" = "enable" -a "$MAIN" = "false" ] && exit`

Does restart the firewall.

>With the changes in this PR, the way it is now (https://github.com/puppylinux-woof-CE/woof-CE/commit/6bb4ddfe2519c31b0d0967d957b1571d2e07cf57), I unticked this checkbox and clicked Apply:

apply

As I expect, the firewall is restarted, not disabled like it's supposed to. I'm running bash -x firewall_ng:

+ cp -af /tmp/firewall_gen395815 /etc/init.d/rc.firewall
+ chmod 755 /etc/init.d/rc.firewall
+ rm -f /tmp/firewall_gen395815
+ '[' '' = enable -a false = false ']'
+ gtkdialog-splash -bg yellow -timeout 2 -text 'Starting firewall'
+ p=395919

'[' '' = enable -a false = false ']' is [ "$1" = "enable" -a "$MAIN" = "false" ]. Yes, unticking the checkbox sets MAIN=false. But, as I said, there's no first argument (it's firewall_ng, not firewall_ng enable), therefore $1 is "", an empty string. "$1" = "enable" is false, therefore "$1" = "enable" -a "$MAIN" = "false" is false (in English, "the first argument is 'enable' and the checkbox is unticked") . Because the big condition is false, the script doesn't exit like it's supposed to (the && exit part doesn't happen), and the script starts the firewall again.<

================
You will find if you run the original firewall_ng from the main repo without our mods, the untick/apply issue already exists. So it is not from this mod. It is a bug (or feature) from Alien Bob's code. I don't propose to try and resolve that. More importantly, I think fixing the apply issue is more critical at this point, No?

Update:  I think fixing the apply restarting the firewall issue is more critical at this point, No?

dimkr: > More importantly, I think fixing the apply issue is more critical at this point, No?

> Update: I think fixing the apply restarting the firewall issue is more critical at this point, No?

At this point, I'm longer sure what you mean by "the apply issue" and "restarting the firewall issue".

With the patch from https://github.com/puppylinux-woof-CE/woof-CE/pull/3952#issuecomment-1445284024, if you click Apply when the firewall is already enabled, it gets restarted, and if you click Apply after the checkbox is unticked, it's disabled. It solves the two issues I see, while this PR doesn't and even breaks `firewall_ng enable`.

Overdrive5: OK, I just tried this code. For some reason I did not see this until well after we started discussing the AND / OR snippet several hours later. Yes, It seems to have solved both issues. My humble apologies for not trying this earlier. I will close this PR as you have the better solution.

PR #3953, opened 2023-02-26
 Increase pfind dialog box size to show correctly

Overdrive5:
Increase pfind dialog box size to show correctly. 450x700 is to small. 750x950 shows the entire app without having to resize every time.

### DIFF
--- a/woof-code/rootfs-packages/pfind/usr/local/pfind/pfindrc
+++ b/woof-code/rootfs-packages/pfind/usr/local/pfind/pfindrc
@@ -19,7 +19,7 @@ export PATH_PUPPY='/bin [OR] /etc [OR] /lib [OR] /lib64 [OR] /opt [OR] /sbin [OR
 export PATH_MNT='/mnt/ [OR] /root'
 
 #geometry
-export HEIGHT="450"
-export WIDTH="700"
+export HEIGHT="750"
+export WIDTH="950"
 export X=""
 export Y=""


Overdrive5:
450x700:
pfind450x700
750x950:
pfind750x950

zigbert:
It should not that big by default in case Puppy is running on small screen.

If size should be enlarged, it should be only on screens that are big enough.
xrandr could give information for X. But these days Puppy is also running on Wayland... ?

xrandr | grep '*' | head -n1

The existing solution for this issue is found in the 'File' menu. - Quit and save geometry.

Overdrive5:
Oh!! I see!. Never observed that before. 

Always just used the standard mouse 'x' close method.

dimkr:
> But these days Puppy is also running on Wayland... ?

Yes, xrandr won't work.

Overdrive5:
Maybe use?: xdpyinfo | grep dimensions

Overdrive5:
I find my scrollbars disappear at 679 x 936

Maybe do a  xdpyinfo test if it exceeds these dimensions then set accordingly?

Overdrive5:
xdpyinfo  | grep -oP 'dimensions:\s+\K\S+'

zigbert:
I am unfamiliar with 'xdpyinfo', and don't know if this would work outside X.

It would still an option to include a test for both xrandr and xdpyinfo. If it doesn't give any valuable feedback, we just go back to the default values.

This test must not override the default settings in $HOME/.pfind/pfindrc. The rc file could have blank values at first startup and the test could save its new value for later use.

Overdrive5:
I found the method of getting each dimension separately.

Width: xdpyinfo | awk '/dimensions/ {print $2}' | cut -d 'x' -f1
Height: xdpyinfo | awk '/dimensions/ {print $2}' | cut -d 'x' -f2

Overdrive5:
So we could confirm both width exceed 936 and height exceed 679 then set pfindrc to those dimensions.

Overdrive5:
And I checked pup-431.iso and xdpyinfo is located in /usr/bin so it should be available for most recent distributions

dimkr:
xdpyinfo won't work in a Puppy that uses Wayland. Make sure you have some fallback if xdpyinfo fails (or if it's missing).

PR #3959, opened 2023-03-04
Start Firefox "centered" to keep previous position

Overdrive5:
Start Firefox "centered" to help it remember previous position when "Open previous windows and tabs" is set in Firefox settings


dimkr:
8532db2 should fix this.

Overdrive5:
Yep, OK, closing.

PR #3965, opened 2023-03-06
resolves#3961

okwizard:
USB installers like Rufus add their own EFI directory, installers that do not, like Unetbootin, cannot create a UEFI USB without the EFI directory in / of the ISO. Solution is to add the EFI directory to the ISO. This will improve the success rate for new users wanting to try and/or install Puppy.


dimkr:
There's nothing that copies these files into the build, and these are binary files of unknown origins, built from unavailable source code. This doesn't do anything, and probably illegal.

okwizard:
Thanks Dima, new at this and doing my best to figure out how to make it work, and not very successfully..  The files came from a EFI directory in a USB created with Stickpup, so think they already exist somewhere in Puppy.  Should I just delete the PR?

Gary


   On Monday, March 6, 2023, 02:08:46 PM CST, Dima Krasner ***@***.***> wrote:  
 
 


There's nothing that copies these files into the build, and these are binary files of unknown origins, built from unavailable source code. This doesn't do anything, and probably illegal.

dimkr:
Look in woof-code/support/mk_iso.sh, that's the script that builds the Puppy ISO. If I understand correctly, all these files are already present, but just need to be copied to a second location under the ISO, rather than inside the EFI image.

peabee:
Copied? or moved? or linked?

dimkr:
Copied from the EFI image to the ISO root, if I understand the issue correctly.

peabee:
I can only see in the .iso:
.iso/boot/efi.img
6.2MB

I don't see the .efi files @okwizard has uploaded but inside +boot+efi.img/EFI/boot is:
bootx64.efi  grubx64.efi  mmx64.efi  puppy.cer

dimkr:
If I understand @okwizard correctly, these files need to be copied to a directory named EFI under the ISO.

okwizard:
Both /boot/efi.img and the directory /EFI (and its contents) have to be included in the ISO.  Attached is the /EFI directory that was taken from a Stickpup USB.  This same /EFI directory is also generated by Frugalpup when it creates a UEFI boot USB.

Thanks
Gary 



   On Monday, March 6, 2023 at 11:56:11 PM CST, Dima Krasner ***@***.***> wrote:  
 
 


Look in woof-code/support/mk_iso.sh, that's the script that builds the Puppy ISO. If I understand correctly, all these files are already present, but just need to be copied to a second location under the ISO, rather than inside the EFI image.

okwizard:
@peabee @dimkr 
The files I originally tested and uploaded came from a /EFI directory in a F96-CE USB created with Stickpup, a total of 7 files.  These look to include both 32 and 64 bit.  After a lot more testing it appears just unpacking the contents of /boot/efi.img and putting those 3 files in a /EFI directory of the ISO gives the same results.

Hope this helps.
Gary

dimkr:
Why not fix this cleanly?

Just pass the ISO root to this function as an additional argument and `cp -r /tmp/efi_img/EFI $4/` (or something similar). It's a two line fix that doesn't add binary files of unknown origin to this repo.

https://github.com/puppylinux-woof-CE/woof-CE/blob/8532db20e42cf8f961d440fa34a11cbd790fe80f/woof-code/support/mk_iso.sh#L496

okwizard:
@dimkr 
"Why not fix this cleanly?"
Yes, this is what I'm suggesting.  Keep in mind my coding skills are poor, but mk_iso.sh should be easy to modify to add the ISO /EFI directory that contains the same files that are in /boot/efi.img.

Thanks
Gary

PR #3992, opened 2023-04-02
Drop bionic64 builds (Ubuntu 18.04 is EOL in May)

dimkr:


peabee:
Why not disable it rather than delete - then it could be run if needed?
And why not wait until May?

dimkr:
This only removes the scheduled builds (= same as disabling them). The configuration files are still there, anyone who wants to build bionic64 can still do that.

I'll merge this PR later this month, once 18.04 goes EOL. If the configuration files for bionic64 break, I won't fix them.

peabee:
EOL only "really" happens when the repos get deleted from "archive" and are only available at "old-releases" and this can be some time after May....
https://old-releases.ubuntu.com/ubuntu/dists/
which is when PPM stops working without a "patch"

dimkr:
Define "EOL". You can keep using Ubuntu 18.04, but you have zero support, zero bug fixes and zero security updates.

And again, this PR only removes the monthly bionic64 builds, to save space and reduce the number of build hours per month. I expect the number of users interested in an "up-to-date" bionic64 build (soon, just an illusion) to drop, so it's a waste to keep it. Even if this PR is merged, you can still build bionic64 using build.yml or release.yml or a local woof-CE build: the configuration files are still there and this PR doesn't break bionic64.

dimkr:
I'll write this here one more time in case this wasn't 100% clear: it's still possible to do bionic64 builds. This PR only disables the monthly bionic64 builds, because they're much more likely to break now, nobody maintains them as far as I see, and a Puppy based on 5 years old packages is not exactly the top recommendation for new users. Anyone who wants to continue building and publishing bionic64 releases is free to do so, but I don't want to get emails about them breaking or waste storage space on EOL packages.

PR #3996, opened 2023-04-05
profile: Added debian library path

rizalmart:
Newly built debian-based puppy unable to detect library installed at /usr/lib/<arch>-linux-gnu. This will fix the problem

### DIFF
--- a/woof-code/rootfs-skeleton/etc/profile
+++ b/woof-code/rootfs-skeleton/etc/profile
@@ -3,11 +3,13 @@
 #110804 fix double-login when exit from X, allow /etc/profile to complete. see also "echo -n '# '" in /usr/bin/xwin at exit.
 
 PATH="/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/root/my-applications/bin:/usr/games"
+
 if [ -f /lib64/libc.so.6 ] ; then #slackware64
-	LD_LIBRARY_PATH="/lib64:/usr/lib64:/root/my-applications/lib:/usr/local/lib"
+	LD_LIBRARY_PATH="/lib64:/usr/lib64:/usr/lib/x86_64-linux-gnu:/root/my-applications/lib:/usr/local/lib"
 else
-	LD_LIBRARY_PATH="/lib:/usr/lib:/root/my-applications/lib:/usr/local/lib"
+	LD_LIBRARY_PATH="/lib:/usr/lib:/usr/lib/i386-linux-gnu:/root/my-applications/lib:/usr/local/lib"
 fi
+
 export PATH LD_LIBRARY_PATH
 
 export GDK_USE_XFT=1 #for gtk...


techrockedge:
I vote to merge this commit! I have had to manually do this exact export manually several times recently and this does the fix.

dimkr:
This is a "dirty" solution for a problem that doesn't exist. The clean solution is to use the upstream distro's `/etc/ld.so.conf{,.d}` instead of forcing Puppy's one, and use the upstream distro's file system layout (usrmerge or not).

dpup uses Debian's library path and the LD_LIBRARY_PATH value set in /etc/profile doesn't do anything really. 3builddistro doesn't force Puppy's bad ld.so.conf if `USR_SYMLINKS=yes`, otherwise many applications you install are broken and multilib doesn't work (it works thanks to files in /etc/ld.so.conf.d).

https://github.com/puppylinux-woof-CE/woof-CE/blob/f0755e696904f10c6103af897a493db5760d6337/woof-code/3builddistro#L845

In dpup, `/lib64/ld-linux-x86-64.so.2 --help` shows that the search path used is a combination of what /etc/ld.so.conf (plus /etc/ld.so.conf.d/* - sourced by ld.so.conf) says plus what LD_LIBRARY_PATH says (/lib, /usr/lib, etc' are meaningless because libraries are under the `x86_64-linux-gnu` subdirectory).

techrockedge:
`The clean solution is to use the upstream distro's /etc/ld.so.conf{,.d} instead of forcing Puppy's one, and use the upstream distro's file system layout (usrmerge or not).
`
That is the best solution. I have manually created files in  /etc/ld.so.conf{,.d} to test this on _F96-CE_ when compiling certain packages. 

Though I have also manually added the lib path to LD_LIBRARY_PATH in /etc/profile on Fossapup64's for a quick fix.

peabee:
Slackware does not have ld.so.conf in glibc - it is generated by:
https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-code/3builddistro#L846

which overwrites:
https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-code/rootfs-skeleton/etc/ld.so.conf

dimkr:
Puppy's ld.so.conf has no advantage over Slackware's ld.so.conf, using Slackware's one can only improve compatibility with Slackware, and maybe it can reduce the number of directories the loader needs to check when you run an executable.

I'd delete Puppy's ld.so.conf and remove this condition (use the `if` flow in all builds and drop the `else`):

https://github.com/puppylinux-woof-CE/woof-CE/blob/f0755e696904f10c6103af897a493db5760d6337/woof-code/3builddistro#L843

peabee:
**Slackware does not have ld.so.conf in glibc**

dimkr:
Maybe not in glibc, maybe somewhere else 🤷🏿 

And if it doesn't have on at all (just uses the loader's default search path), that's also a valid option for a Slackware-based Puppy.

peabee:
Nowhere else in any of the packages currently included in the build.........

PR #3998, opened 2023-04-10
Update sfwbar to a76d4b2

dimkr:
@01micko Are you still here with us?

This breaks some of the "tray applets" 😿


LBCrion:
Is this the SNI tray items or the sfwbar widgets that broke? If it's the layer,  I expect its due to the tighter type checking in the latest version.

dimkr:
Hey @LBCrion, I still haven't looked at everything but https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-code/rootfs-petbuilds/sfwbar/root/.config/sfwbar/pupnet.widget is broken (no icon is displayed) and https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-code/rootfs-petbuilds/sfwbar/root/.config/sfwbar/pupusage.widget now has a syntax error (the `If` condition is formatted into the SVG).

01micko:
I'll download latest vanilla 11 and take a look.

EDIT: just tried latest and the configs at https://github.com/01micko/sfwbar-configs/tree/main/configs/sfwbar work, but icons are smaller so need to investigate that. In saying that, they are all consistently smaller so should be a simple fix.

EDIT0: @LBCrion it seems `min-width` CSS for button and image behave differently using `px`. To achieve similar onscreen size I'm using different values.

Example: 

bar


Taskbar CSS:
```
button#taskbar_group_normal image , button#taskbar_group_active image {
  min-width: 20px;
  min-height: 20px;
}
```

Typical image CSS: (each in separate configs)
```
layout {
  image {
    value = svg
    interval = 2000
    tooltip = Str(CPUTemp,0) + "°C CPU temperature"
    action = "lxtask"
    css = "* { min-height: 30px; min-width: 30px; padding: 4px; }"
  }
}
```

dimkr:
> I'll download latest vanilla 11 and take a look.

This PR is still not merged. https://github.com/dimkr/woof-CE/actions/runs/4655234330 includes this PR.

01micko:
Yeah realised after I posted. :stuck_out_tongue_closed_eyes:

LBCrion:
@01micko, padding, border or margin around the icons may behave differently now (more consistent with the gtk standard).

Also,  you may want to look at the new modules for widgets such as pupnet and pulse. Although for the network module,  you may want to test with the latest git,  beta11 has a bug in it that may cause a segfault.

dimkr:
Merging, let's break what needs to break and fix later. A new development build will be out on Saturday and it will make debugging easier.

PR #4026, opened 2023-04-25
Use xdg-open from xdg-utils if not using ROX-Filer

dimkr:


peabee:
How will this affect builds which already use PcManFM or Thunar for example and are working fine with the "puppy xdg-open"?? Has it been tested with a ydrv patch?

dimkr:
It will only affect a Puppy without preinstalled ROX-Filer, like other previous changes in 3builddistro needed to support other file managers. It's safe.

dimkr:
@peabee Your builds with file managers other than ROX-Filer - do they also include ROX-Filer in the main SFS?

peabee:
Yes

dimkr:
> Yes

To clarify, the "Puppy" xdg-open and /usr/local/bin/default* are hacks needed only because ROX-Filer doesn't respect mimeapps.list like other file managers.

Puppy works unlike all other distros: mimeapps.list in rootfs-skeleton makes /usr/local/bin/default* the default handler for many file types, ROX-Filer uses these scripts too via its internal and non-standard (= not xdg-mime) way of configuring handlers, and other file managers or browsers cannot set the default handler globally. The "single source of truth" in Puppy is /usr/local/bin/default*, not xdg-mime, and that breaks so many things for applications other than ROX-Filer.

In a Puppy that doesn't have ROX-Filer at all, these hacks conflict with the file manager's settings (for example, defaulttexteditor still points to geany while the default handler for .c files is changed to something else) and introduce issues (for example, click a gemini:// link in the browser after you install a Gemini client - it won't work).

Therefore, if ROX-Filer is **replaced** with a different file manager (ROX-Filer is **not** preinstalled alongside another file manager), these hacks should be dropped.

This PR is a step beyond #3500: the former PR made the file manager responsible for managing handlers for different file types (= not /usr/local/bin/default*), the way it is in other distros, and the current PR makes xdg-open respect the user's default application of choice. /usr/local/bin/default* are simulated using xdg-mime for backwards-compatibility with Puppy stuff that assumes ROX-Filer.

TL;DR nothing should change as long as you have ROX-Filer in rootfs-complete. But the tradition of using ROX-Filer shouldn't prevent woof-CE from supporting other file managers (especially those that work under Wayland) as **replacements** for ROX-Filer.

PR #4044, opened 2023-05-09
Fix homebank version missing from debian mirror

fejnartal:
homebank-5.0.9 tarball is no longer available from debian mirror.
Newer tarballs 5.2.2 and 5.6.2 fail to build because libsoup is missing.
Download homebank 5.0.9 from launchpad

Failed runs:
- [Using version missing from debian mirror](https://github.com/fejnartal/woof-CE/actions/runs/4907934763)
- [Using newest version from debian mirror](https://github.com/fejnartal/woof-CE/actions/runs/4908264230)

Successful run:
- [Using code from this PR](https://github.com/fejnartal/woof-CE/actions/runs/4909004531)


woof-code/rootfs-petbuilds/homebank/sha256.sum
```diff
@@ -1 +1 @@
-1eece7c11b3286a53385386fb6d4a39902bffaba027cdd5a284aee58f4aa20d8  homebank-5.0.9.tar.gz
+fb9b59df8fda320f514815ce14e47b705b158ffc744e8c9c69a23edc59f41732  homebank-5.0.9.tar.gz
```
dimkr: Why not http://mirrors.ircam.fr/pub/raspbian/raspbian/pool/main/h/homebank/homebank_5.0.9.orig.tar.gz or some other mirror that points to an identical file?

fejnartal: That's a great point.
I am not very well versed in public package mirrors so i defaulted to "some" homebank's source repository.
I wasn't even sure whether that's the official one or not.

I will amend this PR to use another mirror with the same sha256 as the original one.

Thanks! ☺👍

dimkr: It doesn't really matter if this is an official mirror or not, as long as the file hash is the same. That's why we have those hashes: to protect against accidental change of the source code we're building (and for security).

PR #4078, opened 2023-06-12
Update DISTRO_PKGS_SPECS-debian-sid: added zlib1g and libgcc-s1

rizalmart:
Without zlib1g and libgcc-s1, some core-apps of debian-sid woof builds are broken


dimkr:
`deps:yes` takes care of adding these dependencies

rizalmart:
> `deps:yes` takes care of adding these dependencies

However it does not work. zlib1g and libgccc-s1 went to devx instead of main puppy sfs. Also acl builtin files contents accidentally deletes libacl files which broke the pkg command

dimkr:
The sid64 configuration, the way it is now in testing, does not produce a working devx and it's not supposed to. The user is supposed to install development packages using apt or Synaptic.

PR #4104, opened 2023-07-02
Allow PUPMODE 66 with exfat

dimkr:
FAT32 is very portable but also very limited. exFAT is relatively portable, doesn't suffer from the same limitations and has no journaling, making it a good choice for flash drives when compatibility is a concern.

However, exFAT is not perfect and shares some of the limitations of FAT32: for example, it doesn't support symlinks. If FAT32 supports both save files and PUPMODE 66 to work around these limitations, exFAT should support both options, too.


gyrog:
Sorry, I don't access github very often these days.

PR #4105, opened 2023-07-03
Exclude apt package lists from the PUPMODE 66 archive

dimkr:
They're huge, >100 MB.


gyrog (approved):
I guess that means that the apt lists will need to be downloaded every time apt is used to install.
But then it's not recommended to install lots of software in pupmode=66.

PR #4125, opened 2023-07-30
Update release.yml

clocke125:


dimkr:
@clocke125 You don't need to modify the file, you can specify the parameters when you trigger the job 

input

PR #4141, opened 2023-09-14
Add kernel-6.1 32bit

peabee:


kernel-kit/configs_x86/DOTconfig-6.1.52-x86
```diff
@@ -0,0 +1,8012 @@
...
+# CONFIG_OCFS2_FS is not set
+CONFIG_BTRFS_FS=y
+CONFIG_BTRFS_FS_POSIX_ACL=y
+# CONFIG_BTRFS_FS_CHECK_INTEGRITY is not set
+# CONFIG_BTRFS_FS_RUN_SANITY_TESTS is not set
+# CONFIG_BTRFS_DEBUG is not set
+# CONFIG_BTRFS_ASSERT is not set
+# CONFIG_BTRFS_FS_REF_VERIFY is not set
+CONFIG_NILFS2_FS=m
+CONFIG_F2FS_FS=m
```
dimkr: This kernel won't boot from f2fs partitions

kernel-kit/configs_x86/DOTconfig-6.1.52-x86
```diff
@@ -0,0 +1,8012 @@
+#
```
dimkr, line 1: No CONFIG_VMD, needed for some NVMe drives (not sure if this even exists on x86)

kernel-kit/configs_x86/DOTconfig-6.1.52-x86
```diff
@@ -0,0 +1,8012 @@
...
+# CONFIG_BEFS_FS is not set
+# CONFIG_BFS_FS is not set
+# CONFIG_EFS_FS is not set
+# CONFIG_CRAMFS is not set
+CONFIG_SQUASHFS=y
+# CONFIG_SQUASHFS_FILE_CACHE is not set
+CONFIG_SQUASHFS_FILE_DIRECT=y
+# CONFIG_SQUASHFS_DECOMP_SINGLE is not set
+# CONFIG_SQUASHFS_DECOMP_MULTI is not set
+CONFIG_SQUASHFS_DECOMP_MULTI_PERCPU=y
```
dimkr: As far as I know, CONFIG_SQUASHFS_DECOMP_MULTI results in absolutely terrible performance if you have <=2 CPU cores, especially if you use xz compressed SFSs

kernel-kit/6.1.x-x86-build.conf
```diff
@@ -0,0 +1,129 @@
...
+## if you have a multicore processor you can set this var
+## don't set if you have a single core! >> cooked machine 
+## DO NOT set it to 0 (zero) >> cooked machine
+#JOBS=-j6
+
+### squashfs compression ###
+## unset this for the default of your mksquashfs binary
+#COMP="-comp gzip"
+#COMP="-comp xz"
+COMP="-comp xz -b 512K"
```
dimkr, line 77: Some users (netbooks without much L1/2/3 cache?) say they can't use a Puppy that uses xz with block size larger than 256K

kernel-kit/configs_x86/DOTconfig-6.1.52-x86
```diff
@@ -0,0 +1,8012 @@
+#
+# Automatically generated file; DO NOT EDIT.
+# Linux/x86 6.1.x Kernel Configuration
+#
+CONFIG_CC_VERSION_TEXT="gcc (GCC) 11.2.0"
```
dimkr: I'd heavily recommend you to run `make menuconfig` to re-generate this configuration file **from Debian 12**. You're going to build this kernel under a Debian 12 container (that's what you put in kernel-kit.yml) but you configured the kernel in a different distro (I can tell by the older GCC version), so it's not guaranteed that the build result matches the configuration file (because it will pick default values for configuration keys not available under your build environment).

kernel-kit/configs_x86/DOTconfig-6.1.52-x86
```diff
@@ -0,0 +1,8012 @@
...
+# CONFIG_RD_LZMA is not set
+CONFIG_RD_XZ=y
+# CONFIG_RD_LZO is not set
+# CONFIG_RD_LZ4 is not set
+CONFIG_RD_ZSTD=y
+CONFIG_BOOT_CONFIG=y
+# CONFIG_BOOT_CONFIG_EMBED is not set
+# CONFIG_INITRAMFS_PRESERVE_MTIME is not set
+# CONFIG_CC_OPTIMIZE_FOR_PERFORMANCE is not set
+CONFIG_CC_OPTIMIZE_FOR_SIZE=y
```
dimkr: Size-optimized kernels are not much smaller than performance-optimized kernels ... most old computers either have enough storage to afford this negligible increase in size or storage can be expanded, but the slow CPU won't go away ... IMO this is a bad idea

kernel-kit/configs_x86/DOTconfig-6.1.52-x86
```diff
@@ -0,0 +1,8012 @@
...
+# end of x86 Debugging
+
+#
+# Kernel Testing and Coverage
+#
+# CONFIG_KUNIT is not set
+# CONFIG_NOTIFIER_ERROR_INJECTION is not set
+# CONFIG_FAULT_INJECTION is not set
+CONFIG_CC_HAS_SANCOV_TRACE_PC=y
+CONFIG_RUNTIME_TESTING_MENU=y
```
dimkr: AFAIK this is a big waste of space

kernel-kit/configs_x86/DOTconfig-6.1.52-x86
```diff
@@ -0,0 +1,8012 @@
...
+CONFIG_HID_JABRA=m
+CONFIG_HID_TWINHAN=m
+CONFIG_HID_KENSINGTON=m
+CONFIG_HID_LCPOWER=m
+CONFIG_HID_LED=m
+CONFIG_HID_LENOVO=m
+# CONFIG_HID_LETSKETCH is not set
+CONFIG_HID_LOGITECH=m
+CONFIG_HID_LOGITECH_DJ=m
+CONFIG_HID_LOGITECH_HIDPP=m
```
dimkr: Somebody said, don't remember who and when, that these 3 must be `=y` to make a Logitech wireless keyboard work early in the boot process (long before udev), when one selects a save file to use

kernel-kit/configs_x86/DOTconfig-6.1.52-x86
```diff
@@ -0,0 +1,8012 @@
...
+CONFIG_MIGRATION=y
+CONFIG_BOUNCE=y
+CONFIG_MMU_NOTIFIER=y
+# CONFIG_KSM is not set
+CONFIG_DEFAULT_MMAP_MIN_ADDR=4096
+CONFIG_ARCH_SUPPORTS_MEMORY_FAILURE=y
+# CONFIG_MEMORY_FAILURE is not set
+CONFIG_ARCH_WANT_GENERAL_HUGETLB=y
+CONFIG_TRANSPARENT_HUGEPAGE=y
+CONFIG_TRANSPARENT_HUGEPAGE_ALWAYS=y
```
dimkr: This can be a big waste of RAM on old computers, better enable CONFIG_TRANSPARENT_HUGEPAGE_MADVISE instead

kernel-kit/configs_x86/DOTconfig-6.1.52-x86
```diff
@@ -0,0 +1,8012 @@
...
+# CONFIG_PARIDE is not set
+# CONFIG_BLK_DEV_PCIESSD_MTIP32XX is not set
+CONFIG_ZRAM=m
+# CONFIG_ZRAM_DEF_COMP_LZORLE is not set
+CONFIG_ZRAM_DEF_COMP_ZSTD=y
+# CONFIG_ZRAM_DEF_COMP_LZ4 is not set
+# CONFIG_ZRAM_DEF_COMP_LZO is not set
+# CONFIG_ZRAM_DEF_COMP_LZ4HC is not set
+# CONFIG_ZRAM_DEF_COMP_842 is not set
+CONFIG_ZRAM_DEF_COMP="zstd"
```
dimkr: This is a bad idea, Puppy's init script creates a zram block device and uses it for swap - in old and slow computers that have little RAM plus a slow CPU with <=2 cores, swap can be quite active, and the default of LZO-RLE makes much more sense compared to your choice of zstd - zstd can be CPU intensive

kernel-kit/6.1.x-x86-build.conf
```diff
@@ -0,0 +1,129 @@
...
+#COMP="-comp zstd -Xcompression-level 19 -b 256K -no-exports -no-xattrs"
```
dimkr, line 78: I'd go with zstd (in every SFS, not just zdrv) ... computers that run a 32-bit distro must be super old and the ~10% increase in size is worth it considering the huge boost in decompression speed

peabee: Hmmmmmmmm............
You clearly think that the Porteus kernel config has many problems..........
There has been limited testing on the forum but clearly not enough. And Porteus have been using for some time, presumably without issue...
I think it would be best if you could propose an alternative config that met your concerns. Perhaps an adaptation of the kernel-kit 5.15 32-bit config?
I would like to have a 32-bit 6.1 available given that future problems with aufs-5.x will no longer be fixed.

dimkr: > You clearly think that the Porteus kernel config has many problems..........

You asked me to review, so I did my best to tell you what I think is wrong in this configuration file 🤷‍♂️

I'd recommend the Vanilla Dpup 10.0.x configuration file, grab a 32-bit build and `zcat /proc/config.gz`.

peabee:
Kernel config replaced with dpup-kernel version.

dimkr (approved):
Assuming now it's identical to dpup's, it should be fine. I'd still drop the use of xz or at least reduce block size, though.

If I were you I'd just use the dpup kernel and stop using aufs (because it will drop support for 6.1, it will always push you towards newer and newer kernels that keep getting worse on ancient hardware). Moreover, I'd retire the 5.10.x and 5.15.x kernels to save storage space and reduce build times; as time goes by, they're more and more likely to have a bug or performance regression in aufs that won't get fixed.

PR #4170, opened 2023-10-22
Delete empty directories in sfs_load.overlay only if new (fixes #4169)

### LINKED ISSUE #4169: /opt is being deleted in overlayfs build after extra sfs install
@dimkr ?
BookwormPup32 with a savefolder and an extra sfs with /opt in it loaded - after reboot there is no longer a /opt directory.....
Screenshot

The savefolder upper directory shows /opt has been replaced by "something else":
Screenshot2

dimkr:
Currently sfs_load.overlay removes empty leftover directories when unloading a SFS.

However, if /opt is empty (the default in Puppy) and a SFS contains /opt, it will remove /opt. This is only a cosmetic problem because /opt will be created when you load a SFS or install a package that contains /opt, and an empty /opt is an unused /opt.

Under PUPMODE 12, sfs_load.overlay makes a list of files and directories in the SFS, so it can delete the symlinks from disk once the SFS is unloaded. To fix the issue, I enabled this even for other PUPMODEs and restricted the list so it contains only files and directories that **did not exist when the SFS was loaded**. Additionally, I added locking so a race condition between a sfs_load that loads a SFS and another that unloads a SFS doesn't result in a bad list that causes leftover directories. 

Best viewed with whitespace changes hidden, because this PR is tiny once you ignore the removed `if ... fi` lines:

whitespace

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/sfs_load.overlay
+++ b/woof-code/rootfs-skeleton/usr/sbin/sfs_load.overlay
@@ -55,6 +55,10 @@ do
 	shift
 done
 
+# wait until other instances are done, otherwise we can't list files added by each SFS
+exec 9<>/run/sfs_load.lock
+flock 9
+
 # loading of SFSs in EXTRASFSLIST is handled by initrd, see initrd-progs/0initrd/init
 if [ $START -eq 1 -o $STOP -eq 1 ]; then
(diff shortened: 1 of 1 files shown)


peabee (approved):
Tested and works in that /opt no longer disappears after an extra sfs with /opt contents is loaded, queued and rebooted.

PR #4177, opened 2023-11-03
Drop 4.19.x, 5.4.x and periodic builds that use them

dimkr:
This PR drops 4.19.x, 5.4.x, all kernel-kit configurations other than 5.10.x, 5.15.x and 6.1.x (the longterm branches in use), and periodic builds that use 4.19.x and 5.4.x.


peabee (commented):
I think we could keep the workflows amended to use later kernels but disable the periodic workflows........ and maybe keep the kernel configs? (not that bothered)

dimkr:
The workflows are useless without the periodic trigger, they contain nothing but inputs for build.yml. They don't "do" anything without the trigger.

peabee:
OK - just thought might be possible to retain them "just in case"....

PR #4179, opened 2023-11-07
mtpaint - remove no* build options

peabee:
OscarTalks says:
Looking at the mtpaint recipe I see that the configure options do have webp and tiff (and also lcms and jpeg2000) actively disabled. So I would remove the nowebp, notiff, nolcms and nojp2 configure options from both of the 2 lines (one for gtk3 and the other for gtk2) The configure script will look for the DEV packages and autodetect everything, so if those libraries are in the build environment then support will be compiled in. No reason to actively disable anything unless we are encountering some kind of a problem. I don't think I have ever encountered tiff image files in common usage, but webp is increasing in popularity, including animated ones.


dimkr:
webp is useful, maaaaaaaaybe TIFF is useful, but I doubt the usefulness of the others. Every one you enable will make mtpaint slower to start and maybe increase memory consumption.

PR #4187, opened 2023-11-19
Provide support for "root=UUID=" boot parameter - init and isoboot

gyrog:
This provides support for "root=UUID=" boot paarameter in both init and isoboot.
Such parameters may be generated by grub2 utilitiy scripts.


peabee:
https://forum.puppylinux.com/viewtopic.php?t=5484

needs updating to reflect this new boot parameter?

also README.txt

gyrog:
It's about recognising that just as a Puppy Linux installed grub2 can be used to boot other OS's,
so can a Puppy Linux be booted by a grub2 installed by some other OS, including the generation of the grub.cfg boot entry.
"root=UUID=" is a grub2 convention, not a thing peculiar to Puppy Linux.
There is already support for it in '/sbin/init_full_install'.

PR #4192, opened 2023-11-26
Update init: use binded mountpoint instead of direct path for save folder

rizalmart:
Reason: If savefolder was used and the actual path of savefolder was used. It would not release from layered root filesystem and unmount the drive when the save folder was located during shutdown. This may cause filesystem corruption. In order to detach the savefolder gracefully from layer root filesystem in aufs was to use bind mountpoint. In order to retain the attached path on aufs layered system.

### DIFF
--- a/initrd-progs/0initrd/init
+++ b/initrd-progs/0initrd/init
@@ -503,12 +503,14 @@ setup_psave(){ # setup savefile or savefolder
   fi
  elif [ -d "$SAVE_FN" ];then #savefolder
   echo "--SAVEFOLDER-- $SAVE_FN" #debug
+  mkdir -p /pup_disk
+  mount -o bind $SAVE_FN /pup_disk
   if [ "$UNIONFS" = 'overlay' ]; then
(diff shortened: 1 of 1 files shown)


gyrog:
A couple of observations:

1) When we first introduced "savefolder" , many moons ago, it used a bind mount at /pup_rw.
This created havoc with a number of scripts in the running Puppy because they expected only 1 entry in /proc/mounts for the partition containing the "savefolder", but with the bind mount there were now 2 entries.
I hope that the relevant patches have not been "lost" in the intervening years.
And that there is no new code that makes the same assumption.

2) I hope it makes a difference, but I don't understand bind mounts enough to see how it can,
if the union/stack still considers it mounted read/write.
I suspect that the real solution for rc.shutdown, is to do something like a "pivot-root" away from the union/stack to a filesystem in ram, and then umount the union/stack, and umount the partitions.

In the short term, maybe rc.shutdown should not be trying to umount stuff that is part of the union/stack,
but rather trying to get everything remounted read-only.

PR #4201, opened 2023-12-03
Reload sfwbar on .desktop file change

dimkr:


LBCrion:
Is a full reload needed? The full reload may reorder windows in the taskbar and will close any poups the user may have open. 
 It should be possible to do something like this:

```
TriggerAction "SIGRTMIN+5", Function "RebuildMenu"
Function("RebuildMenu") {
  ClearMenu "MyStartMenuName"
  [Run  the same functions that build menu on startup]
}
```
You can then send signal RTMIN+5 instead of SIGHUP and rebuild the menu in the running process without affecting other components.

dimkr:
@LBCrion Thanks, I'll see if I can do something like this. The problem is that Puppy traditionally relies on `fixmenus`, a script that updates various menus and reloads them. The package manger and other things run this script. Therefore, this script should ask sfwbar to reload after it updates the menu, and it would be cleaner to live the generation in `fixmenus` rather than moving it to sfwbar.

LBCrion:
That should be doable.  I assume fixmenus produces an output file that
sfwbar config consumes in some way? It should be possible to just clear the
menu and re-read the menu config file upon receiving a trigger signal. If
you can point b me to the sfwbar config you're using and the output that
fixmenus produces,  I can take a stab at wiring a menu reload function.

dimkr:
> I assume fixmenus produces an output file that sfwbar config consumes in some way?

fixmenus calls a tool called sfwbar-xdgmenu, which updates ~/.config/sfwbar/menu.source and it's sourced at https://github.com/puppylinux-woof-CE/woof-CE/blob/d57a773c428f395f67e392a1ed95695e4108a426/woof-code/rootfs-petbuilds/sfwbar/root/.config/sfwbar/buttonmenu.widget#L1

LBCrion:
You should be able to change buttonmenu.widget to:
```
include("menu.source")

TriggerAction "SIGRTMIN+5","ReloadMenu"
Function("ReloadMenu") {
  ClearMenu "main_menu"
  PipeRead "cat /path/to/menu.source"
}

layout {
  button {
    value = "/usr/share/doc/puppylogo48.png"
    action = Menu "main_menu"                              # launch menu on click
    tooltip = "Launch menu"
    css = "* { min-height: 22px; min-width: 29px; }"
  }
}
```

This should reload the menu config upon receiving signal RTMIN+5. The only untidy bit here is that you need to specify the a full path to menu.source in PipeRead. I should probably add an action to read a config file (similar to include, but of course an action wouldn't add any sub-layouts into the main layout hierarchy).

dimkr:
Thanks @LBCrion, will try

LBCrion:
I added a Read expression function, so with the latest git version you don't need a full path to menu.source and don't need to run cat just to read a file:

```
include("menu.source")

TriggerAction "SIGRTMIN+5","ReloadMenu"
Function("ReloadMenu") {
  ClearMenu "main_menu"
  Config Read "menu.source"
}

layout {
  button {
    value = "/usr/share/doc/puppylogo48.png"
    action = Menu "main_menu"                              # launch menu on click
    tooltip = "Launch menu"
    css = "* { min-height: 22px; min-width: 29px; }"
  }
}
```

dimkr:
Thank you @LBCrion, will try once beta14 is tagged

PR #4218, opened 2024-01-25
Update geany & geany-gtk2 to version 2.0

lakshayrohila:
- Updates geany and geany-gtk2 to version 2.0
- Enables both plugin and VTE in both versions (adds 2MBs only)
- Tested in Jammy64. Adds ~11MB in ISO size.


lakshayrohila:
Sorry, wrong branch.

woof-code/rootfs-petbuilds/geany-gtk2/petbuild
```diff
@@ -1,13 +1,13 @@
...
 }
 
 build() {
-    tar -xjf geany-1.37.1.tar.bz2
-    cd geany-1.37.1
+    tar -xjf geany-2.0.tar.bz2
+    cd geany-2.0
     patch -p1 < ../defaults.patch
-    ./configure --prefix=/usr --disable-plugins --disable-vte --enable-gtk2
+    ./configure --prefix=/usr --enable-plugins --enable-vte --enable-gtk2
```
dimkr, line 9: @lakshayrohila Are you sure that Geany 2.0 supports GTK+ 2? AFAIK --enable-gtk2 isn't supported and this petbuild actually produces a GTK+ 3 build. Originally, I copied `geany` to `geany-gtk2` and made some build configurations use `geany-gtk2` to avoid breaking Puppy releases based on old Ubuntu or Slackware versions, too old for latest Geany.

lakshayrohila: @dimkr, Well, yes, I didn't test that! I ran it and it worked well, so I thought it might be supported. Which older puppy can I build to test if this works. For Jammy64, I had changed from using geany to geany-gtk2 in __00build.conf

lakshayrohila: The main question is - is this update even required bringing something new?

dimkr: Look at the Geany changelog and decide if you want to update or not 🤷🏿

I'd recommend leaving geany-gtk2 alone, because it's used by build configurations that can't build the latest Geany (stuff like slackware/14.2) where GTK+ 3 is too old.

lakshayrohila: @dimkr, I think there are some things which can be useful for some Puppy users in this update. The only problem is those extra 11 MB it brings with it. Still I'll consider this to be OK to be updated.

dimkr: So update `geany` and leave `geany-gtk2` alone

PR #4219, opened 2024-01-25
Update geany to version 2.0

lakshayrohila:
- Updates geany to version 2.0.
- Tested in Jammy64. ~Adds 9MB to the ISO size.~ *ALMOST NO ADDITIONAL SPACE USED (must be within 1 MB).*

NOTE: It was my fault, I had compared the sizes of two Jammy64s, but somehow it was a wrong comparison. Today, I built a Jammy64 without these changes, and the ISO size with and without these changes are exactly the same 😅.

@peabee please review.


woof-code/rootfs-petbuilds/geany-gtk2/petbuild
```diff
@@ -1,13 +1,13 @@
...
+    [ -f geany-2.0.tar.bz2 ] || wget -t 3 -T 60 https://download.geany.org/geany-2.0.tar.bz2
 }
 
 build() {
-    tar -xjf geany-1.37.1.tar.bz2
-    cd geany-1.37.1
+    tar -xjf geany-2.0.tar.bz2
+    cd geany-2.0
     patch -p1 < ../defaults.patch
-    ./configure --prefix=/usr --disable-plugins --disable-vte --enable-gtk2
```
lakshayrohila: Is this required?

woof-code/rootfs-petbuilds/geany-gtk2/pinstall.sh
lakshayrohila, line 1: @peabee, I added this because this was included in the `geany` petbuild.

Is this required?

woof-code/rootfs-petbuilds/geany-gtk2/petbuild
dimkr: https://github.com/puppylinux-woof-CE/woof-CE/pull/4218#discussion_r1466462876

woof-code/rootfs-petbuilds/geany/petbuild
```diff
@@ -1,13 +1,13 @@
...
 }
 
 build() {
-    tar -xjf geany-1.38.tar.bz2
-    cd geany-1.38
+    tar -xjf geany-2.0.tar.bz2
+    cd geany-2.0
     patch -p1 < ../defaults.patch
-    ./configure --prefix=/usr --disable-plugins --disable-vte
+    ./configure --prefix=/usr --enable-plugins --enable-vte
```
dimkr: The plugins are big and make Geany slower to start - traditionally, Puppy doesn't have them

lakshayrohila: @dimkr, Ok, so that's why I wasn't seeing any increase in ISO size when enabling plugins. What about VTE?

dimkr: I believe most people who use Geany on Puppy don't use it as an IDE but as a text editor, and doubt the usefulness of the integrated terminal

dimkr (approved):
Looks fine to me

peabee:
Happy with the current builds - no problems known - does the new version provide significant improvements?
I think only BookwormPup32 uses the GTK3 petbuild so no objection to the change if it is no bigger.

Void, Debian and Ubuntu are still on 1.38 (see pkgs.org) - why haven't they updated I wonder?

lakshayrohila:
> Happy with the current builds - no problems known - does the new version provide significant improvements?

https://www.geany.org/documentation/releasenotes/ lists some changes. Personally, I don't see many major improvements.

> I think only BookwormPup32 uses the GTK3 petbuild so no objection to the change if it is no bigger.

Jammy64 also uses it. Tested on Jammy; works!

lakshayrohila:
@peabee, @dimkr, if there are no issues with this and my other pull request (#4221), then please accept them as I can't do it since I do not have write permissions to the woof-CE repo.

Otherwise, if there are problems in the PRs, please notify me. I'll try to correct them (or if the ideas about those PRs are wrong, then I'll close the PRs) :+1:.

dimkr:
> Void, Debian and Ubuntu are still on 1.38 (see pkgs.org) - why haven't they updated I wonder?

Debian 12 and Ubuntu 23.10 (the latest released versions) won't receive Geany 2.0, only Debian 13 and Ubuntu 22.04. That's why you don't see it in pkgs.org.

> @peabee, @dimkr, if there are no issues with this and my other pull request (#4221), then please accept them as I can't do it since I do not have write permissions to the woof-CE repo.

I don't have no permissions either, I left the organization.

techrockedge:
I've tested Geany 2.0 on _F96-CE_4_ and vote on merging the commit sooner than later. Though I am happy with 1.38 it seems that v2.0 has a good complement of plugins to start and the interface and work flow remains the same as versions 1.37 and 1.38

PR #4220, opened 2024-01-26
Add and use new color theme in lxterminal

lakshayrohila:
VGA is the default lxterminal preset, but its dark blue color has very little contrast with the background (for lower brightness). This replaces the original dark blue color with the one that has better contrast (slightly brighter).

Old:
old_lxterminal
lxterminal-htop-old

New:
lxterminal-htop new
lxterminal-ls new

@dimkr, @peabee, please review.

### DIFF
--- a/woof-code/rootfs-petbuilds/lxterminal/colors.patch
+++ b/woof-code/rootfs-petbuilds/lxterminal/colors.patch
@@ -1,6 +1,15 @@
 diff -rup lxterminal-0.4.0-orig/src/setting.c lxterminal-0.4.0/src/setting.c
---- lxterminal-0.4.0-orig/src/setting.c	2021-03-12 20:20:26.156393104 +0200
-+++ lxterminal-0.4.0/src/setting.c	2021-03-13 20:43:20.653893675 +0200
+--- lxterminal-0.4.0-orig/src/setting.c	2024-01-26 11:00:15.817511151 +0530
++++ lxterminal-0.4.0/src/setting.c	2024-01-26 11:46:02.729178961 +0530
+@@ -42,7 +42,7 @@ ColorPreset color_presets[] = {
+         .foreground_color = "#aaaaaa",
+         .palette = {
+             "#000000", "#aa0000", "#00aa00", "#aa5500",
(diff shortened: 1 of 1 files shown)


peabee:
Very marginal change - would be better to change to a different colour?

lakshayrohila:
> Very marginal change - would be better to change to a different colour?

Not sure which colour to change to.. if you can suggest one?

---

Or, we can also try changing the background color to some lighter black/grey. I'll try some combinations :+1:

peabee (commented):
Can't this be done in lxterminal.conf rather than changing code???

dimkr:
Changing the hardcoded defaults within the code also applies this change to lxterminal running as spot - much cleaner than having two configuration files that eventually run out of sync.

lakshayrohila:
> Changing the hardcoded defaults within the code also applies this change to lxterminal running as spot - much cleaner than having two configuration files that eventually run out of sync.

I was just now reading through the code of lxterminal's `load_setting` function. It turns out that if there is no user config file (`~/.config/lxterminal/lxterminal.conf`), it falls back to using the system config file (`--prefix/share/lxterminal/lxterminal.conf`, or in the case of this petbuild `/usr/share/lxterminal/lxterminal.conf`). So, instead of creating multiple user configs, according to my intuition, we can create a single system config that applies to all users.

But, that is just another way of doing things. So, why waste our time doing the same thing by some different method which we have done by some other method already 🤷‍♂️?

I am currently just waiting for this ISO to download - my internet connection is extremely slow at the moment (under 100KBps :rofl: ). I tried manually compiling this in QEMU running Jammy64, but I think as I had done this to replace an `lxterminal` install with a new install while running a `lxterminal`, this caused some issues and removed the top bar from the manual install I did. I'll test if this works and then open this PR.

---

EDIT: Ironically, the download finished just as I was writing this :smiley:.

The colour palette I have used here is also bad (some colours have low contrast with the background). Also, explicit change of background and foreground colors are required in the code (as was earlier done). I think I'll leave this work for tomorrow :yawning_face:.

peabee (commented):
No objection

PR #4221, opened 2024-01-26
Add tldr-py and btop to Jammy64

lakshayrohila:
This commit adds `tldr-py` (petbuild) and `btop` to Jammy64 on Clarity's request [here](https://forum.puppylinux.com/viewtopic.php?p=110058#p110058).

It also introduces two petbuilds: `tldr-py` and `python3-shtab` (the latter being a dependency of `tldr-py` petbuild).

I wanted to use `tldr-py` instead of the regular `tldr` because the package `tldr` is quite large as it has [dependencies](https://packages.ubuntu.com/jammy/tldr) such as `git`. There is also a `tldr-py` [Ubuntu package](https://packages.ubuntu.com/jammy/tldr-py), but in Jammy, this one was quite old and, IMHO, not very user-friendly.

`python3-shtab` petbuild was required to be added because it is not available in Ubuntu Jammy's repos (it started with Ubuntu Lunar).


lakshayrohila:
@peabee, @dimkr, please review.

woof-code/rootfs-petbuilds/python3-shtab/pet.specs
```diff
@@ -0,0 +1 @@
+python3-shtab-1.6.5|python3-shtab|1.6.5||Utility|2800K||python3-shtab-1.6.5.pet||Automagic shell tab completion for Python CLI applications|puppy|||
```
lakshayrohila, line 1: Please check the `2800K` here. I have no idea what this is used for.

woof-code/rootfs-petbuilds/tldr-py/tldr.1
lakshayrohila, line 1: NOTE: This manpage was built from the docs' source provided in the `tldr-py`'s this release. The `setup.py` was failing without docs (I mean only the man page), and building docs required large no. of dependencies.

lakshayrohila: This file gets copied to `/` instead of `/usr/share/man/man1`.

peabee:
Not sure they're really needed..... but if only JammyPup64 is affected - no objection.

lakshayrohila:
> Not sure they're really needed..... but if only JammyPup64 is affected - no objection.

Not really required - Just few additional features (we are still trying to build an official Jammy64, right?)

dimkr:
I'd strongly recommend you not to add petbuilds for Python packages, unless you're 100% committed to maintaining them in the long term. They tend to be sensitive to the Python version. In this particular case, I wouldn't do this because I think the benefit is not worth it - documentation beyond man pages is easy to find online and I think the extra size and maintenance cost is not worth it.

lakshayrohila:
> I'd strongly recommend you not to add petbuilds for Python packages, unless you're 100% committed to maintaining them in the long term.

Actually, 0% committed for maintenance :-1:.

> They tend to be sensitive to the Python version. In this particular case, I wouldn't do this because I think the benefit is not worth it - documentation beyond man pages is easy to find online and I think the extra size and maintenance cost is not worth it.

Well, I also agree with this. Thanks for the information and the advice!

I'll close the PR.

PR #4225, opened 2024-01-28
Fix #4224

### LINKED ISSUE #4224: petbuild output is taken from cache even when there is a change in petbuild source (except the petbuild file)
petbuild output is taken from cache even when there is a change in petbuild source (except the petbuild file).

For example, I changed the code of `colors.patch` in `lxterminal` petbuild. But according to https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-code/support/petbuilds.sh#L71, the cached petbuild output was used since there was no change in any of the files used for generating that `HASH` variable.

This can cause confusion while development.

lakshayrohila:
Uses a compressed form of the petbuild directory for generating hash instead of only using the `petbuild` file.

### DIFF
--- a/woof-code/support/petbuilds.sh
+++ b/woof-code/support/petbuilds.sh
@@ -68,7 +68,8 @@ for NAME in $PETBUILDS; do
         continue
     fi
 
-    HASH=`cat ../DISTRO_PKGS_SPECS-${DISTRO_BINARY_COMPAT}-${DISTRO_COMPAT_VERSION} ../DISTRO_COMPAT_REPOS ../DISTRO_COMPAT_REPOS-${DISTRO_BINARY_COMPAT}-${DISTRO_COMPAT_VERSION} ../DISTRO_PET_REPOS ../rootfs-petbuilds/${NAME}/petbuild 2>/dev/null | md5sum | awk '{print $1}'`
+    tar -czf /tmp/petbuild.tar.gz ../rootfs-petbuilds/${NAME}/ # /tmp/petbuild.tar.gz is also used for generating HASH; can be used to compare any change in any file/folder of the petbuild directory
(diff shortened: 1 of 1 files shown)


lakshayrohila:
@peabee. @dimkr, please review.

woof-code/support/petbuilds.sh
```diff
@@ -68,7 +68,8 @@ for NAME in $PETBUILDS; do
         continue
     fi
 
-    HASH=`cat ../DISTRO_PKGS_SPECS-${DISTRO_BINARY_COMPAT}-${DISTRO_COMPAT_VERSION} ../DISTRO_COMPAT_REPOS ../DISTRO_COMPAT_REPOS-${DISTRO_BINARY_COMPAT}-${DISTRO_COMPAT_VERSION} ../DISTRO_PET_REPOS ../rootfs-petbuilds/${NAME}/petbuild 2>/dev/null | md5sum | awk '{print $1}'`
+    tar -czf /tmp/petbuild.tar.gz ../rootfs-petbuilds/${NAME}/ # /tmp/petbuild.tar.gz is also used for generating HASH; can be used to compare any change in any file/folder of the petbuild directory
```
dimkr, line 71: The output of `tar` is non-deterministic, it contains things like timestamps and the order of files can change. The hash is almost guaranteed to change every time this command runs, so you're essentially disabling the cache completely.

https://reproducible-builds.org/docs/archives/

lakshayrohila: Thanks, I didn't know about that.

We will need some other approach then. Closing this PR.

lakshayrohila: @dimkr, I have another idea.

How about using the output of `diff -r [CACHED_PETBUILD_DIR] [WOOF-CE_PETBUILD_DIR]`? If the contents are different, `diff` will output something; otherwise, its output will be empty.

Will that work?

dimkr: I don't know if it's done in deterministic order, and I think the output contains timestamps.

lakshayrohila: I'll test this idea tomorrow and open a PR if it works.

Also, to clear a few things, I was not saying about using diff's output for generating HASH; instead checking it directly if it is empty or not.

dimkr: If you don't hash it, you need to store it on disk as-is, increasing cache size and making the comparison that determines whether or not rebuild is needed slower.

peabee:
I always delete the caches before running a build if I need to guarantee that a new version of something is used.....

PR #4234, opened 2024-02-05
Update puppyinstaller

Kopernicusss:
add "BDRV" to install, fixes apt breakage after hard drive install.

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/puppyinstaller
+++ b/woof-code/rootfs-skeleton/usr/sbin/puppyinstaller
@@ -1,6 +1,6 @@
 #!/bin/bash
 #(c) Copyright Barry Kauler 2006 www.puppylinux.com
-#(c) Copyright 01micko, shinobar, kopernicusss and others
+#(c) Copyright 01micko, shinobar and others
 #2007 Lesser GPL licence v2 (/usr/share/doc/legal/lgpl-2.1.txt)
 #140705 major overhaul of guis and conditions.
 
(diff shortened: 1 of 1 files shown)


woof-code/rootfs-skeleton/usr/sbin/puppyinstaller
```diff
@@ -864,7 +865,7 @@ case $PMEDIA in usb*)
 		yaf-splash -bg orange -text "`eval_gettext \"Please wait, copying files to \\\$DESTPART ...\"`" &
 		XPID=$!
 		mkdir -p /mnt/data
-		for i in vmlinuz initrd.gz ${DISTRO_PUPPYSFS} ${DISTRO_ZDRVSFS} ${DISTRO_FDRVSFS} ${DISTRO_ADRVSFS} ${DISTRO_YDRVSFS}; do
+		for i in vmlinuz initrd.gz ${DISTRO_PUPPYSFS} ${DISTRO_ZDRVSFS} ${DISTRO_FDRVSFS} ${DISTRO_ADRVSFS} ${DISTRO_YDRVSFS} ${DISTRO_BDRVSFS}; do
```
dimkr, line 868: Why not align this list with https://github.com/puppylinux-woof-CE/woof-CE/blob/5ff59258c353b2f8739af7099b9b4ebd7f34b6f7/woof-code/rootfs-skeleton/usr/sbin/bootflash#L873?

peabee: Full installs are deprecated and we prefer to promote frugal installs - however this addition does seem sensible.

There is no need to update the historic Copyright comment as Github records all "blame".

Kopernicusss: patch updated to reverse copyright change.

dimkr: But you haven't added kbuild and other SFSs 🤷🏿

PR #4236, opened 2024-02-11
Create bluez-nologind.conf for pipewire

rizalmart:
This configuration is a workaround in order to make bluetooth devices work on pipewire without active login session like logind/elogind

See https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/3828


peabee:
As this is a workaround which only applies if pipewire is installed, it should be implemented by an amendment to:
https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-code/packages-templates/pipewire_FIXUPHACK

PR #4240, opened 2024-02-15
Sync x-www-browser with defaultbrowser when using Palemoon petbuild

lakshayrohila:
Use `update-alternatives` command to sync by default, `/usr/bin/x-www-browser` with `/usr/local/bin/defaultbrowser`.

Specifically, this is required in `epdfview` since it uses `x-www-browser` by default to open URL links.

@dimkr, @peabee please review.

### DIFF
--- a/woof-code/rootfs-petbuilds/palemoon/pinstall.sh
+++ b/woof-code/rootfs-petbuilds/palemoon/pinstall.sh
@@ -1,3 +1,19 @@
-echo '#!/bin/sh
-exec palemoon "$@"' > usr/local/bin/defaultbrowser
-chmod 755 usr/local/bin/defaultbrowser
+# Set defaultbrowser to run x-www-browser - set the default browser the Debian way, not the Puppy way.
+case ${DISTRO_BINARY_COMPAT} in
+    ubuntu|trisquel|debian|devuan|raspbian)
+        /usr/bin/update-alternatives --install /usr/bin/x-www-browser x-www-browser /usr/bin/palemoon 200
+        echo '#!/bin/sh
+CMD=$(command -v x-www-browser)
+[ -z "$CMD" ] && CMD="/usr/bin/palemoon"
+exec "$CMD" "$@"
+' > /usr/local/bin/defaultbrowser
+        ;;
+    *)
+        echo '#!/bin/sh
+exec /usr/bin/palemoon "$@"
+' > /usr/local/bin/defaultbrowser
+        ;;
+esac
+
+chmod 755 /usr/local/bin/defaultbrowser
+cp /usr/local/bin/defaultbrowser /usr/local/bin/defaulthtmlviewer
--- a/woof-code/rootfs-petbuilds/palemoon/pinstall.sh
+++ b/woof-code/rootfs-petbuilds/palemoon/pinstall.sh
@@ -16,4 +16,4 @@ exec /usr/bin/palemoon "$@"
 esac
 
 chmod 755 usr/local/bin/defaultbrowser
-cp usr/local/bin/defaultbrowser usr/local/bin/defaulthtmlviewer
+cp -p usr/local/bin/defaultbrowser usr/local/bin/defaulthtmlviewer
--- a/woof-code/rootfs-petbuilds/palemoon/pinstall.sh
+++ b/woof-code/rootfs-petbuilds/palemoon/pinstall.sh
@@ -1,19 +1,19 @@
 # Set defaultbrowser to run x-www-browser - set the default browser the Debian way, not the Puppy way.
 case ${DISTRO_BINARY_COMPAT} in
     ubuntu|trisquel|debian|devuan|raspbian)
-        /usr/bin/update-alternatives --install /usr/bin/x-www-browser x-www-browser /usr/bin/palemoon 200
+        chroot . /usr/bin/update-alternatives --install /usr/bin/x-www-browser x-www-browser /usr/bin/palemoon 200
         echo '#!/bin/sh
 CMD=$(command -v x-www-browser)
 [ -z "$CMD" ] && CMD="/usr/bin/palemoon"
 exec "$CMD" "$@"
-' > /usr/local/bin/defaultbrowser
+' > usr/local/bin/defaultbrowser
         ;;
     *)
         echo '#!/bin/sh
 exec /usr/bin/palemoon "$@"
-' > /usr/local/bin/defaultbrowser
+' > usr/local/bin/defaultbrowser
         ;;
 esac
 
-chmod 755 /usr/local/bin/defaultbrowser
-cp /usr/local/bin/defaultbrowser /usr/local/bin/defaulthtmlviewer
+chmod 755 usr/local/bin/defaultbrowser
+cp usr/local/bin/defaultbrowser usr/local/bin/defaulthtmlviewer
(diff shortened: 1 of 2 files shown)


dimkr:
In Vanilla Dpup I went for a different solution, defaultbrowser runs x-www-browser. You set the default browser the Debian way, not the Puppy way.

lakshayrohila:
> In Vanilla Dpup I went for a different solution, defaultbrowser runs x-www-browser. You set the default browser the Debian way, not the Puppy way.

@dimkr, thanks for the suggestion! I have re-implemented this fix to your suggestion. Please re-review.

dimkr:
I think it would be cleaner to change the palemoon petbuild, so it makes x-www-browser run palemoon, then sets defaultbrowser to x-www-browser, only in Debian or Ubuntu based build.

https://github.com/puppylinux-woof-CE/woof-CE/blob/4c8a372cd9bec68d687d2f8018e7121d5563efca/woof-code/packages-templates/firefox_FIXUPHACK#L87

lakshayrohila:
> I think it would be cleaner to change the palemoon petbuild, so it makes x-www-browser run palemoon, then sets defaultbrowser to x-www-browser, only in Debian or Ubuntu based build.
> 
> https://github.com/puppylinux-woof-CE/woof-CE/blob/4c8a372cd9bec68d687d2f8018e7121d5563efca/woof-code/packages-templates/firefox_FIXUPHACK#L87

@dimkr, done, please review.

---

One question: is `pinstall.sh` run in `chmod`? Reading some lines of code in `3builddistro`, I don't think it does. Then I'll have to update this commit.

woof-code/rootfs-petbuilds/palemoon/pinstall.sh
```diff
@@ -1,3 +1,19 @@
...
+' > /usr/local/bin/defaultbrowser
+        ;;
+    *)
+        echo '#!/bin/sh
+exec /usr/bin/palemoon "$@"
+' > /usr/local/bin/defaultbrowser
+        ;;
+esac
+
+chmod 755 /usr/local/bin/defaultbrowser
```
dimkr: pinstall.sh runs without `chroot` during woof-CE, you must use relative paths - usr/local/bin/defaulbrowser, etc'

woof-code/rootfs-petbuilds/palemoon/pinstall.sh
```diff
@@ -1,3 +1,19 @@
-echo '#!/bin/sh
-exec palemoon "$@"' > usr/local/bin/defaultbrowser
-chmod 755 usr/local/bin/defaultbrowser
+# Set defaultbrowser to run x-www-browser - set the default browser the Debian way, not the Puppy way.
+case ${DISTRO_BINARY_COMPAT} in
+    ubuntu|trisquel|debian|devuan|raspbian)
+        /usr/bin/update-alternatives --install /usr/bin/x-www-browser x-www-browser /usr/bin/palemoon 200
+        echo '#!/bin/sh
+CMD=$(command -v x-www-browser)
```
dimkr, line 6: You'll need to use `chroot . sh -c 'command -v x-www-browser'`, because the shell that runs pinstall.sh runs on the host system and not what you're building (i.e. you're checking if the OS you're running woof-CE on has x-www-browser, not the OS you're building)

woof-code/rootfs-petbuilds/palemoon/pinstall.sh
```diff
@@ -1,3 +1,19 @@
-echo '#!/bin/sh
-exec palemoon "$@"' > usr/local/bin/defaultbrowser
-chmod 755 usr/local/bin/defaultbrowser
+# Set defaultbrowser to run x-www-browser - set the default browser the Debian way, not the Puppy way.
+case ${DISTRO_BINARY_COMPAT} in
+    ubuntu|trisquel|debian|devuan|raspbian)
+        /usr/bin/update-alternatives --install /usr/bin/x-www-browser x-www-browser /usr/bin/palemoon 200
```
dimkr: You'll need to wrap this with `chroot`

woof-code/rootfs-petbuilds/palemoon/pinstall.sh
lakshayrohila: Thanks! I'll make the changes.

woof-code/rootfs-petbuilds/palemoon/pinstall.sh
lakshayrohila: Please note this line is wrapped within single quotes. So, this is the content of the file `/usr/local/bin/defaultbrowser`, not actually run in the `pinstall.sh` file.

lakshayrohila: Although, I am not really sure if this check is needed. By what I understood by the code you referred to, I thought that `defaultbrowser` always makes this check (in Firefox); so I also added this check here.

dimkr: > not actually run in the pinstall.sh file.

You're right, I admit I reviewed this very quickly while also trying to make a living :)

dimkr: The situation in Puppy is a bit weird because when browsers ask to set themselves as the default they do so through xdg-mime and mimeapps.list: they basically override Puppy's hack to set defaultbrowser.desktop as the default browser. You end up with two default browsers: one that's the default in ROX-Filer and Puppy's replacement for xdg-open, and one for everything else.

lakshayrohila: > one that's the default in ROX-Filer and Puppy's replacement for xdg-open, and one for everything else

I think that although it is not a desired behaviour, still when I use Puppy, I do not mind going to "Set default apps" application and setting the new default (for example if I installed a new browser).

dimkr: The "Set default apps" thing changes defaultbrowser and friends instead of changing mimeapps.list (which sets defaultbrowser as the default for https:// URLs, etc'). In contrast, the "set as default" feature of browsers changes mimeapps.list, while root and spot have separate mimapps.list files.

dimkr: https://github.com/puppylinux-woof-CE/woof-CE/blob/4c8a372cd9bec68d687d2f8018e7121d5563efca/woof-code/3builddistro#L1030

This block in 3builddistro makes Puppy behave pretty much like other distros by 1) removing Puppy's mimeapps.list, which sets defaultbrowser as the handler for https:// URLs, etc' 2) making defaultbrowser run the default handler for https:// URLs instead of the other way around (if the user sets the default browser through the browser or through the file manager, defaultbrowser respects this decision).

This is activated when you build a Puppy without ROX-Filer, because defaultbrowser, Puppy's mimeapps.list and other Puppy-specific hacks exist only to make ROX-Filer happy.

I don't see a clean way to make the browser's "set as default" feature work in a Puppy with ROX-Filer, unless you're willing to make very significant changes like removing Puppy's mimeapps.list and xdg-open.

woof-code/rootfs-petbuilds/palemoon/pinstall.sh
```diff
@@ -1,3 +1,19 @@
...
+        ;;
+    *)
+        echo '#!/bin/sh
+exec /usr/bin/palemoon "$@"
+' > usr/local/bin/defaultbrowser
+        ;;
+esac
+
 chmod 755 usr/local/bin/defaultbrowser
+cp usr/local/bin/defaultbrowser usr/local/bin/defaulthtmlviewer
```
dimkr: Better use `cp -p` or change to `chmod 755 usr/local/bin/defaultbrowser usr/local/bin/defaulthtmlviewer` after the copying

dimkr (approved):
Looks good to me - again, I have no merge rights

lakshayrohila:
> Looks good to me - again, I have no merge rights

But your reviews are great and helpful! Thanks :+1:!

@peabee, please merge this if you have no objections.

peabee:
> Looks good to me - again, I have no merge rights

@puppylinux-woof-CE/members - please note that this is not a sustainable situation

PR #4242, opened 2024-02-15
Add `tldr_c_client` petbuild and add it to Jammy64

lakshayrohila:
Tested and working in Jammy64. I'll keep this as a draft for few days, in case `tldr-c-client` makes a new release with the fix I proposed for bash auto-completion (for now, it is an open PR there; the same file is also copied in woof-CE directly, which would be removed as soon as the former repo is updated).

### DIFF
--- a/woof-code/3builddistro
+++ b/woof-code/3builddistro
@@ -1073,10 +1073,6 @@ else
 fi
 echo
 
-## TESTS
-echo "TLDR_C_CLIENT TESTS"
-find ./ -name *tldr*
-
 #-------------------------
 #before building puppy.sfs from rootfs-complete, check for any invalid symlinks
 #and move them to the devx...
--- a/woof-code/3builddistro
+++ b/woof-code/3builddistro
@@ -1073,6 +1073,10 @@ else
 fi
 echo
 
+## TESTS
+echo "TLDR_C_CLIENT TESTS"
+find ./ -name *tldr*
+
 #-------------------------
 #before building puppy.sfs from rootfs-complete, check for any invalid symlinks
 #and move them to the devx...
--- a/woof-code/rootfs-petbuilds/tldr_c_client/pet.specs
+++ b/woof-code/rootfs-petbuilds/tldr_c_client/pet.specs
@@ -1,0 +1,1 @@
+tldr-1.6.1|tldr|1.6.1|||84K||tldr-1.6.1.pet||C command-line client for tldr pages|puppy|||
--- a/woof-code/rootfs-petbuilds/tldr_c_client/petbuild
+++ b/woof-code/rootfs-petbuilds/tldr_c_client/petbuild
@@ -1,0 +1,21 @@
+# see deps.sh in the downloaded archive's tldr-c-client-*/ folder for dependencies depending on your target system
+
+download() {
+    [ -f tldr_c_client-1.6.1.tar.gz ] || wget -t 3 -T 60 -O tldr_c_client-1.6.1.tar.gz https://github.com/tldr-pages/tldr-c-client/archive/refs/tags/v1.6.1.tar.gz
+}
+
+build() {
+    tar -xzf tldr_c_client-1.6.1.tar.gz
+    cd tldr-c-client-1.6.1/
+
+    make
+    make install
+
+    # enable autocompletion for bash
+
+    # uncomment the following lines when https://github.com/tldr-pages/tldr-c-client/pull/113 is merged and new release is made
+    # the fix is currently directly included in woof-CE
+
+    # mkdir -p /usr/share/bash-completion/completions/
+    # mv autocomplete/complete.bash /usr/share/bash-completion/completions/tldr
+}
--- a/woof-code/rootfs-petbuilds/tldr_c_client/sha256.sum
+++ b/woof-code/rootfs-petbuilds/tldr_c_client/sha256.sum
@@ -1,0 +1,1 @@
+1a2aa8e764846fad1f41a0304e28416f5c38b6d3a3131ad1e85116749aec34ba  tldr_c_client-1.6.1.tar.gz
--- a/woof-code/rootfs-petbuilds/tldr_c_client/usr/share/bash-completion/completions/tldr
+++ b/woof-code/rootfs-petbuilds/tldr_c_client/usr/share/bash-completion/completions/tldr
@@ -14,10 +14,10 @@ _tldr_complete() {
     local word="${COMP_WORDS[COMP_CWORD]}"
     local cmpl=""
     if [[ "$word" == "--"* ]] || [ -z "$word" ]; then
-        cmpl=$'--version\n--help\n--update\n--clear-cache\n--platform\n--render'
+        cmpl=$'--help\n--color\n--platform\n--render\n--update\n--version\n--clear-cache\n--verbose\n--list'
     elif [[ "$word" == "-"* ]]; then
-        cmpl=$'\n-v\n-h\n-u\n-c\n-p\n-r'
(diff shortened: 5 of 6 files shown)


peabee:
Are all pets going to be only available as petbuilds? This means that they are not available to retrofit to existing systems already installed - makes them only available for new builds. I understand that some petbuilds do need to be built in the environment of the particular lib mix of the target system - but is this app one of those? What does it depend on? 

Also, the more things are put into __build.conf it diminishes DISTRO_PKGS_SPECS-... as the definition of the system build.

Would it not be better as a .pet uploaded to:
https://distro.ibiblio.org/puppylinux via
Packages-puppy-bionic64-official
Packages-puppy-xenial64-official
Packages-puppy-tahr64-official
Packages-puppy-fossa64-official

lakshayrohila:
> Are all pets going to be only available as petbuilds? This means that they are not available to retrofit to existing systems already installed - makes them only available for new builds. I understand that some petbuilds do need to be built in the environment of the particular lib mix of the target system - but is this app one of those? What does it depend on?

The dependencies are listed here: https://github.com/tldr-pages/tldr-c-client?tab=readme-ov-file#requirements

> Also, the more things are put into __build.conf it diminishes DISTRO_PKGS_SPECS-... as the definition of the system build.
> 
> Would it not be better as a .pet uploaded to: https://distro.ibiblio.org/puppylinux via Packages-puppy-bionic64-official Packages-puppy-xenial64-official Packages-puppy-tahr64-official Packages-puppy-fossa64-official

I do not have access to that repo; don't know how to get it uploaded there. Also, I followed this ideology - keep most things in the same spot, i.e. woof-CE; since I understood that maintainers (@dimkr) used this approach. But, maybe that is not the case? (that is why we require clear contribution guidelines and vision of maintainers listed somewhere)

Also, wouldn't it be better for Jammy64 to have a Jammy's repo, instead of having its applications put in other repos?

---

This version is a smaller version of the `tldr` package available in official Ubuntu repo. I was requested by the forum member Clarity to add tldr to Jammy64, but since the official one requires `git` as an essential requirement for it, it is quite large. This version adds less than 80kb size to the ISO.

---

I am willing to make a pet for this application and get it uploaded in one of the repos, but please guide me through how I can do that. Thanks!

dimkr:
As far as I see, things that become .pet packages that sit in some repo eventually get frozen in time. The source code is lost or fails to build on newer Puppy releases, so the package is impossible to update. A reproducible package built from source can be updated if a bug is found, and it's future proof. For Adding rootfs-petbuilds and converting all core Puppy packages to petbuilds allowed me to port everything to GTK+ 3 and save Puppy from extinction once X.Org dies.

peabee:
https://distro.ibiblio.org/puppylinux/pet_packages-jammy64/
is created as link to:
https://distro.ibiblio.org/puppylinux/pet_packages-fossa64/

Will need a modified (pruned - only jammy64 entries)
https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-distro/x86_64/ubuntu/focal64/Packages-puppy-fossa64-official
renamed as
Packages-puppy-jammy64-official
in
https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-distro/x86_64/ubuntu/jammy64

I and others(?) have upload details

peabee:
I don't see any run-time dependencies in https://github.com/tldr-pages/tldr-c-client?tab=readme-ov-file#requirements

only build-time dependencies

lakshayrohila:
> I don't see any run-time dependencies in https://github.com/tldr-pages/tldr-c-client?tab=readme-ov-file#requirements
> 
> only build-time dependencies

IMO, `libcurl` and `libzip` should be the runtime dependencies; otherwise I don't think it will make any sense of including their `dev` counterparts in their documentation. These are already present in Jammy64.

I also found that their `Makefile` was reporting an error when `git` is not present, although I see that their Makefile was making checks if git is present or not. But perhaps, `git` is only required for building version information on compile time:

https://github.com/tldr-pages/tldr-c-client/blob/60b9a977b620ac64c9446eb348ca266149bf12d1/Makefile#L36-L43

Searching for the word `git` in their Makefile does not return with any more results. Also, this application works fine without `git` on runtime.

lakshayrohila:
> https://distro.ibiblio.org/puppylinux/pet_packages-jammy64/ is created as link to: https://distro.ibiblio.org/puppylinux/pet_packages-fossa64/
> 
> Will need a modified (pruned - only jammy64 entries) https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-distro/x86_64/ubuntu/focal64/Packages-puppy-fossa64-official renamed as Packages-puppy-jammy64-official in https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-distro/x86_64/ubuntu/jammy64
> 
> I and others(?) have upload details

I would love to work on this, but due to my personal life commitments, I won't be able to make many changes for about next one month. I will do this when I return :+1:.

dimkr:
If you're busy and won't be able to maintain this package after this first upload, ask yourself if it's worth it. People who want to use tldr instead of man pages (probably a tiny minority) can use https://tldr.inbrowser.app/ or just install tldr themselves, and every time you add something just because 1-2 users asked for it you're adding more stuff that can break or rot over time.

peabee:
Sound advice!

peabee:
> IMO, `libcurl` and `libzip` should be the runtime dependencies; otherwise I don't think it will make any sense of including their `dev` counterparts in their documentation. These are already present in Jammy64.

The "strength" of those dependencies can be checked by seeing if your app runs on older systems like BionicPup64 and FossaPup64.....

lakshayrohila:
> If you're busy and won't be able to maintain this package after this first upload, ask yourself if it's worth it. People who want to use tldr instead of man pages (probably a tiny minority) can use https://tldr.inbrowser.app/ or just install tldr themselves, and every time you add something just because 1-2 users asked for it you're adding more stuff that can break or rot over time.

Thanks for the tip :+1:, and sorry for the late response. Didn't read your message earlier.

Considering the time I've invested in this one (maybe just under 1 hour, but still I have put in some), and the size it adds (<100KBs), I don't think it will be great to reject this application even if no one wants it (although, there will be people who may require such application). This won't be required in a NoblePup (if we'll want to include this package in it) since `tldr-py` will have required updates; but in Jammy64, due to the reasons I gave earlier, this is the most viable version of tldr out there.

But since this is a small addition, I also think a PET package might just be great for this application (and I will create that package :+1:, just not now since I will be busy the next few days).

Now, for the maintenance, you see it's just a simple client for getting `tldr` online pages - which wouldn't need much updates. In fact, their repo hasn't got any new commits after December 2023; so this is rather currently not in the development stage (I mean not fit for regular users, or lacks certain features that are provided by other versions of tldr client). Installation process for this program is easy as well, just run `make install`.

I should also say that I've come to develop a liking for this program, so I will myself like that it is included in Jammy64 :grin:. The versions apt provides are out of question: one is big, one is outdated. Including this in Puppy itself will also reduce my work of compiling the program for each Puppy VM I create (directly translating the Hindi idiom: One arrow, two birds :smile:).


> The "strength" of those dependencies can be checked by seeing if your app runs on older systems like BionicPup64 and FossaPup64.....

Thanks for the comment! I will test that.

lakshayrohila:
Closing the PR in favour of creating a pet package for it instead.

peabee:
The TLDR menu item opening the web based app provided by radky in BookwormPup64-10.0.5 seems an entirely reasonable approach to providing this facility.....

lakshayrohila:
Thanks, I'll look into it. I am currently working on adding `dunst` and `dunst-config` in Jammy64, so probably may not be able to take a look at it currently.

PR #4245, opened 2024-03-03
Fix breakage of bash in noble

dimkr:
jammy:

```
/bin/bash
/usr/bin/bashbug
```

noble:

```
/usr/bin/bash
/usr/bin/bashbug
```

The package template breaks bash because it replaces /usr/bin/bash with a symlink to /bin/bash, so it's a broken symlink or a recursive symlink (depending on whether or not /bin is a symlink to /usr/bin).

### DIFF
--- a/woof-code/packages-templates/bash_FIXUPHACK
+++ b/woof-code/packages-templates/bash_FIXUPHACK
@@ -1,6 +1,6 @@
 DISABLE_POST_INSTALL_SCRIPT=yes
 
-if [ "$DISTRO_BINARY_COMPAT" = "void" ]; then
+if [ -f usr/bin/bash -a ! -e bin/bash -a ! -L bin/bash ] ; then
 	mkdir -p bin
 	mv -f usr/bin/bash bin
 


woof-code/packages-templates/bash_FIXUPHACK
```diff
@@ -1,6 +1,6 @@
 DISABLE_POST_INSTALL_SCRIPT=yes
 
-if [ "$DISTRO_BINARY_COMPAT" = "void" ]; then
+if [ -f usr/bin/bash && ! -e bin/bash -a ! -L bin/bash ] ; then
```
lakshayrohila: Shouldn't that be `[ -f usr/bin/bash -a ! -e bin/bash -a ! -L bin/bash ]`, or even better `[ -f usr/bin/bash ] && [ ! -e bin/bash ] && [ ! -L bin/bash ]`

dimkr: Yep

lakshayrohila:
As far as I can see, we do the following steps to fix the problem:

```lang-sh
mkdir -p bin
mv -f usr/bin/bash bin
ln -sf bash bin/sh
```

But that leaves no bash in `usr/bin`. Is that supposed to be OK?

dimkr:
> But that leaves no bash in usr/bin. Is that supposed to be OK?

This depends on `USR_SYMLINKS` - if `yes`, this doesn't matter because /bin and /usr/bin have the same contents.

Otherwise, line 17 takes care of this.

peabee:
I will try a NoblePup32 usrmerge build with:
if [ -f usr/bin/bash -a ! -e bin/bash -a ! -L bin/bash ] ; then

and report

peabee:
Results from test build:
1. busybox petbuild completes with #4247 but still puts busybox into /usr/bin whereas /support/busybox_symlinks.sh line 14 requires it to be in /bin
2. bash package template works OK - this PR
3. the rootfs-complete is not viable when first constructed - usrmerge rootfs-complete 1 must be run to allow the build to continue

dimkr:
> the rootfs-complete is not viable when first constructed - usrmerge rootfs-complete 1 must be run to allow the build to continue

Makes sense, it needs to run immediately after of copying of packages into rootfs-complete. Then, the busybox symlink issue doesn't matter because /bin and /usr/bin have the same contents.

peabee:
3builddistro changes??:

insert before line 472:
[ "$USR_SYMLINKS" = "yes" ] && usrmerge sandbox3/rootfs-complete 1

delete line 1096

PR #4246, opened 2024-03-03
Add noble64 kernel built from the Ubuntu source

dimkr:
Has ad1d655 and aa1c548 applied, and forces the traditional `CONFIG_SQUASHFS_COMPILE_DECOMP_MULTI=y`

PR #4248, opened 2024-03-04
Add noble64 template

dimkr:
Depends on:
- #4247
- #4246
- #4245

Almost 1:1 bump of jammy (with @lakshayrohila's work) to noble: updated packages and Ubuntu 24.04's 6.8 kernel. I had to drop `claws-mail` because it refuses to build, and didn't want to bump it to a newer version that might not work in other build configurations that use older libraries.

Problems so far:
- [x] Boots very slowly
- [ ] Ugly /{,s}bin/usr-is-merged directories
- [x] Missing busybox symlinks - at least ash and login

### DIFF
--- a/woof-distro/x86_64/ubuntu/noble64/_00build.conf
+++ b/woof-distro/x86_64/ubuntu/noble64/_00build.conf
@@ -1,0 +1,168 @@
+#
+#  persistent configuration options
+#
+#  see also DISTRO_SPECS DISTRO_PET_REPOS DISTRO_COMPAT_REPOS-*
+#
+#  **NOTE**: check the original file every once in a while
+#            settings might be added or removed...
+#
+
+# 2createpackages
+STRIP_BINARIES=no
+
+## Kernel tarballs repo URL for choosing/downloading kernel
+KIT_KERNEL_REPO_URL=http://distro.ibiblio.org/puppylinux/huge_kernels
+
+### Kernel tarball URL - avoid being asked questions about downloading/choosing a kernel
+#KERNEL_TARBALL_URL=http://distro.ibiblio.org/puppylinux/huge_kernels/huge-5.10-kernel-kit.tar.bz2
+
+## an array of generically named programs to send to the ADRIVE, FDRIVE, YDRIVE
+## ADRV_INC="abiword gnumeric goffice"
+ADRV_INC=""
+## YDRV_INC=""
+YDRV_INC=""
+## FDRV_INC="" #this one is very experimental and it's recommended to be left unset
+FDRV_INC="linux-firmware firmware-sof-signed intel-microcode amd64-microcode"
+
+## Include kernel-kit generated FDRIVE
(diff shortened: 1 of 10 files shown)


peabee:
`Missing busybox symlinks - at least ash and login`

That is what I also reported for NoblePup32 - it is because the petbuild puts busybox into /usr/bin but /support/busybox_symlinks.sh expects it to be in /bin

dimkr:
Looks good other than the weird directories issue, which can be solved later.

woof-distro/x86_64/ubuntu/noble64/_00build.conf
```diff
@@ -0,0 +1,168 @@
...
+## Include the windows puppy installer LICK by Luke Lorimer aka <noryb009>
+LICK_IN_ISO=yes
+
+## compression method to be used (SFS files)
+#SFSCOMP='-comp xz -Xbcj x86 -b 512K'
+#SFSCOMP='-comp xz -Xbcj arm,armthumb -b 512K'
+#SFSCOMP='-comp gzip'
+#SFSCOMP='-noI -noD -noF -noX'
+#SFSCOMP='-comp xz -Xbcj x86 -b 256K -no-exports -no-xattrs'
+SFSCOMP='-comp zstd -Xcompression-level 19 -b 256K -no-exports -no-xattrs'
```
lakshayrohila, line 58: I had this doubt (not any issue) as to why we use zstd compression while xz provides much more space saving.

I tested this earlier for Jammy64, devx saved 16MB, docx saved 0.2MB, main iso saved 65MB, nls saved 11.5MB.

Is there a specific reason why we use chose zstd in newer puppy versions (since I don't think fossa uses it).

dimkr: The difference in size is ~10% but zstd is many times faster to decompress. With xz, you get a slightly smaller Puppy, but applications start much slower, CPU consumption is higher and boot times are worse.

dimkr: zstd vs xz is a question of tradeoffs. Storage is easy and cheap to expand or upgrade, and most old computers can afford a 10% increase in the size of Puppy even without any hardware change, but the CPU can't be upgraded. Therefore, if I need to decide what to compromise on, it must be size and not CPU consumption. I think it's a shame that many Puppy developers choose xz and blindly assume that smaller size equals better performance, because this is a very obvious case where this assumption is false.

IdfbAn:
> I had to drop `claws-mail` because it refuses to build

I had that problem too! Glad to see that it wasn't me doing something wrong.

dimkr:
> I had that problem too! Glad to see that it wasn't me doing something wrong.

Nope, it's just too old, but the latest version can be too new for existing build configurations that have old libraries. I'm leaving this for somebody else to fix because I'm not interested in maintaining the claws-mail petbuild.

PR #4255, opened 2024-03-15
Jammy64: Add `conky-std` & `Conky_GTK_2024`.

lakshayrohila:


peabee:
Why is the conky.desktop being removed? If it doesn't fit with Puppy then it is better to edit it or to disable it with NoDisplay=true

The fixup hack would be better in _0build.conf on a per build basis?

peabee:
It would be better to update DISTRO_PET_REPOS to use the Github version of Packages-puppy-tahr64-official
and revert the revert!

lakshayrohila:
@peabee,

> Why is the conky.desktop being removed? If it doesn't fit with Puppy then it is better to edit it or to disable it with NoDisplay=true
> 
> The fixup hack would be better in _0build.conf on a per build basis?

The conky.desktop will only run conky, but not with the conky-gtk's configs, but the default conky configs. Maybe, we need a `Conky_GTK_2024_FIXUPHACK`? Or else, it is also easy to, as you mentioned, add it to __0build.conf (I'll chose the later).

> It would be better to update DISTRO_PET_REPOS to use the Github version of Packages-puppy-tahr64-official

I was not able to get the jammy repo showing in PPM, and thought I messed something up (although I think it was just disabled in PPM).

I think woof-CE already uses the Github version of Packages-puppy-tahr64-official:
```
z|https://raw.githubusercontent.com/puppylinux-woof-CE/woof-CE/${WCE_BRANCH}/woof-distro/x86_64/ubuntu/Packages-puppy-tahr64-official|z
```

Also, in that commit, I added Packages-puppy-jammy64-official, moving `Conky_GTK_2024` from there to the new Packages-puppy-jammy64-official.

If you allow me to add `Packages-puppy-jammy64-official`, then I can revert the revert. Also, I am unsure if `Conky_GTK_2024` really needs to go from Packages-puppy-tahr64-official to Packages-puppy-jammy64-official.

Thanks :+1:

peabee:
This Pull Request has got very big and complicated..... could we start again with just the necessary changes?

Presumably the same changes need to be applied to NoblePup64 as well....

dimkr:
> This Pull Request has got very big and complicated

It only adds 4 lines.

If you're worried about the number of commits, you can squash merge it.

lakshayrohila:
@peabee,
Sorry for late reply. I keep adding various commits for trials - some for test due to network bandwidth restrictions - and then remove it.

So, all you need to see is the "files changed" tab, and you can ignore the commits (for my next PRs as well). Squash merge removes all the unnecessary commits.

lakshayrohila:
As far as noble64 is concerned, it will need additional tests before adding this.

PR #4256, opened 2024-03-17
Update pipewire_FIXUPHACK: Added bluez workaround

rizalmart:
Added bluez workaround

See [https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/3828](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/3828)

### DIFF
--- a/woof-code/packages-templates/pipewire_FIXUPHACK
+++ b/woof-code/packages-templates/pipewire_FIXUPHACK
@@ -5,3 +5,11 @@ cat << EOF > usr/bin/apulse
 exec "\$@"
 EOF
 chmod 755 usr/bin/apulse
+
+cat << EOF > /etc/pipewire/wireplumber.conf.d/bluez-nologind.conf
+wireplumber.profiles = {
+  main = {
+    monitor.bluez.seat-monitoring = disabled
+  }
+}
+EOF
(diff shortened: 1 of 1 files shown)


peabee:
@dimkr - any comments?

dimkr:
@peabee Frankly - I don't see the need for this, because Bluetooth audio works great on dpup without this

peabee:
@rizalmart 
Is this fix specific to your builds?
If so it should be fixed by a build specific mechanism rather than being applied to all builds at the Woof-CE level.
Please let us know why you need it and why you think it needs to be a Woof-CE level fix. Thanks.

lakshayrohila:
Whenever I try to connect to my Bluetooth dev in Jammy64, I get the `br-connection-timeout` error. I'll see if this fixes it for me and report.

rizalmart:
> @rizalmart Is this fix specific to your builds? If so it should be fixed by a build specific mechanism rather than being applied to all builds at the Woof-CE level. Please let us know why you need it and why you think it needs to be a Woof-CE level fix. Thanks.

Because wireplumber-0.5 onwards now looks for active login session in order to make bluez work on pipewire otherwise bluetooth device will not work on pipewire

See [https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/3828](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/3828)

rizalmart:
> @peabee Frankly - I don't see the need for this, because Bluetooth audio works great on dpup without this

Because you are using lower version of wireplumber.

peabee:
This should be a change to wireplumber_FIXUPHACK ? rather than pipewire_FIXUPHACK

PR #4270, opened 2024-04-10
Update VoidPup64's kernel to 6.6 and VoidPup32's to 6.1

IdfbAn:
Note that I've only tested VoidPup64, and not very much, but it seems to be working as usual, at least on my computer.

### DIFF
--- a/.github/workflows/void-voidpup.yml
+++ b/.github/workflows/void-voidpup.yml
@@ -11,10 +11,10 @@ jobs:
         include:
           - arch: x86_64
             compat-distro-version: voidpup64
-            kernel: 5.15.x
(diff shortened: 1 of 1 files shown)


peabee:
Hi there
I can see where you're going and why you might want to update kernels - but this is the sort of change which all current users of VoidPup would have to be happy with. Perhaps float it as a proposal on the forum?
It should really wait until a new version of VoidPup is created.....
There is nothing to stop the new 6.x kernel being downloaded separately and replacing the 5 series kernels.....

IdfbAn:
> Perhaps float it as a proposal on the forum?

https://forum.puppylinux.com/viewtopic.php?t=11215

PR #4272, opened 2024-04-18
Reduce boot time by ~20%

dimkr:
In a very non-scientific test, boot time drops from ~25s to ~20s in a VM on my old laptop. I did some basic profiling and these three things are the top 3 time wasters in rc.sysinit.

(With many more optimizations, a dpup built using my woof-CE fork achieves 15-17s, but it also doesn't have X.Org, xorgwizard and other sources of slowness, so I don't think I have more optimizations to upstream)

### DIFF
--- a/woof-code/3builddistro
+++ b/woof-code/3builddistro
@@ -870,6 +870,8 @@ if [ $CROSSBUILD -eq 0 -o -n "$QEMU" ];then
 	# 'chooselocale' is called from /etc/rc.d/rc.country at first boot and
 	# pre-creating this table speeds things up. 111123
 	chroot rootfs-complete /usr/sbin/chooselocale composeonly
+	chroot rootfs-complete localedef -f ISO-8859-1 -i en_US --no-archive en_US
+	chroot rootfs-complete localedef -f UTF-8 -i en_US --no-archive en_US.utf8
 else
 	#*** cross-build
 	touch rootfs-complete/etc/gtk-2.0/gtk.immodules
--- a/woof-code/rootfs-skeleton/etc/rc.d/rc.sysinit
+++ b/woof-code/rootfs-skeleton/etc/rc.d/rc.sysinit
@@ -420,14 +420,8 @@ fi
 echo 5000 > /sys/module/block/parameters/events_dfl_poll_msecs
 
 #replay uevents from /sys...
-for ONEMODALIAS in $MODALIASES
-do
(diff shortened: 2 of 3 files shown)


peabee:
Presumably the changes have been fully tested on real hardware and confirmed to work OK with all current builds and kernels?
I haven't seen any problem reports about slow booting - although I guess being a bit quicker would be nice.

dimkr:
> Presumably the changes have been fully tested on real hardware

Also tested on real hardware, but it's harder to measure boot time (too many external variables)

peabee:
@gyrog @techrockedge any views?

gyrog:
> Also tested on real hardware, but it's harder to measure boot time (too many external variables)
Aren't the numbers at the beginning of a "dmesg" line the seconds since the begining of the boot?
Maybe they would provide some indication of the speed of the boot.

I don't think anyone would complain about faster boots.
So, provided it works reliably, and it's not slower, then it's probably worth a try.

I have no idea if the change away from a "udev replay loop" creates any issues for any Puppies, e.g. a re-woofed FossaPup,  or spup.

techrockedge:
"no idea if the change away from a "udev replay loop" creates any issues for any Puppies, e.g. a re-woofed FossaPup, or spup"

I have not tested what the effects would be, but if we can determine that Fossapup64 build will not be effected in  it's boot cycle this commit should be merged.

dimkr:
Puppy is the only distro that has this weird loop, which (as far as I see) tries to do the same thing as the `udevadm trigger` that follows but slower (because it's a shell script). udev is udev - if other distros start udev without this weird loop, it should work equally well on Puppy.

peabee:
Pull request merged - will be used in future builds on 2 May - these need to be downloaded and tested on real hardware I suggest. I will do this for my builds but leave it up to others for legacy builds etc.

peabee:
@ozsouth is reporting issues which _may_ arise from this commit. He says using S15Pup64 & VoidPup64:

> I've since found that the issue continues from 240505 on, & worse still has intermittent consequences - Xkeyboard appears on taskbar if you blacklist hid_multitouch, & appears & disappears on alternate X restarts if I run my unload-reload script. Without hid_multitouch, most of my touchpads fail, but not all. I tried for days to solve this, but my unload-reload script seemed the only (mostly) ok solution. ..... I'm sure this is related to woof-ce updates after 240420 issued. 240323 doesn't have issues

PR #4278, opened 2024-04-27
If using an external monitor, turn off the internal display on lid close

dimkr:
This PR makes Puppy:
1) Turn off the internal display and disable suspend on lid close, **if there's at least one external monitor**
2) Turn on the internal display on lid open, in case the laptop was suspended while connected to an external monitor, then disconnected and opened

### DIFF
--- a/woof-code/rootfs-packages/acpid_busybox/etc/acpi.map
+++ b/woof-code/rootfs-packages/acpid_busybox/etc/acpi.map
@@ -1,0 +1,4 @@
+"EV_KEY"	0x01	"KEY_POWER"	116	1	"button/power PWRF 00000080"
+"EV_KEY"	0x01	"KEY_POWER"	116	1	"button/power PWRB 00000080"
+"EV_SW"		0x05	"SW_LID"	0	1	"button/lid LID0 00000080"
+"EV_SW"		0x05	"SW_LID"	0	0	"button/lid LID0 00000080"
--- a/woof-code/rootfs-packages/acpid_busybox/etc/acpi/actions/suspend.sh
+++ b/woof-code/rootfs-packages/acpid_busybox/etc/acpi/actions/suspend.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 # suspend.sh 28sep09 by shinobar
 # 12feb10 pass poweroff
 # 23apr12 fix was not suspend from acpi_poweroff.sh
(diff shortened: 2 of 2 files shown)


woof-code/rootfs-packages/acpid_busybox/etc/acpi.map
```diff
@@ -0,0 +1,4 @@
+"EV_KEY"	0x01	"KEY_POWER"	116	1	"button/power PWRF 00000080"
+"EV_KEY"	0x01	"KEY_POWER"	116	1	"button/power PWRB 00000080"
+"EV_SW"		0x05	"SW_LID"	0	1	"button/lid LID0 00000080"
```
dimkr, line 3: The first 3 are hardcoded in busybox acpid.c, if acpi.map is missing - by default, it runs /etc/acpi/LID/00000080 only on lid close and the fourth line runs the same script on lid open

woof-code/rootfs-packages/acpid_busybox/etc/acpi/actions/suspend.sh
```diff
@@ -4,6 +4,29 @@
...
 #20140526 shinobar: avoid multiple run
 #20140629 shinobar: ACPI_CONFIG
+case "$(awk '{print $2}' /proc/acpi/button/lid/LID0/state)" in
+open)
+  [ "$DISTRO_TARGETARCH" = "x86" ] && exit 0
+  # turn on the internal display on lid open
+  if [ -n "$WAYLAND_DISPLAY" ]; then
+    wlr-randr --output eDP-1 --on
+  elif [ -n "$DISPLAY" -a -z "`pidof -s xlock`" ]; then
+    xrandr --output eDP-1 --on
```
dimkr, line 14: - [ ] Test under X.Org
- [ ] Verify that `eDP-1` is a good assumption on various old laptops

woof-code/rootfs-packages/acpid_busybox/etc/acpi/actions/suspend.sh
```diff
@@ -35,6 +58,9 @@ if [ "$DISTRO_TARGETARCH" = "x86" ]; then
 	do
 		mount | grep -q "^$USB" && rm -f "$LOCKFILE" && exit 0
 	done
+else
+	# do not suspend if at least one monitor is enabled
+	[ -n "`grep -l '^enabled$' /sys/class/drm/*/enabled | head -n 1`" ] && rm -f "$LOCKFILE" && exit 0
```
dimkr: - [x] Ensure that xrandr disables monitors with `--off`

woof-code/rootfs-packages/acpid_busybox/etc/acpi/actions/suspend.sh
```diff
@@ -4,6 +4,29 @@
...
+  fi
+  exit 0
+  ;;
+closed)
+  if [ "$DISTRO_TARGETARCH" != "x86" ]; then
+    # turn off the internal display on lid close
+    if [ -n "$WAYLAND_DISPLAY" ]; then
+      wlr-randr --output eDP-1 --off
+    elif [ -n "$DISPLAY" ]; then
+      xrandr --output eDP-1 --off
```
dimkr: - [x] Test under X.Org
- [x] Verify that `eDP-1` is a good assumption on various old laptops

peabee:
Ready to merge? Still seems to be being worked on?

dimkr:
> Ready to merge?

Up to you. Puppy always had this weird behavior of not suspending on lid close at all or suspending on lid close even when connected to an external monitor, so I wonder why nobody is either complaining about this or commenting here.

> Still seems to be being worked on?

Only a bug fix after 2w without activity 😸 

EDIT: and another one today

peabee:
I guess not that many people use Puppy in that configuration - I certainly don't.... more likely to be users on the forum than here..... MochiMoppel seems to be testing but is not on Github?
I will wait until you confirm that all bugs are squashed and tested.

dimkr:
> I guess not that many people use Puppy in that configuration

Puppy doesn't have a display settings panel where you can move and rotate your displays, I bet very few people actually use Puppy with multiple monitors, and those that do just use the default arrangement.

> I will wait until you confirm that all bugs are squashed and tested.

Up to you, but don't hold your breath: without a display settings panel and a daemon that constantly re-applies the user's chosen settings when monitors are connected and disconnected, there's no way to make Puppy handle multiple monitors to the same standard as GNOME or KDE. This PR will never be 'ready'.

Wayland has kanshi, wdisplays and other tools, so it would be best to adopt one of these instead of building an in-house solution just for Puppy. And developing a solution for X.Org is a waste of time at this point.

This PR is merely a 80/20 solution for the use case of people who use a laptop with external monitor(s) and want to close the lid. It will mess up the display settings of people who somehow configure their monitors, once they reopen the lid, which is practically nobody (again, because Puppy doesn't have a display settings panel: those who configure their monitors probably have their ugly `xrandr` startup script).

peabee:
I think I am unsure whether this is a PR that you would like to see applied to all future Pup builds or whether it is a private optional PR just to be applied to your builds. Can it be supplied as an optional add-on (.pet?)? It could then be tested more widely on the forum?

dimkr:
https://forum.puppylinux.com/viewtopic.php?p=122436

peabee:
??

peabee:
Ready to merge?

dimkr:
> Ready to merge?

Up to you, I haven't changed anything since May 11 and I think it's an improvement over the current state (although it's not perfect)

peabee:
I have no view or need.

PR #4280, opened 2024-05-01
Update pipewire_FIXUPHACK: Add socket file

rizalmart:
for run-as-spot instead of running a separate pipewire process. Just use the socket file instead for accessing the sound server

### DIFF
--- a/woof-code/packages-templates/pipewire_FIXUPHACK
+++ b/woof-code/packages-templates/pipewire_FIXUPHACK
@@ -5,3 +5,177 @@ cat << EOF > usr/bin/apulse
 exec "\$@"
 EOF
 chmod 755 usr/bin/apulse
+
+#Add pipewire socket file
+cat << EOF > ./etc/pipewire/pipewire-pulse.conf
+# PulseAudio config file for PipeWire version "0.3.77" #
+#
+# Copy and edit this file in /etc/pipewire for system-wide changes
+# or in ~/.config/pipewire for local changes.
+#
+# It is also possible to place a file with an updated section in
+# /etc/pipewire/pipewire-pulse.conf.d/ for system-wide changes or in
+# ~/.config/pipewire/pipewire-pulse.conf.d/ for local changes.
+#
+
+context.properties = {
+    ## Configure properties in the system.
+    #mem.warn-mlock  = false
+    #mem.allow-mlock = true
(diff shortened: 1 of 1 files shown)


peabee:
For this and #4281 and #4282 ..... (should all be in one PR?)
Have these changes been tested? In which build?
Needs review by a Pipewire expert - @dimkr ?

dimkr:
I understand what this PR does but I don't understand the motivation. The current PipeWire setup runs a single instance (of pipewire+pipewire-pulse+wireplumber) for spot, and both spot and root use this one instance. This prevents weird things like volume inconsistency between applications running as root and spot, and saves resource (it uses RT scheduling policy, so having only one process saves CPU in addition to RAM). What needs/should to be changed and why?

rizalmart:
> I understand what this PR does but I don't understand the motivation. The current PipeWire setup runs a single instance (of pipewire+pipewire-pulse+wireplumber) for spot, and both spot and root use this one instance. This prevents weird things like volume inconsistency between applications running as root and spot, and saves resource (it uses RT scheduling policy, so having only one process saves CPU in addition to RAM). What needs/should to be changed and why?

to simplify the process running pipewire as other user complicates the process. Instead of running the pipewire as other user just run as current user and share pipewire to other users through socket.

dimkr:
> to simplify the process running pipewire as other user complicates the process.

Simplify in what sense?

I don't think there's a lot to simplify, it's a single, short block:

https://github.com/puppylinux-woof-CE/woof-CE/blob/e2a6ee4927b53f163d54bccc207b8b92d20367da/woof-code/rootfs-skeleton/root/.xinitrc#L51-L69

https://github.com/puppylinux-woof-CE/woof-CE/blob/e2a6ee4927b53f163d54bccc207b8b92d20367da/woof-code/rootfs-petbuilds/labwc-puppy/usr/bin/startlabwc#L21-L32

rizalmart:
Then why running pipewire as other user?

dimkr:
> Then why running pipewire as other user?

As I explained, this code runs PipeWire, WirePlumber, etc' as spot so spot and root can share the same instance. These are CPU intensive processes, and they eat extra RAM. One instance of each saves resources and ensures consistency (i.e. pavucontrol that runs as spot and root show and set the same configuration values).

PR #4291, opened 2024-05-19
find_cat: use categories.dat from root-fs/skeleton, not system-wide.

### LINKED ISSUE #4290: debdb2pupdb and find_cat depends on system-wide /usr/local/petget/categories.dat
`debdb2pupdb` and `find_cat` are the two core programs used in `0setup` to convert foreign package database format to puppy database format. 

It turns out that both `debdb2pupdb` and `find_cat` have an unlisted dependency on system-wide /usr/local/petget/categories.dat (as the file itself says):
```
#130124 format modified to be bash-include compatible.
#130124 loaded at runtime by 'debdb2pupdb.bac' and 'find_cat.bac' in Woof.
```

But system-wide  `/usr/local/petget/categories.dat` only exists in Puppy Linux. It doesn't exist in other distros. Hence, running `debdb2pupdb` and `find_cat` will fail. In other words, running `0setup` will fail, without any clear error message. Instead, it will only say:
```
FATAL: failed find to find $pkgname
```
which is very _misleading_. The cause of this "FATAL" message is because the converted pup package database is _empty_, thanks to debdb2pupdb and find_cat failing silently.

This dependency on `/usr/local/petget/categories.dat` is not necessary, because woof-CE itself carries this file in `./woof-code/rootfs-skeleton/usr/local/petget/categories.dat`. 

My suggestion is to modify merge2out to copy `./woof-code/rootfs-skeleton/usr/local/petget/categories.dat` and put it to 'woof-out' directory, and to modify `find_cat`, `debdb2pupdb` (and possible other package database converter that depends on categories.dat, so far I've only tested `debdb2pupdb`) to use the `./categories.dat` inside `woof-out` instead of reaching out to system-wide `/usr/local/petget/categories.dat`.

If this is done, then `0setup` will run on any reasonable Linux distro. This will be the first step to enable building Puppies on mainstream distros.  I'd be happy to submit a PR but I can't see the source for find_cat and debdb2pupdb.

jamesbond3142:
This enables 0setup to run successfully on non-Puppy platform.

Only x86_64 and x86 is updated, the ARM find_cat is not updated and hence still broken.


jamesbond3142:
Sorry @peabee, I fixed the source but not the binaries, so you're pulling the wrong binaries.
I'm going to issue another pull request with the correct binaries.

PR #4293, opened 2024-06-04
PPM: Code refinement, integrates web-browser advice window

rserwin1:
Finalizes PPM updates to v2.5 with code cleanup, unnecessary-message reduction and shellcheck resolutions.


peabee:
Addition of ppm_advice.sh
This is specific to certain builds and also relies on lx_sfs_mgr being present (which is also specific to a subset of the builds). Not sure this should be a PPM feature?

There is a typo at line 4: web- should be get-

peabee:
Maybe ppm should just check if ppm_advice.sh exists and call it if it does - then each build can decide whether they want ppm_advice.sh and what they want it to display......

techrockedge:
that does sound interesting

PR #4294, opened 2024-06-12
PPM tweaks: Update installpkg.sh

rizalmart:
1. Use the save file path instead of /tmp as working directory. This will able to install large packages.

2. Install the extracted packages on savepath location rather than on "/" to avoid quick filling of working tmpfs on odd-numbered pupmodes

### DIFF
--- a/woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
+++ b/woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
@@ -291,9 +291,21 @@ elif [ $PUPMODE -eq 13 ] && [ "$PUNIONFS" != "overlay" ];then
 fi
 
 #Make directory to extract pkgs to  #jrb
-mkdir -p ${DIRECTSAVEPATH}/wkdir > /dev/null 2>&1 #230305 #jrb
-WKDIR=${DIRECTSAVEPATH}/wkdir #230305 #jrb
+
+TMPFREE=$(df -k /tmp | awk 'NR==2 {print $4}')
+DSAVEFREE=$(df -k /${DIRECTSAVEPATH} | awk 'NR==2 {print $4}')
+
+if [ $DSAVEFREE -ge $TMPFREE ]; then
+ WKDIR=${DIRECTSAVEPATH}/wkdir #230305 #jrb
+else
+ WKDIR=/tmp/petget_proc/wkdir #230305 #jrb
+fi
+
+mkdir -p ${WKDIR} > /dev/null 2>&1 #230305 #jrb
+
 wkdir_memcheck
+
+
 
 if [ -n "$DISPLAY" ] || [ -n "$WAYLAND_DISPLAY" ] && [ ! -f /tmp/petget_proc/install_quietly ];then #131222
  LANG=$LANG_USER
--- a/woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
+++ b/woof-code/rootfs-skeleton/usr/local/petget/installpkg.sh
@@ -70,15 +70,16 @@ Pfiles () {
 }
 
 Clear_wkdir () {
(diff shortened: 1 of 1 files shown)


rserwin1:
I feel that jrb and Marv need to be consulted about the removal of their work directory logic.

I infer from the clear_workdir code that the copy is made to replace the old version of each file.  The changes seem to copy the files to another /tmp directory.  How do the changes make the replacement?

I don't feel confident in recommending either accepting or rejecting the PR, at this point.

rserwin1 (commented):
See my comment (above).

peabee:
What happens if a "save file" does not exist?

Will ask Marv and JRB by pm on forum for comments.

rizalmart:
> What happens if a "save file" does not exist?
> 
> Will ask Marv and JRB by pm on forum for comments.

DIRECTSAVEPATH was a value of SAVE_LAYER variable in /etc/rc.d/PUPSTATE file

So it means if the save file doesn't exists it will just create working folder at "/"

Marv1943:
I have a couple of questions/reservations here.

One: I run all of my personal pups without savefiles so there should be an explicit default to /tmp in that case.

Two:  For my users, I add a very small (often only 128 or 256 Mb) savefile to customize the install for them.  That greatly simplifies updating for me.  Thus, a wkdir in /tmp is far larger on any of my (older -circa 2012 with 4 to 6 Gb RAM) laptops than a savefile based one.  I guess a size check and use of the larger space would work there but adds complexity.

Three:  I much prefer ONLY using/updating those savefiles when I choose to do so.  Either manually or on shutdown via dialog box.   This is for 'safety' reasons and has served me well.  They are not set up to be working memory.

rizalmart:
> Two: For my users, I add a very small (often only 128 or 256 Mb) savefile to customize the install for them. That greatly simplifies updating for me. Thus, a wkdir in /tmp is far larger on any of my (older -circa 2012 with 4 to 6 Gb RAM) laptops than a savefile based one. I guess a size check and use of the larger space would work there but adds complexity.

What if compare the free space of tmp againsts savefile in order to set working directory?

Marv1943:
Hi,

That suggestion made via email would work (testing for and using the larger workspace).  In the original code, there was also a built-in check to see that the workspace available (in /tmp) was big enough for the pet/deb being installed.  I assume that is still in there.

As an aside, almost all 'newer' pups run with a zram swap enabled by default.  That should expand the available size in /tmp though I've never tested it. 

My reservations about using a savefile as workspace remain though given my install style that would not impact me.  

Regards, Jim

Edited once, Jim

On Tue, 18 Jun 2024 16:56:57 -0700
Rizal Martin ***@***.***> wrote:

> > Two: For my users, I add a very small (often only 128 or 256 Mb) savefile to customize the install for them. That greatly simplifies updating for me. Thus, a wkdir in /tmp is far larger on any of my (older -circa 2012 with 4 to 6 Gb RAM) laptops than a savefile based one. I guess a size check and use of the larger space would work there but adds complexity.
> 
> What if compare the free space of tmp againsts savefile in order to set working directory?
> 
> 
> -- 
> Reply to this email directly or view it on GitHub:
> https://github.com/puppylinux-woof-CE/woof-CE/pull/4294#issuecomment-2177270163
> You are receiving this because you commented.
> 
> Message ID: ***@***.***>


-- 
Jim ***@***.***>

peabee:
I also don't like cluttering up the savefile/folder unless 100% necessary. With /tmp you're guaranteed it's gone on next boot.

rizalmart:
> I also don't like cluttering up the savefile/folder unless 100% necessary. With /tmp you're guaranteed it's gone on next boot.

It does not clutter because it stays on a single folder which was a folder for working dir. Also, it removes the working directory if the installation of package was complete

PR #4359, opened 2025-05-16
Update probepart: block device revision

rizalmart:
* filtered out /dev/loop*, /dev/zram*, and /dev/ram* block devices
* added /dev/scd[0-9] block devices for probing optical drives

### DIFF
--- a/woof-code/rootfs-skeleton/sbin/probepart
+++ b/woof-code/rootfs-skeleton/sbin/probepart
@@ -170,18 +170,22 @@ probepart_func() {
 
 # $1: device
 do_probepart() {
+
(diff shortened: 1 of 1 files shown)


peabee:
Why are you proposing this change? What problem reports are you attempting to fix? How extensively has this change been tested on a wide range of hardware? Has a temporary fix been offered to the forum for widespread testing?

peabee:
Closed without explanation??

rizalmart:
> Closed without explanation??

Sorry for late reply @peabee . I'm dealing with an error on shutdownconfig script where the error is the mounted iso became target drive for storing savefile on ventoy boot setup as I saw on the forum reported by Clarity

image

I thought it was on probepart first. But it turns out it was on the shutdownconfig script. So I have to close this PR

PR #4363, opened 2025-06-20
Update .profile: detect display managers

rizalmart:
Initiate xorg or wayland compositor only if there is no display manager running
Reason: Display Managers already initiated Xorg or Wayland compositors so skip some routines

### DIFF
--- a/woof-code/rootfs-skeleton/root/.profile
+++ b/woof-code/rootfs-skeleton/root/.profile
@@ -11,8 +11,11 @@ mkdir -p /tmp/services
 	echo "PATH='${PATH}'"
 ) > /tmp/services/user_info
 
+#Check for display managers
+DM_RUNNING=$(pidof lightdm gdm kdm sddm xdm lxdm slim ly wdm)
+
 if command -v startlabwc >/dev/null 2>&1 ; then
-	if [ ! -f /tmp/bootcnt.txt ] ; then
+	if [ ! -f /tmp/bootcnt.txt ] && [ "$DM_RUNNING" == "" ]; then
 		touch /tmp/bootcnt.txt
 		startlabwc
 	else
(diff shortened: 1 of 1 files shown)


peabee:
> DM_RUNNING=$(pidof lightdm gdm kdm sddm xdm lxdm slim ly wdm)

Don't like code like this which is likely to need updates in future.... is the list complete? Is there no other way to detect a display manager more generally? If you must have a list put it into a separate text file so code stays constant.

rizalmart:
> > DM_RUNNING=$(pidof lightdm gdm kdm sddm xdm lxdm slim ly wdm)
> 
> Don't like code like this which is likely to need updates in future.... is the list complete? Is there no other way to detect a display manager more generally? If you must have a list put it into a separate text file so code stays constant.

Ok I revised the code. I created list of display managers. See PR #4364

peabee:
It is not possible to guarantee that all these names are unique and only ever used by Display Managers? Other instances potentially would create havoc.
It would be better if there was only 1 entry in the list created during the Display Manager installation process. This can be done with a package template or by code in _00build.conf @ EXTRA_COMMANDS="

rizalmart:
> It is not possible to guarantee that all these names are unique and only ever used by Display Managers? Other instances potentially would create havoc. It would be better if there was only 1 entry in the list created during the Display Manager installation process. This can be done with a package template or by code in _00build.conf @ EXTRA_COMMANDS="

So where to put the display manager parameter? If you will be able to do that, I will just close this PR. The goal here was to add display manager support and enable multiuser puppy

peabee:
I can accept the code change, but the list in Woof-CE should be empty - an entry into the list should be inserted as part of the Display Manager install.

dimkr:
> enable multiuser puppy

To make that happen you'll need to many **so many** changes in other places first, like fixing many Puppy scripts and packages so they don't assume /root and the permissions to do anything.

If you can't get a tiny change merged, you should consider forking woof-CE, working on mutli-user support in this fork, and consider upstreaming the changes later once everything is ready.

(In addition, it's usually hard to convince somebody to merge proposed changes when they're split across many PRs, many of them changing only a single file or even a single line; if you open PRs like this because you use GitHub's web editor instead of using git, learning how to use git can be a huge time saver both for you and for the person reviewing your changes)

peabee:
> you should consider forking woof-CE, working on mutli-user support in this fork, and consider upstreaming the changes later once everything is ready.

Probably a good idea - although it may be better for the result to remain a fork.......... or at least it must be shown that the fork can successfully build all current active Github builds.....

PR #4365, opened 2025-06-21
Update snapmergepuppy: make sure use busybox pidof

rizalmart:
util-linux pidof does not recognized scripts. busybox does

### DIFF
--- a/woof-code/rootfs-skeleton/usr/sbin/snapmergepuppy
+++ b/woof-code/rootfs-skeleton/usr/sbin/snapmergepuppy
@@ -72,9 +72,9 @@ NEXTPASSTHRU=""
 [ -e /tmp/flagnextpassthru ] && NEXTPASSTHRU="yes"
 
 SHUTDOWN="no"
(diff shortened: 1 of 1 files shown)


peabee:
Why has this (& the 3 others) suddenly become an issue?? How and when are the 'scripts' being used and by whom? Dropping back from FULL to busybox seems to be a retrograde step?

also #4366 #4367 #4368

rizalmart:
> Why has this (& the 3 others) suddenly become an issue?? How and when are the 'scripts' being used and by whom? Dropping back from FULL to busybox seems to be a retrograde step?
> 
> also #4366 #4367 #4368

I made a test on using pidof. I ran a script and pidof it. the full pidof failed to detect a running script but busybox version does. Only puppy core files that using pidof to detect a running scripts are affected. This is an issue on detecting snapmergepuppy (which is a script) process and shutdown script process. 

You can try the experiment to see it

peabee:
As far as I know builds have been running successfully for many years with FULL pidof - what problem reports / issues are being fixed by this change?

PR #4367, opened 2025-06-21
Update rc.shutdown: use busybox pidof

rizalmart:
util-linux pidof does not recognize scripts. busybox does


techrockedge:
seems like we could merge the pull request.

dimkr:
Alternatively, you can change `snapmergepuppy` so it uses a lock file to prevent concurrent run (e.g. rename to `snapmergepuppy.real` and add a `snapmergepuppy` wrapper that simply runs `flock -Fx /tmp/snapmergepuppy.lock snapmergepuppy.real`).

PR #4386, opened 2025-10-07
Update rc.services: Avoid executing sysvinit scripts

rizalmart:


peabee:
Please explain why this is needed and for what Woof-CE build?
If it is specific to a particular build then there are many mechanisms within Woof-CE to accomplish this change just for that build.

rizalmart:
> Please explain why this is needed and for what Woof-CE build? If it is specific to a particular build then there are many mechanisms within Woof-CE to accomplish this change just for that build.

Because rc.services indiscriminately executes scripts as long the script marked as executable. Traditional sysvinit distro's such as devuan stores scripts on /etc/init.d then symlinked to its runlevel directory. Also rc scripts, traditional rc.sysinit, rcS, and even rc,shutdown scripts are stored in /etc/init.d. The goal here was to make sure that the service script to be executed is a legitimate service script.

peabee:
How has this problem arisen? What Woof-CE build?

rizalmart:
> How has this problem arisen? What Woof-CE build?
On Devuan builds since it uses sysvinit setup. I noticed that /etc/init.d/rc script setup where I tried devuan via distrosea

peabee:
I don't know what "distrosea" is??
The devuan woof-ce builds are working: https://github.com/puppylinux-woof-CE/woof-CE/actions/workflows/devuan-daedalus64.yml
I think this is code that has been working for many pup builds for a long time and if you need these changes they should be made locally in your build. Sorry.

PR #4415, opened 2026-03-22
Update rc.sysinit: enable zswap when using swap partition or swap file

rizalmart:


peabee:
See PM sent on forum seeking clarification.

PR #4416, opened 2026-03-22
Create 10-disk-io-scheduler.rules

rizalmart:
Improve disk io performance by applying appropriate io scheduler with tweaks


peabee:
See PM sent on forum seeking clarification.

PR #4420, opened 2026-03-30
Reorder PATH variable in /etc/profile

rizalmart:
Make path order aligned with the FHS

### DIFF
--- a/woof-code/rootfs-skeleton/etc/profile
+++ b/woof-code/rootfs-skeleton/etc/profile
@@ -2,7 +2,8 @@
 #100622 PKG_CONFIG_PATH is determined in 3builddistro (Woof), written to /etc/profile.d/pkgconfig
 #110804 fix double-login when exit from X, allow /etc/profile to complete. see also "echo -n '# '" in /usr/bin/xwin at exit.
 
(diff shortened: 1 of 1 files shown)


peabee (commented):
Why is this suddenly needed when the existing scheme has been working without any problems for a looooong time??
You seem to be giving precedence to /usr/local over the system level lib folders? Why?

rizalmart:
To make it fully aligned and compliant with the standard file system hierarchy about path ordering. I noticed the order when I check the $PATH values under pkexec. It go /usr/local first then /usr. also /usr/local is default path for make install command upon compiling source code

peabee:
@techrockedge FYI
I feel that such a significant change might have unintended consequences.
I don't see a need for Puppy to conform and be the same as other distros.
I may open a thread on the forum to describe what you are proposing and to ask if anyone can see problems.
As always you are free to implement example systems with these changes if you want - there are many Woof-CE mechanisms you could use - without imposing the changes on all other Woof-CE builds which are building and running satisfactorily.

rizalmart:
> I don't see a need for Puppy to conform and be the same as other distros.

However woof-ce uses major distros to build puppy and those major distros follows FHS standard path ordering. Much better if puppy follow the path ordering as close as possible to the mainstream distro.

Also putting /usr/local first makes wrapping, overriding of linux commands easier rather than renaming to "-FULL" suffix

PR #4447, opened 2026-07-07
Update DISTRO_SPECS

SimonHue6131:


woof-distro/x86_64/ubuntu/noble64/DISTRO_SPECS
peabee, line 1: Hi
You cannot change the noble64 DISTRO_SPECS in this way......
You seem to be making your own version - this should be kept local to your branch....
Please explain what you are trying to do if I have got this wrong.
Thanks

SimonHue6131: Shti bruh, sorry, not so familiar with github sorryyyyyyyyyyyyy

PR #4451, opened 2026-07-18
Add color scheme support to puppy.svg

rizalmart:
Add dark mode support


peabee (changes requested):
Why does changing it from "dark grey" to black "add color scheme support"????
Wouldn't it be easy to amend the svg during build (with a sed in _00build.conf) if the color change was needed for some reason?
Why change this for all builds?

PR #4461, opened 2026-09-10
set_hwclock_type: sync /etc/adjtime file

rizalmart:
make sure that /etc/adjtime was set either UTC or localtime hardware clock


peabee:
Please explain why you think this is needed? Have there been any problem reports on the forum that this is fixing? Does it apply to all builds?

rizalmart:
> Please explain why you think this is needed? Have there been any problem reports on the forum that this is fixing? Does it apply to all builds?

/etc/adjtime is one of the method of storing the clock type settings across multiple distro. much better to be sync it against /etc/clock

peabee:
"Much better" in what way?
I would like to see evidence of it being tested on the forum.

peabee:
Can this be closed?

